PR #12, opened 2016-08-05
IAR Floating Point 

sarahmarshy:
This is a known bug for exporters. This instructs users how to utilize FPU instructions and avoid IAR errors.


sarahmarshy:
Just saw a message from @screamerbg  that we might not be supporting IAR in release. The duplicate filenames issue?

iriark01:
@screamerbg or @sg- are we supporting IAR?

sg-:
we are supporting but i don't think this is the place for point docs about exporting errata... but maybe it is? Exporters in general need a really good hard refactor and cleanup

sarahmarshy:
@sg- where is a good place? It looks like there is a section [https://github.com/sarahmarshy/Handbook/blob/5fa80ca27f3aa510d39ec0379f8d5e4aefb4c7b4/docs/dev_tools/third_party.md](here) about third party tool specifics, and it documents individual known issues with each tool.

sarahmarshy:
@sg- obviously, this issue should ideally be fixed in software, but I thought it might be good to document it somewhere in the meantime.

sg-:
That is a fine place. Just need to get used to the change in workflow from just edit a file to PR and rebuild the docs 👍

iriark01:
Then I shall accept and rebuild the docs :) Thank you both

iriark01:
Added to 5.1

PR #23, opened 2016-11-08
Add a page about adding new exporters

theotherjimmy:
Not sure if this goes here. Feel free to move this.
@iriark01

### DIFF
--- a/docs/dev_tools/adding_exporters.md
+++ b/docs/dev_tools/adding_exporters.md
@@ -59,7 +59,8 @@ Each exporter is expected to implement one method, `generate`, which is responsi
 This method may use any of the attributes and APIs included by the common code.
 
 ___The `TARGETS` class variable___
-Each exporter reports its specific target support through a class varibale, `TARGETS`. This class variable is simply a list of targets that a user could export. Requesting an export of a target that's not in the list will generate an error.
+
+Each exporter reports its specific target support through a class varibale, `TARGETS`. This class variable is simply a list of targets to which you can export. Requesting an export to a target that's not on the list will generate an error.
 
 
 ___The `build` method___
--- a/docs/dev_tools/adding_exporters.md
+++ b/docs/dev_tools/adding_exporters.md
@@ -54,9 +54,13 @@ The Python code for the plugin should be:
 
 ___The `generate` method___
 
-Each exporter is expected to implement one method, `generate`, which is responsible for creating all of the required project files for the IDE or toolchain that the plugin targets. The method should provide the exporter's supported targets in a `TARGETS` class variable.
+Each exporter is expected to implement one method, `generate`, which is responsible for creating all of the required project files for the IDE or toolchain that the plugin targets. 
 
 This method may use any of the attributes and APIs included by the common code.
+
+___The `TARGETS` class variable___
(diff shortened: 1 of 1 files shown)


docs/dev_tools/adding_exporters.md
```diff
@@ -1,74 +1,97 @@
...
+
+The Python code for the plugin should be:
+
+1. Placed into an `__init__.py` file.
+1. Imported into `tools/export/__init__.py`.
+1. Added to the exporter map.
+
+___The `generate` method___
+
+Each exporter is expected to implement one method, `generate`, which is responsible for creating all of the required project files for the IDE or toolchain that the plugin targets. The method should provide the exporter's supported targets in a `TARGETS` class variable.
```
theotherjimmy: The `generate` method does not provide the supported targets. It's run on an object, where as `TARGETS` is a class variable.

docs/dev_tools/adding_exporters.md
```diff
@@ -1,74 +1,97 @@
...
 
-Each exporter is expected to implement one method, `generate`, which is responsible for creating all of the required project files for the IDE or Toolchain that the plugin targets and provide their supported Targets in a `TARGETS` class variable.
-This `generate` method may use any of the attributes and APIs included by the common code to accomplish its goal.
+___The `build` method___
 
-Further, a plugin that would like to be tested by CI may implement the `build` method. 
-This method is expected to be run after `generate` on an object that inherits from `Exporter`, and is responsible for invoking the build tools that the IDE or Toolchain would when a user instructs it to compile.
+A plugin that would like to be tested by CI may implement the `build` method. 
+
+This method should run after `generate` on an object that inherits from `Exporter`. It is responsible for invoking the build tools that the IDE or toolchain needs when a user instructs it to compile.
```
theotherjimmy: This is not a should. It actually happens that way.

docs/dev_tools/adding_exporters.md
iriark01: Could you rephrase what I wrote, then? Who runs it when and on what?

docs/dev_tools/adding_exporters.md
```diff
@@ -54,10 +54,14 @@ The Python code for the plugin should be:
 
 ___The `generate` method___
 
-Each exporter is expected to implement one method, `generate`, which is responsible for creating all of the required project files for the IDE or toolchain that the plugin targets. The method should provide the exporter's supported targets in a `TARGETS` class variable.
+Each exporter is expected to implement one method, `generate`, which is responsible for creating all of the required project files for the IDE or toolchain that the plugin targets. 
 
 This method may use any of the attributes and APIs included by the common code.
 
+___The `TARGETS` class variable___
+Each exporter reports its specific target support through a class varibale, `TARGETS`. This class variable is simply a list of targets that requesting an export is a valid operation for.
```
iriark01: This class variable is simply a list of targets to which you can export. Requesting an export for a target that's not on the list will generate an error.

theotherjimmy: That's correct.

iriark01:
Merging, but I think I'll move it to the contribution section, since the dev tools section is for using our dev tools, not enhancing them.

PR #29, opened 2016-11-30
Add compilers to path for make

theotherjimmy:
https://github.com/ARMmbed/mbed-os/issues/3338#issuecomment-263917168

This new blurb should prevent issues like the one linked above.


theotherjimmy:
@iriark01 @AnotherButler Edits welcome!

PR #45, opened 2017-01-04
Update intro.md

AnotherButler:
-Add punctuation for consistency

### DIFF
--- a/docs/advanced/intro.md
+++ b/docs/advanced/intro.md
@@ -3,7 +3,7 @@
 
 ## How to...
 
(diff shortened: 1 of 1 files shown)


AnotherButler:
Did we recently decide to combine the examples into one examples list? On docs.mbed.com, I see all of the examples, but on GitHub, I see only an exclamation mark and a URL.

iriark01:
!{} is how we transclude a page from another repo. We have a single repo for the examples list so we can use the page again and again without having to edit it in different places.

PR #72, opened 2017-01-11
Update file names in mkdocs.yml

AnotherButler:
Change file names in YML for clarity in organization


iriark01:
Did you also change all the links to the files? I have a vague memory that we rejected this change in another PR.

AnotherButler:
We did, but then I thought after that there was a note to do this? Or maybe I dreamed that? Shall I close this?

iriark01:
We don't want to change the file name - it's too much hassle and it doesn't give us anything since it's not where the page title comes from.

PR #75, opened 2017-01-11
Fix broken internal links on memap.md

AnotherButler:
Fix broken internal links


docs/concepts/memap.md
```diff
@@ -9,9 +9,9 @@
 ## Table of contents
 
 1. [Using memap](#using-memap)
```
iriark01: We don't need ToCs in the handbook; the left side navigation takes care of that

PR #79, opened 2017-01-12
Fix minor miswording in tls_porting.md

AnotherButler:
Fix minor miswording


iriark01:
Although suddenly I'm wondering if it's a function of mbed TLS or something we bundle with it?

AnotherButler:
@iriark01 I asked the engineers in the office before I created it, so I think it's correct as is

PR #68, opened 2017-01-10
Update build_script.md

AnotherButler:
Make initial grammar changes

PR #103, opened 2017-02-14
Update build_script.md

AnotherButler:
Make changes from #68 on 5.3 instead of master

NOTE: Please do NOT merge this because the content must be rewritten.


AnotherButler:
@studavekar This content is old and no longer correct. Do you feel comfortable updating/rewriting it for us?

PR #107, opened 2017-02-17
advanced/exporters: fix code block end

ilg-ul:
The triple back quote ending the code block is misplaced. Extra spaces removed.

### DIFF
--- a/docs/advanced/exporters.md
+++ b/docs/advanced/exporters.md
@@ -22,7 +22,7 @@ optional arguments:
   -S, --supported       Shows supported matrix of targets and toolchains
   -v, --verbose         Verbose diagnostic output
   -vv, --very_verbose   Very verbose diagnostic output
-  ```
+```
 
 # Adding export support for a target
 


ilg-ul:
I hope 5.3 is the right branch to send patches.

BlackstoneEngineering:
@ilg-ul The correct branch is always the `latest`, currently that is 5.3 . Thanks for letting us know. 

@AnotherButler just removed a space, doesnt cause issues on github but may cause issues with docs engine.

ilg-ul:
> The correct branch is always the latest, currently that is 5.3

the addition that I plan to do, i.e. the documentation for the GNU ARM Eclipse exporter, will be part of 5.4 (if I understood it right).

in this case, could you add the 5.4 branch?

> but may cause issues with docs engine.

that's exactly what happened, previously the exporters page showed the markdown following the code block _inside_ the code block. now it is ok.

BlackstoneEngineering:
The 5.4 branch will be pushed according to the docs team's schedule. I believe it will coincide with the 5.4 release. If you have the PR Ready then make it against the 5.3 branch and it will get picked up and carried into 5.4.

AnotherButler:
We recently removed the master branch to avoid confusion, so 5.3 is the latest branch. @BlackstoneEngineering is correct in that the 5.4 branch will coincide with the 5.4 release. Such releases usually occur every three months or so. A 5.4 branch will be added as we get closer to the release date. If you want to add documentation for a 5.4 feature and there's no 5.4 branch, please make your PR against 5.3 and note that the feature is for 5.4 in your PR or comment. We'll be sure not to merge it too early.

Also, this seems to be the same change that was merged in PR #109. Is that correct, or am I missing something?

ilg-ul:
I'll try to prepare a separate page for the GNU ARM Eclipse exporter, and make a new PR.

PR #117, opened 2017-02-27
Update mkdocs.yml

AnotherButler:
Add debugging docs to YML


iriark01:
"Lights of dead" sounds like the next Pirates of the Caribbean film.
Also, this should now exist in 5.4 as well

PR #125, opened 2017-03-07
Add specific toolchain versions

sg-:


iriark01:
Why not in https://github.com/ARMmbed/Handbook/blob/5.4/docs/dev_tools/third_party.md ?

sg-:
Its not about exporting, rather about specifying versions that mbed OS is compatible with.

iriark01:
Right, my bad

0xc0170:
Whats the status of this ? can it be integrated?

0xc0170:
@sg- Should IAR be 7.8 ? I overlooked it previously.

PR #129, opened 2017-03-22
Add hal/flash porting guide

0xc0170:
There is a new folder hal where a porting guides will be placed.
The first one, flash.md file is part of this commit. It contains information
about flash HAL API porting plus changes required for bootloader to use this API.

This is for 5.4 (flash API is part of it). This porting guide should describe flash HAL API, how to port it (2 options, both of them should be described).

Please review, I expect some changes within a day.

cc @c1728p9 @sg- @ashok-rao @AnotherButler

[Rendered](https://github.com/0xc0170/Handbook/blob/ead500140ed3b80153c88601cd32f584b8986182docs/hal/flash.md)

PR #130, opened 2017-03-23
Create flash.md

AnotherButler:
Move PR #129

### DIFF
--- a/docs/advanced/flash.md
+++ b/docs/advanced/flash.md
@@ -1,0 +1,120 @@
+# FLASH IAP porting guide
+
+Update target to support bootloader.
+   1. Update linker script
+   1. Implement flash HAL API
+   1. Verify changes with tests
+
+## Linker script updates
+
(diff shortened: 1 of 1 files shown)


0xc0170:
@AnotherButler I dont think this is better location. We are going to provide porting guides for HAL (more documents to come). We are mixing different worlds in advanced. I would like to sort them a bit.

Imagine hal would become
/hal/startup
/hal/linker_script
/hal/spi
/hal/i2c
/hal/flash

All of these are regarding porting a target to mbed (I go one by one to get the full mbed support).

cc @sg-

sg-:
This needs to move back where https://github.com/ARMmbed/Handbook/pull/129 had it. The porting guide for hal should workout like shown above https://github.com/ARMmbed/Handbook/pull/130#issuecomment-288650518

Eventually, this should not even be in the handbook rather the porting guide which is its own thing.

PR #131, opened 2017-03-23
Bootloader - add new references

0xc0170:
Please review

The bootloader should link to flash porting guide thus I added a reference here.

cc @c1728p9 @ashok-rao

### DIFF
--- a/docs/advanced/bootloader.md
+++ b/docs/advanced/bootloader.md
@@ -4,7 +4,7 @@ This guide explains how to create a bootloader, how to create a main program to 
 
 ## Porting bootloader
 
-Please follow [the bootloader porting guide](#flash.md) for adding a bootloader support for your target.
+Please follow [the bootloader porting guide](https://docs.mbed.com/docs/mbed-os-handbook/en/latest/advanced/flash/) for adding a bootloader support for your target.
 
 ## Creating the bootloader
 
--- a/docs/advanced/bootloader.md
+++ b/docs/advanced/bootloader.md
@@ -4,7 +4,7 @@ This guide explains how to create a bootloader, how to create a main program to 
 
 ## Porting bootloader
 
-Please follow [the bootloader porting guide](https://docs.mbed.com/docs/mbed-os-handbook/en/latest/advanced/flash/) for adding a bootloader support for your target.
+A target requires flash HAL functionality in order to support the bootloader. Please follow [the bootloader porting guide](https://docs.mbed.com/docs/mbed-os-handbook/en/latest/advanced/flash/) for more details.
 
 ## Creating the bootloader
 
--- a/docs/advanced/bootloader.md
+++ b/docs/advanced/bootloader.md
@@ -2,6 +2,10 @@
 
 This guide explains how to create a bootloader, how to create a main program to go with the bootloader and how to use this bootloader to perform software updates.
 
+## Porting bootloader
+
+Please follow [the bootloader porting guide](#flash.md) for adding a bootloader support for your target.
+
 ## Creating the bootloader
 
 Creating a bootloader is similar to creating a regular application. The only additional step you need is to specify the size of the bootloader as a target override in mbed_app.json in the target field "target.restrict_size":
@@ -9,7 +13,7 @@ Creating a bootloader is similar to creating a regular application. The only add
 ```
     "target_overrides": {
         ...
-        "NUCLEO_F429ZI": {
+        "<TARGET_NAME>": {
             "target.restrict_size": "0x20000"
         },
         ...
@@ -56,7 +60,7 @@ To create an application using a bootloader, you must first have created the boo
 ```
     "target_overrides": {
         ...
-        "NUCLEO_F429ZI": {
+        "<TARGET_NAME>": {
             "target.bootloader_img": "bootloader/my_bootloader.bin"
         },
         ...
--- a/docs/advanced/flash.md
+++ b/docs/advanced/flash.md
@@ -122,3 +122,13 @@ They test all flash API functionality. To run the tests, use these commands:
 
 - FlashIAP: `mbed test -m TARGET_NAME -n tests-mbed_drivers-flashiap`.
 - Flash HAL: `mbed test -m TARGET_NAME -n tests-mbed_hal-flash`.
+
+## Troubleshooting
+
+1. For targets with VTOR, a target might have VTOR address defined to a hardcoded address as mentioned in the ``Linker script updates`` section.
+
+1. Using FlashIAP might introduce latency as it might disable interrupts for longer periods of time.
+
+1. Program and erase functions operate on different blocks - erasing a block and programing a page. Use getter methods - get sector size and get page size.
+
+1. Sectors might have different sizes within a device.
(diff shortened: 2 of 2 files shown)


docs/advanced/bootloader.md
```diff
@@ -2,14 +2,18 @@
 
 This guide explains how to create a bootloader, how to create a main program to go with the bootloader and how to use this bootloader to perform software updates.
 
+## Porting bootloader
+
+Please follow [the bootloader porting guide](https://docs.mbed.com/docs/mbed-os-handbook/en/latest/advanced/flash/) for adding a bootloader support for your target.
```
ashok-rao: How about adding: "In order for a target to support Bootloader, the flash on the target needs to support IAP. Porting guide for the FLASH IAP is located here: https://docs.mbed.com/docs/mbed-os-handbook/en/latest/advanced/flash/ . " Then comes all the below i.e. how to create bootloader.. modify the target.restrict_size etc etc.
I feel this gives better context to why the flash IAP needs to be ported rather than starting off directly..

docs/advanced/bootloader.md
```diff
@@ -2,14 +2,18 @@
...
 ## Creating the bootloader
 
 Creating a bootloader is similar to creating a regular application. The only additional step you need is to specify the size of the bootloader as a target override in mbed_app.json in the target field "target.restrict_size":
 
 ```
     "target_overrides": {
         ...
-        "NUCLEO_F429ZI": {
+        "<TARGET_NAME>": {
             "target.restrict_size": "0x20000"
```
ashok-rao: Might want to make this generic too and provide an example in comments to show the format (hex). Example: "target.restrict_size": "<target_specific_mem_range>" #For nucleo_f429zi use 0x20000

0xc0170: The problem is that json does not support comments thus copy-paste users would fail. As we already removed specific target, this value then is just to show a number.

docs/advanced/bootloader.md
0xc0170: OK will rephrase to:

A target requires must implement flash HAL functionality in order to support the bootloader. Please follow [the bootloader porting guide](https://docs.mbed.com/docs/mbed-os-handbook/en/latest/advanced/flash/) for more details.

0xc0170:
Updated - rephrased the porting section.

docs/advanced/bootloader.md
ashok-rao: Ok. Let's keep it as it is.

docs/advanced/flash.md
```diff
@@ -118,3 +122,13 @@ They test all flash API functionality. To run the tests, use these commands:
 
 - FlashIAP: `mbed test -m TARGET_NAME -n tests-mbed_drivers-flashiap`.
 - Flash HAL: `mbed test -m TARGET_NAME -n tests-mbed_hal-flash`.
+
+## Troubleshooting
```
0xc0170: @ashok-rao I added Troubleshooting here to summarize what could go wrong. If there's anything that bootloader should also pinpoint, let me know, can be added.

0xc0170: Added that section, I found one that we could emphasize. commit 31cdb68

AnotherButler: Would it make more sense to title this section "Notes When Troubleshooting" or something similar because it's a list of things to remember as opposed to a list of steps to take when troubleshooting?

docs/advanced/flash.md
```diff
@@ -118,3 +122,13 @@ They test all flash API functionality. To run the tests, use these commands:
...
+
+1. For targets with VTOR, a target might have VTOR address defined to a hardcoded address as mentioned in the ``Linker script updates`` section.
+
+1. Using FlashIAP might introduce latency as it might disable interrupts for longer periods of time.
+
+1. Program and erase functions operate on different blocks - erasing a block and programing a page. Use getter methods - get sector size and get page size.
```
AnotherButler: Could you please clarify line 132? Are erasing a block and programming a page the two different blocks? Also, what are getter methods? Is that a term our readers already know and will recognize?

docs/advanced/flash.md
```diff
@@ -118,3 +122,13 @@ They test all flash API functionality. To run the tests, use these commands:
 
 - FlashIAP: `mbed test -m TARGET_NAME -n tests-mbed_drivers-flashiap`.
 - Flash HAL: `mbed test -m TARGET_NAME -n tests-mbed_hal-flash`.
+
+## Troubleshooting
+
+1. For targets with VTOR, a target might have VTOR address defined to a hardcoded address as mentioned in the ``Linker script updates`` section.
+
+1. Using FlashIAP might introduce latency as it might disable interrupts for longer periods of time.
```
AnotherButler: Please change "FlashIAP" to "Flash IAP" [two words].

docs/advanced/flash.md
```diff
@@ -118,3 +122,13 @@ They test all flash API functionality. To run the tests, use these commands:
 
 - FlashIAP: `mbed test -m TARGET_NAME -n tests-mbed_drivers-flashiap`.
 - Flash HAL: `mbed test -m TARGET_NAME -n tests-mbed_hal-flash`.
+
+## Troubleshooting
+
+1. For targets with VTOR, a target might have VTOR address defined to a hardcoded address as mentioned in the ``Linker script updates`` section.
```
AnotherButler: Should we link to the Linker script updates section instead of putting it in code? Also, please change "might have VTOR address" to "might have a VTOR address".

docs/advanced/flash.md
```diff
@@ -56,6 +56,10 @@ Bootloader-ready declaration of flash VTOR address:
 #endif
 ```
 
+## Start application
+
+The ``mbed_start_application`` is currently implemented only for Cortex-M3/M4/M7. The implementation is located in [the mbed_application code file](https://github.com/ARMmbed/mbed-os/blob/master/platform/mbed_application.c). If your target is not supported, you must implement this function in the target HAL.
```
AnotherButler: I'd like to change this paragraph to active voice. Would it be correct to say "The `mbed_start_application` implementation exists only for Cortex-M3, Cortex-M4 and Cortex-M7. You can find it in [the mbed_application code file](https://github.com/ARMmbed/mbed-os/blob/master/platform/mbed_application.c). If `mbed_start_application` does not support your target, you must implement this function in the target HAL."?

0xc0170: +1

docs/advanced/flash.md
0xc0170: lets link it

docs/advanced/flash.md
0xc0170: "In computer science, a mutator method is a method used to control changes to a variable. They are also widely known as setter methods. Often a setter is accompanied by a getter (also known as an accessor), which returns the value of the private member variable." 

accessor sounds better?

0xc0170:
@AnotherButler I pushed an update, please review

AnotherButler:
@0xc0170 Is this ready to merge?

0xc0170:
If you are happy with this, yes!

PR #137, opened 2017-03-28
Update contributing.md

jlovejoy:
Updated incorrect reference to Public Binary License (should be permissive) and updated header notice to better wording

### DIFF
--- a/docs/cont/contributing.md
+++ b/docs/cont/contributing.md
@@ -71,7 +71,7 @@ In order to clearly reflect the PBL license, please create three text files:
 
 * A *LICENSE* file with:
 
-<pre>Unless specifically indicated otherwise in a file, files are licensed under the Public Binary License, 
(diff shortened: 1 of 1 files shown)


docs/cont/contributing.md
```diff
@@ -74,20 +74,20 @@ In order to clearly reflect the PBL license, please create three text files:
 <pre>Unless specifically indicated otherwise in a file, files are licensed under the Public Binary License,
```
adamgreen: Still says Public Binary License on this line.

AnotherButler: @jlovejoy Should we change line 74 to Permissive Binary License, too?

jlovejoy: yes, fixed, missed that one, obviously!

AnotherButler:
Thanks so much!

PR #139, opened 2017-03-31
Remove CMSIS-DAP which is confused with DAPLink

tz-arm:
Remove CMSIS-DAP which is confused with DAPLink, otherwise we need give more description for CMSIS-DAP and the difference between CMSIS-DAP and DAPLink.

### DIFF
--- a/docs/advanced/DAP.md
+++ b/docs/advanced/DAP.md
@@ -1,8 +1,5 @@
 # DAPLink
 
-
-## DAPLink
-
 DAPlink is an open source project that implements the embedded firmware required for a Cortex debug probe. The project is hosted on GitHub and is published under an Apache 2.0 license, making it attractive for commercial developments.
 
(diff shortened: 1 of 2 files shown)


docs/advanced/DAP.md
```diff
@@ -1,4 +1,4 @@
-# CMSIS-DAP and DAPLink
+# DAPLink
 
 
 ## DAPLink
```
AnotherButler: Please delete the second heading, so we don't have two headings in a row that say the same thing

AnotherButler:
@tz-arm Thank you for the PR. This is a good point. I left one comment.

docs/advanced/DAP.md
tz-arm: Done, thanks for your review and comment!

AnotherButler: Thanks 👍

PR #140, opened 2017-03-31
Build profile docs change related to sleep

bulislaw:
This change will be needed when https://github.com/ARMmbed/mbed-os/pull/4097 is merged.

### DIFF
--- a/docs/dev_tools/build_profiles.md
+++ b/docs/dev_tools/build_profiles.md
@@ -5,6 +5,9 @@ mbed OS 5 supports three primary build profiles: *develop*, *debug* and *release
 * Small and fast code.
 * Full error information. For example, asserts have file name and line number.
 * Hard to follow code flow when using a debugger.
+* Chip is put to sleep when going idle:
+  * Debugger is likely to drop connection.
+  * Breaks the local file system on the [mbed interface](https://docs.mbed.com/docs/mbed-os-handbook/en/latest/getting_started/mbed_interface/) on some boards.
 
 ## Debug profile
 * Largest and slowest profile.
--- a/docs/dev_tools/build_profiles.md
+++ b/docs/dev_tools/build_profiles.md
@@ -10,6 +10,7 @@ mbed OS 5 supports three primary build profiles: *develop*, *debug* and *release
 * Largest and slowest profile.
 * Full error information. For example, asserts have file name and line number.
 * Easy to step through code with a debugger.
+* Disabled sleep mode.
 
 ## Release profile
 * Smallest profile and still fast.
--- a/docs/dev_tools/build_profiles.md
+++ b/docs/dev_tools/build_profiles.md
@@ -5,7 +5,7 @@ mbed OS 5 supports three primary build profiles: *develop*, *debug* and *release
 * Small and fast code.
 * Full error information. For example, asserts have file name and line number.
 * Hard to follow code flow when using a debugger.
-* Chip is put to sleep when going idle:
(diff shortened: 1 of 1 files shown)


bulislaw:
@AnotherButler

docs/dev_tools/build_profiles.md
```diff
@@ -10,6 +10,7 @@ mbed OS 5 supports three primary build profiles: *develop*, *debug* and *release
 * Largest and slowest profile.
 * Full error information. For example, asserts have file name and line number.
 * Easy to step through code with a debugger.
+* Sleep is disabled.
```
AnotherButler: To eliminate passive voice, I recommend changing this to "Disabled sleep" or "Disabled sleep mode". Would that keep the same meaning?

bulislaw:
Good point, will update.

Patater:
There is probably a way to rejigger things so you can avoid repeating yourself.

bulislaw:
I think it's cleaner to repeat small amount of text than to make some 'common' options with exclusions.

Patater:
Yes, a repeating small amount is much more readable.

However, the issues listed here for when sleep is enabled sound like good candidates for a known issues document. This location is probably not the first place a user would expect to find these known issues, although it may be handy to link to them from here. It's also probably difficult to keep these caveats up to date as they aren't as visible as a known issues document which would get reviewed on every release.

bulislaw:
I agree, unfortunately our docs are not necessarily very well structured or clear to find things in. https://docs.mbed.com/docs/mbed-os-handbook/en/latest/
To be frank I wouldn't be able easily find known issues (I think they are in blog posts per release but not in docs themselves).

Patater:
As an outsider, I'd look for known issues in the GitHub issue tracker or in a dedicated release notes document.

Anyway, making sure these issues are in the GitHub issue tracker or in release notes is not really something for this PR to deal with.

docs/dev_tools/build_profiles.md
```diff
@@ -5,11 +5,15 @@ mbed OS 5 supports three primary build profiles: *develop*, *debug* and *release
 * Small and fast code.
 * Full error information. For example, asserts have file name and line number.
 * Hard to follow code flow when using a debugger.
+* Chip is put to sleep when going idle:
```
AnotherButler: Query: Could we change this to "Chip goes to sleep we idle" and keep the meaning? I would like to avoid the passive voice in "is put to sleep".

Patater: "The RTOS idle task puts the chip to sleep." is accurate, but maybe too detailed.

"The chip sleeps when idle" is accurate as well.

AnotherButler:
@geky @theotherjimmy Should I merge this? The instructions say to merge it after we merge [ARMmbed/mbed-os#4097](https://github.com/ARMmbed/mbed-os/pull/4097), but I see [ARMmbed/mbed-os#4200](https://github.com/ARMmbed/mbed-os/pull/4200) reverts that.

geky:
I would wait, it does look like the revert will go through, but I suspect a slightly revised pr will be merged later.

AnotherButler:
@geky Thanks. I'll do that.

AnotherButler:
@geky Should I merge this yet or still wait?

bulislaw:
The revert PR was closed https://github.com/ARMmbed/mbed-os/pull/4200 I think that should be merged to reflect what's going on in the code. Then we should update the docs again when/if it changes.

PR #142, opened 2017-04-03
Edit and update targets.json docs

theotherjimmy:
Now with more key definitions!

### DIFF
--- a/docs/advanced/mbed_targets.md
+++ b/docs/advanced/mbed_targets.md
@@ -27,11 +27,11 @@ The style that we use for targets is:
 
 # Standard properties
 
-This section lists all the properties understood by the mbed build system. Unless specified otherwise, all properties are optional.
+This section lists all the properties the mbed build system understands. Unless specified otherwise, all properties are optional.
 
 ## `inherits`
 
-The description of an mbed target can "inherit" from one or more descriptions of other targets. When a target, called a _child_ inherits from another target, called its _parent_, the child automatically copies all the properties from the parent. After the child has copied the properties of the parent, It may then overwrite, add or remove from those properties. In our example above, `TEENSY3_1` inherits from `Target`. This is the definition of `Target`:
+The description of an mbed target can "inherit" from one or more descriptions of other targets. When a target, called a _child_ inherits from another target, called its _parent_, the child automatically copies all the properties from the parent. After the child has copied the properties of the parent, it may then overwrite, add or remove those properties. In our example above, `TEENSY3_1` inherits from `Target`. This is the definition of `Target`:
 
 ```
 "Target": {
@@ -51,7 +51,7 @@ Because `TEENSY3_1` inherits from `Target`:
 - `core` is a property defined both in `TEENSY3_1` and `Target`. Because `TEENSY3_1` overwrites it, the value of `core` for `TEENSY3_1` is `Cortex-M4`.
 - `default_toolchain` is not defined in `TEENSY3_1`. `TEENSY3_1` copies the definition of `default_toolchain` from `Target`, so the value of `default_toolchain` for `TEENSY3_1` is `ARM`.
 
-A child target may add properties that don't exist in any of its parents. For example, `OUTPUT_EXT` is defined in `TEENSY3_1`, but is not defined in `Target`.
+A child target may add properties that don't exist in any of its parents. For example, `OUTPUT_EXT` is defined in `TEENSY3_1` but is not defined in `Target`.
 
 It's possible, but discouraged, to inherit from more than one target. For example:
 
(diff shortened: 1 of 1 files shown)


docs/advanced/mbed_targets.md
```diff
@@ -18,15 +18,20 @@ mbed uses JSON as a description language for its build targets. The JSON descrip
         "detect_code": ["0230"]
 ```
 
-The definition of the target called **TEENSY3_1** is a JSON object. The properties in the object are either "standard" (understood by the mbed build system) or specific to the target.
+The style that we use for targets is:
+ - Each property is on its own line.
+ - The open brace `{` is on the same line as the target name, or the property name that is its key.
+ - The close brace `}` or close brace and comma `},` are on their own line.
+ - Lists take up at most one line.
+ - Lines are not split at any number of columns.
```
AnotherButler: Query: What does this mean?

docs/advanced/mbed_targets.md
```diff
@@ -108,31 +110,35 @@ For example:
...
 
-The list of **labels** defines how the build system looks for sources, libraries, include directories and any other files that are needed at compile time. `extra_labels` makes the build system aware of additional directories that must be scanned for such files.
+When you use target inheritance, you may alter the values of `extra_labels` using `extra_labels_add` and `extra_labels_remove`. This is similar to the `macros_add` and `macros_remove` mechanism described in the previous section.
 
-If target inheritance is used, it's possible to alter the values of `extra_labels` using `extra_labels_add` and `extra_labels_remove`. This is similar to the `macros_add` and `macros_remove` mechanism described in the previous paragraph.
+## `features`, `features_add` and `features_remove`
 
-## features, features_add, features_remove
+The list of _features_ defines which software features are enabled for a platform. Like `extra_labels`, `features` makes the build system aware of additional directories it must scan for resources.
```
AnotherButler: How can we change "are enabled" to active voice? What's enabling them?

docs/advanced/mbed_targets.md
```diff
@@ -141,13 +147,11 @@ Some mbed targets require specific actions for generating a binary image that ca
 }
 ```
 
-In this example, after the initial binary image for the target is generated, the build system will call the function `binary_hook` in the `TEENSY3_1Code` class. 
-
-<span class="notes">**Note:** The definition of the `TEENSY3_1Code` class **must** exist in the *targets.py* file. </span>
+After the generation of an initial binary image for the `TEENSY3_1`, the build system calls the function `binary_hook` in the `TEENSY3_1Code` class within `tools/targets/__init__.py`. 
 
-Because `toolchains` is also specified, `binary_hook` will only be called if the toolchain used for compiling the code is either `ARM_STD`, `ARM_MICRO` or `GCC_ARM`. Note that specifying `toolchains` is optional: if it's not specified, the hook will be called no matter what toolchain is used.
+The `TEENSY3_1` `post_binary_hook` only is called if the toolchain that compiles the code is `ARM_STD`, `ARM_MICRO` or `GCC_ARM`.
```
AnotherButler: Query: How can we change "is called" to active voice? What's calling it? You, the user?

docs/advanced/mbed_targets.md
```diff
@@ -161,13 +165,30 @@ class TEENSY3_1Code:
...
+
+The `delault_lib` property controls which library, small or standard, the `GCC_ARM` toolchain links. The `default_lib` property may take on the values `std` for the standard library and `small` for the reduced size library.
+
+## `bootloader_supported`
+
+The `bootloader_supported` property controls whether the build system allows a bootloader or a bootloader-using application to be built for a target. The default value of `bootloader_supported` is `false`.
+
+## `release_versions`
+
+The `release_versions` property is a list of major versions of mbed OS that the target supports. The list within `release_versions` may only contain `2`, indicating that the support of mbed OS 2, and `5`, indicating the support of mbed OS 5. Any target a `release_version` with a `2` in its list will be built as a static library as part of the mbed OS 2 release process.
```
AnotherButler: Could you please clarify the last sentence? I'm having trouble understanding it.

docs/advanced/mbed_targets.md
theotherjimmy: Normally, you would split long lines at some reasonable number of characters. 80 is common. Our targets.json file has no limit, and the style is not to split long lines.

theotherjimmy: I should change this to ` - Long lines are not split.`

docs/advanced/mbed_targets.md
theotherjimmy: The list of _features_ enables software features on a platform.

docs/advanced/mbed_targets.md
theotherjimmy: The bulid tools. so:

The bulid tools call the `TEENSY3_1` `post_binary_hook` when they build with the `ARM_STD`, `ARM_MICRO` or/and `GCC_ARM` compiler.

I leave writing `or` versus `and` up to you. The rules for English or versus and are not the same as discrete math, so I don't know them anymore.

docs/advanced/mbed_targets.md
theotherjimmy: We build all targets that are released for mbed OS 2. targets are released for mbed OS 2 by putting a `2` in the `release_version` list.

theotherjimmy: Sometimes it takes a few passes.

docs/advanced/mbed_targets.md
```diff
@@ -18,15 +18,20 @@ mbed uses JSON as a description language for its build targets. The JSON descrip
         "detect_code": ["0230"]
 ```
 
-The definition of the target called **TEENSY3_1** is a JSON object. The properties in the object are either "standard" (understood by the mbed build system) or specific to the target.
+The style that we use for targets is:
+ - Each property is on its own line.
+ - The open brace `{` is on the same line as the target name, or the property name that is its key.
+ - The close brace `}` or close brace and comma `},` are on their own line.
+ - Lists take up at most one line.
+ - Long Lines are not split.
```
AnotherButler: Please lower case "lines". Other than that, this looks great.

theotherjimmy: Whoops. Fixed.

docs/advanced/mbed_targets.md
```diff
@@ -18,15 +18,20 @@ mbed uses JSON as a description language for its build targets. The JSON descrip
...
+This section lists all the properties the mbed build system understands. Unless specified otherwise, all properties are optional.
 
-## inherits
+## `inherits`
 
-The description of an mbed target can "inherit" from one or more descriptions of other targets. When a target **A** inherits from another target **B**, (**A** is the _child_ of **B** and **B** is the _parent_ of **A**), it automatically "borrows" all the definitions of properties from **B** and can modify them as needed (if you're familiar with Python, this is very similar to class inheritance). In our example above, `TEENSY3_1` inherits from `Target` (most mbed targets inherit from `Target`). This is how `Target` is defined:
+The description of an mbed target can "inherit" from one or more descriptions of other targets. When a target, called a _child_ inherits from another target, called its _parent_, the child automatically copies all the properties from the parent. After the child has copied the properties of the parent, it may then overwrite, add or remove those properties. In our example above, `TEENSY3_1` inherits from `Target`. This is the definition of `Target`:
```
theotherjimmy: Probably keep the from that you removed. you cannot add or remove the property as a whole when it's inherited, you can add or remove members of the property's list though.

PR #144, opened 2017-04-05
Create greentea.md

AnotherButler:
Recreate README from greentea, so we can have docs about it in the Handbook


iriark01:
Really? Okay...

PR #147, opened 2017-04-10
Add `custom_targets.json` description

theotherjimmy:
It's a file in the root of a project directory where you can define
custom targets.

Documentation for ARMmbed/mbed-os#4103

# DO NOT MERGE BEFORE ARMmbed/mbed-os#4103 IS MERGED

### DIFF
--- a/docs/advanced/mbed_targets.md
+++ b/docs/advanced/mbed_targets.md
@@ -1,6 +1,6 @@
 # Adding and configuring mbed targets
 
-mbed uses JSON as a description language for its build targets. The JSON description of mbed targets can be found in `targets/targets.json`. To better understand how a target is defined, we'll use this example (taken from `targets.json`):
+mbed uses JSON as a description language for its build targets. The JSON description of mbed targets can be found in `targets/targets.json` and in `custom_targets.json` in the root of a project directory. Definitions of new targets in a `custom_targets.json` will be added to the list of available targets. Any re-definitions of existing targets will be merged with the original definition by replacing values found in `targets/targets.json` with values found in the same key in `custom_targets.json`. To better understand how a target is defined, we'll use this example (taken from `targets.json`):
 
 ```
     "TEENSY3_1": {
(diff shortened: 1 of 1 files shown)


theotherjimmy:
@AnotherButler Editing and recommendations appreciated!

@sg- Would you like to say more?

andrewleech:
@theotherjimmy Thanks for starting this.
I think it needs more, definitions in custom_targets aren't just overwriting existing keys, they're merged in.
Maybe rather than "Definitions in a `custom_targets.json` will overwrite definitions" something along the lines of:
`Definitions of new targets in a custom_targets.json will be added to the list of available targets. Any re-definitions of existing targets will be merged into the original one.`

Thinking about it now, perhaps the original overwrite approach in ARMmbed/mbed-os#4103 of replacing existing targets that are redefined rather than a recursive dict merge would be less confusing. Users who want to edit existing targets would then have to copy the whole target definition and edit it rather than just cherry picking keys, at least then the whole definition for a target is in one place.

theotherjimmy:
@andrewleech I like your suggestions. Just a moment. I'm adding them right now.

theotherjimmy:
@andrewleech I updated that sentence to read:
 Definitions in a `custom_targets.json` will be added to the list of available targets. Any re-definitions of existing targets will be merged with the original definition by replacing values found in `targets/targets.json` with values found in the same key in `custom_targets.json`.

sg-:
Just need to say something or mention this is early access and there will be changes as we provide stricter requirements for target inheritance and validation. Just cause it works now doesn't mean it wont fail a linter in the future.

docs/advanced/mbed_targets.md
```diff
@@ -1,6 +1,6 @@
 # Adding and configuring mbed targets
 
-mbed uses JSON as a description language for its build targets. The JSON description of mbed targets can be found in `targets/targets.json`. To better understand how a target is defined, we'll use this example (taken from `targets.json`):
+mbed uses JSON as a description language for its build targets. The JSON description of mbed targets can be found in `targets/targets.json` and in `custom_targets.json` in the root of a project directory. Definitions of new targets in a `custom_targets.json` will be added to the list of available targets. Any re-definitions of existing targets will be merged with the original definition by replacing values found in `targets/targets.json` with values found in the same key in `custom_targets.json`. To better understand how a target is defined, we'll use this example (taken from `targets.json`):
```
AnotherButler: Nice clarification. :)

I recommend changing the following:

- Change "The JSON description of mbed targets can be found in ..." to "You can find the JSON description of mbed targets in ..." for active voice.
- Change the sentence starting with "Definitions of new targets" to active voice. Who or what is adding the definitions to the list of available targets?
- Delete hyphen in "redefinitions".
- Change sentence starting with "Any redefinitions of existing targets" to active voice. Who or what is merging the redefinitions with original definitions?

AnotherButler:
@theotherjimmy Is this ready to merge now?

theotherjimmy:
Yep

PR #148, opened 2017-04-11
vscode: Note about the path to arm-none-eabi-gdb

janjongboom:
See also https://github.com/ARMmbed/mbed-os/pull/3915

@anotherbutler

Maybe the wording is a bit much...

### DIFF
--- a/docs/debugging/vscode.md
+++ b/docs/debugging/vscode.md
@@ -52,6 +52,8 @@ To configure the debugger for your project:
 
 ![Configuring the debugger](Images/vscode3.png)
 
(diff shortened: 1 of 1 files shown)


docs/debugging/vscode.md
```diff
@@ -52,6 +52,8 @@ To configure the debugger for your project:
 
 ![Configuring the debugger](Images/vscode3.png)
 
+<span class="notes">**Note:** If you installed the GNU ARM Embedded Toolchain in a non-default location (f.e. via the mbed CLI installer) you need to update the `MIDebuggerPath` to the full path of your copy of `arm-none-eabi-gdb`. To find out the new path, open a terminal and run `where arm-none-eabi-gdb` (Windows) or `which arm-none-eabi-gdb` (macOS and Linux).</span>
```
AnotherButler: Nice work. 

I recommend changing the following:
- Delete the hyphen in "nondefault"
- Spell out "for example," to match our style
- Add comma after "(installer)" because an introductory dependent clause requires a comma after it

Query: Would it be better to replace "find out" with "find" or "see", or does that change the meaning?

AnotherButler:
@janjongboom Would you like me to go ahead and make these changes, or would you prefer to make them yourself?

janjongboom:
@AnotherButler Go ahead!

PR #170, opened 2017-06-15
Create docs.json

micque01:
@iriark01 

Creating the new docs.json for mbed OS. I've added a provisional config for the mbed OS API which I hope gets us close to what we want.


iriark01:
Cheers!

PR #181, opened 2017-06-20
Create compiling_cli.md

AnotherButler:
Move part of cli.md here.


iriark01:
This reads (without me checking, to be honest) like we've left people without basic compilation instructions in the getting started section. Will need to look at it later.

PR #169, opened 2017-06-14
Update link in options.md

AnotherButler:
Update GNU ARM Embedded link to point to latest update

PR #182, opened 2017-06-22
Add IAR version information

yennster:
Added IAR version information to the "Exporting to third party Toolchains" page, since the [mbed-cli docs](https://github.com/ARMmbed/mbed-cli#through-mbed_settingspy) explicitly state that IAR versions older or newer than 7.80 will likely be incompatible with the tools.

Relevant mbed developer question: https://developer.mbed.org/questions/78375/use-iar-system-811-recompile-found-error/

### DIFF
--- a/docs/dev_tools/third_party.md
+++ b/docs/dev_tools/third_party.md
@@ -7,11 +7,11 @@ If you'd like to develop on mbed OS with a third party tool, or migrate to one, 
 - Make using:
     - GNU ARM Embedded Toolchain.
     - ARM Compiler 5.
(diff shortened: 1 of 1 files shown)


AnotherButler:
@yennster Does this contradict what @sg- said in #169 , or am I misreading this? I'm a little confused.

yennster:
@AnotherButler Just saw that, having personally tried an IAR version later than 7.80 anything later than 7.80 doesn't work with the tools. However 7.70 may very well work as @sg- said. Maybe we should update the pages to say versions 7.70 to 7.80.x will work?

PR #203, opened 2017-08-09
Created memory_intro article

yennster:
Content derived from https://developer.mbed.org/blog/entry/Optimizing-memory-usage-in-mbed-OS-52/

### DIFF
--- a/docs/tutorials/optimize/memory/memory_intro.md
+++ b/docs/tutorials/optimize/memory/memory_intro.md
@@ -1,5 +1,56 @@
 ## Memory Optimization
 
-[intro to the pages below]
+Beginning with Mbed OS 5, new features such as RTOS created an increase in flash and RAM usage. This guide explains how to optimize program memory usage for release builds using Mbed OS 5.
 
-[A generic document about writing code and optimizations for mbed OS. We should incorporate this blog: https://developer.mbed.org/blog/entry/Optimizing-memory-usage-in-mbed-OS-52/.]
+**Note:** More information about the memory usage differences between Mbed OS 2 and Mbed OS 5 is available [here](https://developer.mbed.org/blog/entry/Optimizing-memory-usage-in-mbed-OS-52/).
+
+### Removing Unused Modules
+
+For a simple program like [Blinky](https://github.com/ARMmbed/mbed-os-example-blinky), a program that flashes an LED, typical memory usage is split between the following modules: 
+
+```
++---------------------+-------+-------+-------+
+| Module              | .text | .data |  .bss |
++---------------------+-------+-------+-------+
+| Fill                |   132 |     4 |  2377 |
+| Misc                | 28807 |  2216 |    88 |
+| features/frameworks |  4236 |    52 |   744 |
+| hal/common          |  2745 |     4 |   325 |
+| hal/targets         | 12172 |    12 |   200 |
+| rtos/rtos           |   119 |     4 |     0 |
+| rtos/rtx            |  5721 |    20 |  6786 |
+| Subtotals           | 53932 |  2312 | 10520 |
++---------------------+-------+-------+-------+
+```
+
+Even though we are creating a release build, the `features/frameworks` module includes the Mbed OS test tools. Because of this, we are building one of our test harnesses into every binary. Since we do not need a testing framework for a release build, [removing this module](https://github.com/ARMmbed/mbed-os/pull/2559) will save a significant amount of RAM and flash memory.
+
+#### Printf and UART
+
+Other modules that are not used in your program can also be removed by the linker. For example, in [Blinky](https://github.com/ARMmbed/mbed-os-example-blinky) neither `printf` or UART drivers are used within its `main` program. However, every Mbed OS module handles traces and assertions by redirecting their error messages to `printf` on serial output - forcing the `printf` and UART drivers to be compiled in, requiring a large amount of flash memory.
+
+To disable error logging to serial output, set the `NDEBUG` macro and the following configuration parameter in your program's mbed_app.json file:
+
+```
+{
+    "macros": [ 
+        "NDEBUG=1"
+    ],
+    "target_overrides": {
+        "*": {
+            "platform.stdio-flush-at-exit": false
+        }
+    }
+}
+```
+
+**Note:** Different compilers, different results; compiling with one compiler with yield different memory usage savings than when compiling with another.
+
+#### Embedded Targets
+
+We can also take advantage of the fact that we run our programs only on embedded targets. When you run a C++ application on a desktop computer, the runtime constructs every global C++ object before `main` is called. It also registers a handle to destroy these objects when the program ends. The compiler injects this and has some implications for the application:
+
+* The code injected by the compiler consumes memory.
+* It implies dynamic memory allocation, and thus requires `malloc` to be included in the binary, even when not used by the application.
+
+When we run an application on an embedded device we don't need handlers to destroy objects when the program exits, because the application will never end. By [removing destructor registration](https://github.com/ARMmbed/mbed-os/pull/2745) on application startup, and by eliminating the code to destruct objects when `exit()` is called, saving even more RAM and flash memory usage.
--- a/docs/tutorials/optimize/memory/memory_intro.md
+++ b/docs/tutorials/optimize/memory/memory_intro.md
@@ -4,9 +4,9 @@ Beginning with Mbed OS 5, new features such as RTOS created an increase in flash
 
 **Note:** More information about the memory usage differences between Mbed OS 2 and Mbed OS 5 is available [here](https://developer.mbed.org/blog/entry/Optimizing-memory-usage-in-mbed-OS-52/).
 
-### Removing Unused Modules
+### Removing unused modules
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/optimize/memory/memory_intro.md
```diff
@@ -1,5 +1,56 @@
...
+| features/frameworks |  4236 |    52 |   744 |
+| hal/common          |  2745 |     4 |   325 |
+| hal/targets         | 12172 |    12 |   200 |
+| rtos/rtos           |   119 |     4 |     0 |
+| rtos/rtx            |  5721 |    20 |  6786 |
+| Subtotals           | 53932 |  2312 | 10520 |
++---------------------+-------+-------+-------+
+```
+
+Even though we are creating a release build, the `features/frameworks` module includes the Mbed OS test tools. Because of this, we are building one of our test harnesses into every binary. Since we do not need a testing framework for a release build, [removing this module](https://github.com/ARMmbed/mbed-os/pull/2559) will save a significant amount of RAM and flash memory.
```
AnotherButler: - Query: Is this true for the other build profiles [not only release, but also debug and develop]?
- Please change "we" to "you" in the first and second sentences.
- In the third sentence, please change "Since" to "Because" because "since" shows progression of time, and "because" shows cause and effect.
- In the third sentence, please change "we do not need a testing framework for a release build," to "a release build does not require a testing framework" for consistent person.
- In the third sentence, please change "will save" to "saves" for consistent tense.

docs/tutorials/optimize/memory/memory_intro.md
```diff
@@ -1,5 +1,56 @@
...
+
+#### Printf and UART
```
AnotherButler: Please put "Printf" in code format.

docs/tutorials/optimize/memory/memory_intro.md
```diff
@@ -1,5 +1,56 @@
...
+
+Other modules that are not used in your program can also be removed by the linker. For example, in [Blinky](https://github.com/ARMmbed/mbed-os-example-blinky) neither `printf` or UART drivers are used within its `main` program. However, every Mbed OS module handles traces and assertions by redirecting their error messages to `printf` on serial output - forcing the `printf` and UART drivers to be compiled in, requiring a large amount of flash memory.
```
AnotherButler: - Please change first sentence to "The linker can also remove other modules that your program does not use." for active voice.
- Please change second sentence to "For example, [Blinky's main program](https://github.com/ARMmbed/mbed-os-example-blinky) doesn't use `printf` or UART drivers." for active voice.
- Query: In the last sentence, could we change "compiled in" to "compiled", or would that change the meaning?
- In the last sentence, please change "compiled in, requiring a large" to "compiled in and requiring a large" for clarity.

docs/tutorials/optimize/memory/memory_intro.md
```diff
@@ -1,5 +1,56 @@
...
+}
+```
+
+**Note:** Different compilers, different results; compiling with one compiler with yield different memory usage savings than when compiling with another.
+
+#### Embedded Targets
+
+We can also take advantage of the fact that we run our programs only on embedded targets. When you run a C++ application on a desktop computer, the runtime constructs every global C++ object before `main` is called. It also registers a handle to destroy these objects when the program ends. The compiler injects this and has some implications for the application:
+
+* The code injected by the compiler consumes memory.
```
AnotherButler: Please change "The code injected by the compiler" to "The code the compiler injects" for active voice.

docs/tutorials/optimize/memory/memory_intro.md
```diff
@@ -1,5 +1,56 @@
...
+        }
+    }
+}
+```
+
+**Note:** Different compilers, different results; compiling with one compiler with yield different memory usage savings than when compiling with another.
+
+#### Embedded Targets
+
+We can also take advantage of the fact that we run our programs only on embedded targets. When you run a C++ application on a desktop computer, the runtime constructs every global C++ object before `main` is called. It also registers a handle to destroy these objects when the program ends. The compiler injects this and has some implications for the application:
```
AnotherButler: - Please change "We" to "You" at the start of the first sentence for consistent person.
- Please change "we run our programs only" to "these programs only run" for clarity.
- Query: Who or what calls `main`? Please clarify for active voice.

docs/tutorials/optimize/memory/memory_intro.md
```diff
@@ -1,5 +1,56 @@
...
+
+* The code injected by the compiler consumes memory.
+* It implies dynamic memory allocation, and thus requires `malloc` to be included in the binary, even when not used by the application.
+
+When we run an application on an embedded device we don't need handlers to destroy objects when the program exits, because the application will never end. More RAM and flash memory usage can be saved by [removing destructor registration](https://github.com/ARMmbed/mbed-os/pull/2745) on application startup, and by eliminating the code to destruct objects when `exit()` is called.
```
AnotherButler: - Please change "we" to "you" both times in the first sentence for consistent person.
- In the first sentence, please add a comma after "device" because an introductory dependent clause requires a comma after it.
- In the second sentence, please change "More RAM and flash memory usage can be saved by" to "You can save more RAM and flash memory usage by" for active voice.
- Please delete the comma after "startup".
- Query: Who or what calls `exit()`? Please clarify for active voice.

docs/tutorials/optimize/memory/memory_intro.md
```diff
@@ -1,5 +1,56 @@
...
+```
+
+**Note:** Different compilers, different results; compiling with one compiler with yield different memory usage savings than when compiling with another.
+
+#### Embedded Targets
+
+We can also take advantage of the fact that we run our programs only on embedded targets. When you run a C++ application on a desktop computer, the runtime constructs every global C++ object before `main` is called. It also registers a handle to destroy these objects when the program ends. The compiler injects this and has some implications for the application:
+
+* The code injected by the compiler consumes memory.
+* It implies dynamic memory allocation, and thus requires `malloc` to be included in the binary, even when not used by the application.
```
AnotherButler: Please change to "It implies dynamic memory allocation and thus requires the binary to include `malloc`, even when the application does not use it." for active voice.

docs/tutorials/optimize/memory/memory_intro.md
yennster: Even though it's in a heading?

docs/tutorials/optimize/memory/memory_intro.md
yennster: Query: the operating system calls the main() function at runtime, not sure how to clarify that in the article

docs/tutorials/optimize/memory/memory_intro.md
yennster: I'm so bad at passive vs. active 🤣

docs/tutorials/optimize/memory/memory_intro.md
yennster: Query: if the exit() function is placed in the program's code, then it is called at runtime by the OS

yennster:
@AnotherButler I made most of those changes, except:

- Are you sure Printf in [this heading](https://github.com/yennster/Handbook/blob/new_engine/docs/tutorials/optimize/memory/memory_intro.md#printf-and-uart) should be in code format?

docs/tutorials/optimize/memory/memory_intro.md
```diff
@@ -1,5 +1,56 @@
...
+        }
+    }
+}
+```
+
+**Note:** Different compilers, different results; compiling with one compiler yields different memory usage savings than compiling with another.
+
+#### Embedded targets
+
+You can also take advantage of the fact that these programs only run on embedded targets. When you run a C++ application on a desktop computer, the runtime constructs every global C++ object before `main` is called. It also registers a handle to destroy these objects when the program ends. The code the compiler injects has some implications for the application:
```
AnotherButler: Query: Would the content still be correct if we changed the second sentence to "When you run a C++ application on a desktop computer, the operating system constructs every global C++ object before calling `main` at runtime." for active voice?

docs/tutorials/optimize/memory/memory_intro.md
```diff
@@ -1,5 +1,56 @@
...
+
+* The code injected by the compiler consumes memory.
+* It implies dynamic memory allocation and thus requires the binary to include `malloc`, even when the application does not use it.
+
+When you run an application on an embedded device, you don't need handlers to destroy objects when the program exits, because the application will never end. You can save more RAM and flash memory usage by [removing destructor registration](https://github.com/ARMmbed/mbed-os/pull/2745) on application startup and by eliminating the code to destruct objects when `exit()` is called at runtime.
```
AnotherButler: Query: Would the content still be correct if we changed the end of the second sentence to "and by eliminating the code to destruct objects when the operating system calls `exit()` at runtime." for active voice?

docs/tutorials/optimize/memory/memory_intro.md
yennster: @AnotherButler Makes sense to me, I'm changing it now

yennster:
@AnotherButler made those changes :) and fixed a "we" to "you"

yennster:
I'm going to squash these commits actually @AnotherButler

yennster:
@AnotherButler done

PR #210, opened 2017-08-31
Add callback documentation

kegilbert:
Documentation on mbed Callbacks. Ended up a bit lengthy as we wanted to include information on callbacks as a whole to provide context for the mbed Callback class.

@AnotherButler 
@geky


AnotherButler:
@kegilbert @geky I just made my first edits. Please note I cut large chunks to use elsewhere. Please review my edits carefully, and let me know if the content is still correct. Please make any changes necessary to keep this accurate and making sense.

docs/refs/api/rtos/callbacks.md
```diff
@@ -1,9 +1,384 @@
...
+
+``` c++
+// Callback</*return type*/(/*parameters*/)> cb;
+Callback<int(float)> cb;          // A callback that takes in a float and returns an int
+Callback<void(float)> cb;         // A callback that takes in a float and returns nothing
+Callback<int()> cb;               // A callback that takes in nothing and returns an int
+Callback<void(float, float)> cb;  // A callback that takes in two floats and returns nothing
+```
+
+You can create a Callback directly from a C function or function pointer with the same type:
```
AnotherButler: Query: Didn't we already tell them about the content between lines 130 and 154? How is this different from the content above? Can we combine them?

AnotherButler:
@kegilbert @geky Also, could y'all do a quick Ctrl + F and make sure we consistently capitalize "callback" when it refers to the class and keep it lower case the rest of the time?

docs/refs/api/rtos/callbacks.md
```diff
@@ -169,25 +70,13 @@ int main() {
 }
 ```
 
-#### Combinatorial explosion damage control (aka why a separate class for Callbacks)
+#### A separate class for Callbacks)
```
kegilbert: Left in )

docs/refs/api/rtos/callbacks.md
kegilbert: Essentially, the following lines are more of an active example whereas the above lines are more of a template.

docs/refs/api/rtos/callbacks.md
```diff
@@ -1,9 +1,384 @@
...
+     serial.attach(&echo, Serial::RxIrq);
+ }
+```
+
+#### The importance of state
+
+Callbacks have two important pieces of information, the code to execute and the state associated with the callback.
+
+Without state, callbacks offer limited composability. In C, you can fix this by adding an additional “state” argument to the function pointer, so you can pass in the opaque “state” when you register the callback.
+
```
geky: Why remove the ADC examples?
1. Theres no longer an example of the limitations that come with not supporting state. Most of the value of the Callback class is lost if this point is missed. 
2. We are not covering all of the function types that the Callback class supports.

If we're moving these elsewhere we should probably just remove these sections (importance of state and separate callback class), since they only have limited value without showing all of the options available for callbacks.

We've also removed the motivation for introducing the Callback class. Is this going elsewhere?

docs/refs/api/rtos/callbacks.md
geky: Where did we tell them? The above code blob just covers the type declaration, which is different from the function's type declaration. The two types are independent and if a user is unfamiliar with c function declarations they could be lost.

It's repetitive for the sake of not missing information.

docs/refs/api/rtos/callbacks.md
```diff
@@ -169,25 +70,13 @@ int main() {
...
-
-5. Another C++ feature is volatile-correctness in case the underlying state must be volatile, but if necessary we can probably expect the user to hide volatile members inside of a non-volatile class.
-
-Long story short, C++ requires a large set of overloads to support all of the standard function types. It is very error-prone and unreasonable to expect a new library author to add all of these overloads to every function that could take in a callback.
-
-Fortunately, C++ does provide the tools to delegate this problem to a single class. This class is the Callback class. The Callback class should be familiar to users of the std::function class that was introduced in C++11, but is available for older versions of C++.
-
-**An overly-simplified description of the Callback class is that is contains all of this madness so you don’t have to.**
+**An overly-simplified description of the Callback class is that it contains all of this madness, so you don’t have to.**
```
geky: The "this madness" in the sentence doesn't refer to anything if we drop the previous bullet points. Are we just wanting to drop this section?

docs/refs/api/rtos/callbacks.md
```diff
@@ -263,11 +152,11 @@ ADC adc;
 adc.attach(dosomething);
 ```
 
-But what about state? The Callback type also supports passing a state pointer for a function. This state can be either a pointer to an object that is passed to a member function, or it can be a pointer passed to a C-style function.
+The Callback type also supports passing a state pointer for a function. This state can be either a pointer to an object that is passed to a member function, or a pointer passed to a C-style function.
```
geky: Are we not wanting to tie this sentence to the previous emphasis on state?

docs/refs/api/rtos/callbacks.md
```diff
@@ -306,13 +195,14 @@ thing_t thing2;
 adc.attach(callback(catinthehat, &thing2));
 ```
 
-An important thing to note, is that this state is restricted to a single pointer. This means you can’t bind both an object and argument to a callback.
+This state is restricted to a single pointer. This means you can’t bind both an object and argument to a callback.
```
geky: Can we make this a note thing or add some sort of emphasis, I get asked this question a lot and this specific point surprises people.

docs/refs/api/rtos/callbacks.md
```diff
@@ -353,7 +243,7 @@ void callmemaybe(Callback<void(float)> cb) {
 }
 ```
 
-The Callback class is what’s known in C++ as a “Concrete Type”, that is, the Callback class is lightweight enough to be passed around like an int, pointer, or other primitive type.
+The Callback class is a “Concrete Type”. That is, the Callback class is lightweight enough to be passed around like an int, pointer or other primitive type.
```
geky: Is there a way we can emphasize the "Concrete Type" is a C++ term? This scoping is important if the reader wants to google for "Concrete Type" since the term has different meanins in different languages.

docs/refs/api/rtos/callbacks.md
```diff
@@ -169,25 +70,13 @@ int main() {
...
 ```
 
-#### Combinatorial explosion damage control (aka why a separate class for Callbacks)
+#### A separate class for Callbacks)
 
-Unfortunately, supporting all of the standard C++ function types is difficult.
+C++ requires a large set of overloads to support all of the standard function types. It is unreasonable to expect a new library author to add all of these overloads to every function that could take in a callback.
 
-1. State is important, so need to support either C-style function pointers with state, or C++ member function pointers.
+However, C++ provides the tools to delegate this problem to a single class. This class is the Callback class. The Callback class should be familiar to users of the `std::function` class that C++11 introduced but is available for older versions of C++.
```
geky: A bit ambiguous, how about:
The Callback class should be familiar to users of the `std::function` class that C++11 introduced but the Callback class is available for older versions of C++

docs/refs/api/rtos/callbacks.md
```diff
@@ -169,25 +70,13 @@ int main() {
 }
 ```
 
-#### Combinatorial explosion damage control (aka why a separate class for Callbacks)
+#### A separate class for Callbacks)
 
-Unfortunately, supporting all of the standard C++ function types is difficult.
+C++ requires a large set of overloads to support all of the standard function types. It is unreasonable to expect a new library author to add all of these overloads to every function that could take in a callback.
```
geky: What about combining these two:
Unfortunately, supporting all of the standard C++ function types is difficult. C++ requires a large set of overloads to support all of the standard function types. It is unreasonable to expect a new library author to add all of these overloads to every function that could take in a callback.

Although now we're making claims without backing up the points. I'm worried that's just going to lead to more users underestimating the problems arount function argument overloads and roll their own hacks rather than seeing the value of the Callback class and adopting it.

The problems with incompatible Callback types only grows if we don't convince readers that staying in a single boat is the best thing to do.

geky: Although now we're making claims without backing up the points. I'm worried that's just going to lead to more users underestimating the problems arount function argument overloads and roll their own hacks rather than seeing the value of the Callback class and adopting it.

The problems with incompatible Callback types only grows if we don't convince readers that staying in a single boat is the best thing to do.

docs/refs/api/rtos/callbacks.md
```diff
@@ -37,96 +19,15 @@ Serial serial(USBTX, USBRX);
...
-For example, consider applying a low-pass filter to two different ADC modules
+Without state, callbacks offer limited composability. In C, you can fix this by adding an additional “state” argument to the function pointer, so you can pass in the opaque “state” when you register the callback.
 
-``` c++
-// Here is a small running-average low-pass filter.
-float low_pass_result;
-void low_pass_step(float data) {
-     low_pass_result = low_pass_result*0.99 + data*0.01;
-}
+One of the core features of C++ is the encapsulation of this “state” in classes, with operations that modify the state being represented as member functions in the class. Unfortunately, member function pointers are not compatible with standard function pointers, but you can use member function pointers to pass in state as a C++ object.
```
geky: I'm worried comparing the terms "standard function pointers" and "member function pointers" is just going to be confusing if we don't establish what they are.

geky:
I'm curious what the "elsewhere" is going to be and what's going to be in it.

We shouldn't finalize this document until all of the information has a place to live.

If we're moving the "motivation" sections elsewhere, we should probably just remove those sections completely from this document, since their contents kinda fall apart.

geky:
@AnotherButler split thrice, feel free to reedit :D

docs/refs/api/rtos/callbacks.md
```diff
@@ -1,9 +1,184 @@
...
+
+ADC adc;
+adc.attach(dosomething);
+```
+
+But what about state? The Callback type also supports passing a state pointer for a function. This state can be either a pointer to an object that is passed to a member function, or a pointer passed to a C-style function.
+
+Because this form of creating Callbacks requires two arguments, you need to create the Callback explicitly using the Callback constructor. The Callback also comes with the lowercase callback function, which creates callbacks based on the arguments type and avoids the need to repeat the template type.
+
+YOu can create a callback with a member function.
```
kegilbert: Capitalized O

AnotherButler:
@geky @kegilbert Are y'all ready for me to merge this and then add it to the .json?

docs/refs/api/rtos/callbacks.md
```diff
@@ -149,7 +149,7 @@ void callmemaybe(Callback<void(float)> cb) {
 }
 ```
 
-The Callback class is what’s known in C++ as a “Concrete Type”, that is, the Callback class is lightweight enough to be passed around like an int, pointer, or other primitive type.
+The Callback class is what’s known in C++ as a “Concrete Type”. That is, the Callback class is lightweight enough to be passed around like an int, pointer or other primitive type.
```
geky: Good bye oxford comma 😢

docs/refs/api/rtos/callbacks_the_importance_of_state.md
```diff
@@ -1,15 +1,14 @@
 
 ### The importance of state
 
-A callback is a user provided piece of code that is passed to another function. The callback allows an API to execute the user’s code in its own context. More info on how to use callbacks can be found in the [technical callback documentation](callbacks.md)
+A callback is a user provided function that a user may pass to an API. The callback allows the API to execute the user’s code in its own context. You can find more information on how to use callbacks in the [technical callback documentation](callbacks.md).
 
 #### Why not function pointers?
 
-Callbacks have two important pieces of information, the code to execute, and the state associated with the callback.
+Callbacks have two important pieces of information, the code to execute and the state associated with the callback.
```
geky: You will be missed oxford comma 😢

geky:
Looks great now : D

docs/refs/api/rtos/callbacks.md
```diff
@@ -1,9 +1,184 @@
...
+class Thing {
+    int state;
+    void catinthehat(float f) {
+        state = // do something
+    }
+}
+
+// We can create a Callback with the Callback constructor
+Thing thing1;
+adc.attach(Callback<void(int)>(&thing1, &Thing::catinthehat));
```
sarahmarshy: The param of catinthehat is float, so should it be:
`adc.attach(Callback<void(float)>(&thing1, &Thing::catinthehat));`

docs/refs/api/rtos/callbacks.md
```diff
@@ -1,9 +1,184 @@
...
+    int state;
+}
+
+void catinthehat(thing_t *thing, float f) {
+    thing->state = // do something
+}
+
+// We can create a Callback with the Callback constructor
+thing_t thing1;
+adc.attach(Callback<void(int)>(catinthehat, &thing1));
```
sarahmarshy: Should it be:
`adc.attach(Callback<void(float)>(catinthehat, &thing1));`

docs/refs/api/rtos/callbacks.md
```diff
@@ -1,9 +1,184 @@
...
+void callme(Callback<void(float)> cb) {
+    cb(1.0f);
+}
+```
+
+The only thing to watch out for is that the Callback type has a null Callback, just like a null function pointer. Uninitialized callbacks are null and assert if you call them. If you want a call to always succeed, you need to check if it is null first.
+
+``` c++
+void callmemaybe(Callback<void(float)> cb) {
+    If (cb) {
```
sarahmarshy: `If` => `if`

docs/refs/api/rtos/callbacks_sonar_example.md
```diff
@@ -0,0 +1,106 @@
+### Sonar Example
+
+Here is an example that uses everything discussed in the form of a minimal Sonar class.
+
+```c++
```
sarahmarshy: Can this sample be made into a repository so that it can be edited without changes to the docs?

docs/refs/api/rtos/callbacks.md
```diff
@@ -1,9 +1,184 @@
-#### Callbacks
+### Callbacks
 
-[Document about callbacks, their best practices and dos and don'ts.]
+A callback is a user provided function that a user may pass to an API. The callback allows the API to execute the user’s code in its own context.
 
-###### Thread example with callbacks
+```c++
```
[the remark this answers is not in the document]
geky: Hmm, it does have the quick description of callback above, what do you think should go here?

docs/refs/api/rtos/callbacks.md
geky: I'll actually make catinthehat take int

docs/refs/api/rtos/callbacks.md
geky: Haha guess which editor we were using

docs/refs/api/rtos/callbacks_sonar_example.md
geky: Hmm, probably, do examples each get their own repos?

docs/refs/api/rtos/callbacks.md
geky: made catinthehat take int 👍

docs/refs/api/rtos/callbacks.md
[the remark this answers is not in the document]
sarahmarshy: Just a little blurb like, here is an example of an API invoking user code, or something. Does the serial attach use mbed callback class? Or is this just an example of some API that can call a user defined function when some event occurs?

docs/refs/api/rtos/callbacks_sonar_example.md
sarahmarshy: Might be a question for @AnotherButler. Would you prefer a link to a repo or a markdown file with code?

docs/refs/api/rtos/callbacks.md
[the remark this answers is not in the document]
kegilbert: Serial attach uses callbacks: https://github.com/ARMmbed/mbed-os/blob/master/drivers/SerialBase.h#L102

Agree on the blurb, it is a bit abrupt. How's something like

> For example, the following code allows a user to provide a customized response whenever the serial line receives data. 

?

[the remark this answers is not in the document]
geky: That sounds gooder

docs/refs/api/rtos/callbacks_sonar_example.md
sarahmarshy: https://developer.mbed.org/teams/mbed_example/code/callback-sonar-example/

geky:
LGTM!

docs/refs/api/platform/callbacks_api.md
```diff
@@ -0,0 +1,19 @@
...
+
+ ###### Thread example with callbacks
+
+The Callback API provides a convenient way to pass arguments to spawned threads.
+
+ [![View code](https://www.mbed.com/embed/?url=https://developer.mbed.org/teams/mbed_example/code/rtos_threading_with_callback/)](https://developer.mbed.org/teams/mbed_example/code/rtos_threading_with_callback/file/d4b2a035ffe3/main.cpp)
+
+###### Serial passthrough example with callbacks
+[![View code](https://www.mbed.com/embed/?url=https://developer.mbed.org/teams/mbed_example/code/rtos_threading_with_callback/)](https://developer.mbed.org/users/mbedAustin/code/SerialPassthrough/file/96cb82af9996/main.cpp)
+
```
kegilbert: Should we put the sonar example here too now since it's hosted on the developer site?

docs/refs/api/platform/callbacks_api.md
```diff
@@ -8,12 +8,11 @@ This is the technical reference for callbacks. You should read the [introduction
...
 
+###### Serial passthrough example with callbacks
+[![View code](https://www.mbed.com/embed/?url=https://developer.mbed.org/teams/mbed_example/code/rtos_threading_with_callback/)](https://developer.mbed.org/users/mbedAustin/code/SerialPassthrough/file/96cb82af9996/main.cpp)
+
  ###### Thread example with callbacks
 
 The Callback API provides a convenient way to pass arguments to spawned threads.
 
  [![View code](https://www.mbed.com/embed/?url=https://developer.mbed.org/teams/mbed_example/code/rtos_threading_with_callback/)](https://developer.mbed.org/teams/mbed_example/code/rtos_threading_with_callback/file/d4b2a035ffe3/main.cpp)
-
```
kegilbert: Is this being moved somewhere else or are we deleting it completely?

kegilbert: Scratch that, misread the commit.

kegilbert:
LGTM!

PR #215, opened 2017-09-05
update mesh section

mbartling:
Moved bulk of mesh section to top level. Simplified api page to look like other api docs. Need to move examples to hosted repos in dev site

@sg- There are a few examples that are inlined right now and I think these should be hosted somewhere as stand alone examples. Where would you like me to put them so you guys can have control over them?


sg-:
the only inline should be from mesh-minimal-example

mbartling:
Sorry, let me rephrase. From a users' perspective, the main code will still appear inlined but the actual code will reside in a separate repo. This is preferable to having an example embedded directly in the md file

PR #209, opened 2017-08-31
Porting guide: Add Sleep API pg page

bulislaw:
Status: *Work in progress*

It's basically Martin's proposal.

@c1728p9 @0xc0170

PR #234, opened 2017-09-07
Add sleep API documentation

0xc0170:
- sleep API
- sleep manager (new in mbed OS 5.6)

This depends on https://github.com/ARMmbed/mbed-os/pull/4912 , as this one gets to 5.6, this should be part of 5.6 documentation. I noticed all PR are to the new_engine branch, this also then.

The aim is to describe how sleep API should work, new API for sleep manager API (use `DeepSleepLock` class in applications). 

An example there was inspired based on SPI changes we did, we can either point directly to the file, or keep this, or any other suggestions?

### DIFF
--- a/docs/refs/api/platform/sleep.md
+++ b/docs/refs/api/platform/sleep.md
@@ -1,0 +1,102 @@
+#### Sleep
+
+There is only one sleep function in the Mbed OS 5.6
+
+```c++
+void sleep();
+```
+
+This function invokes Sleep manager that is introduced below.
+
+It is invoked in the idle loop by default. This default behaviour can be overwritten by attaching a different the idle hook function pointer.
+
+
+```c++
+void new_idle_loop()
+{
+    // do nothing
+}
+
+void main()
+{
+    rtos_attach_idle_hook(&new_idle_loop);
+}
+```
+
+##### Sleep modes
+
+There are two available sleep modes:
+
+1. Sleep mode
+
+    The system clock to the core is stopped until a reset or an interrupt occurs. This eliminates dynamic power used by the processor, memory systems and buses. The processor, peripheral and memory state are maintained, and the peripherals continue to work and can generate interrupts.
+
+    The processor can be woken up by any internal peripheral interrupt or external pin interrupt.
+
+    The wake-up time should be less than 10 us.
+
+2. Deep sleep mode 
+
+    This processor is setup ready for deep sleep, and sent to sleep. This mode
+    has the same sleep features as sleep plus it powers down peripherals and high speed clocks. All state is still maintained. 
+    The processor can only be woken up by lp ticker, RTC, an external interrupt on a pin or a watchdog timer.
+
+    The wake-up time should be less than 10 ms.
+
+##### Latency
+
+The sleep modes introduce some interrupt + wake-up latencies that might affect the application. The times are as follows:
+
+Sleep latency - up to 10 microseconds
+Deep sleep latency - up to 10 milliseconds
+
+##### Sleep manager
+
+The sleep manager provides API to control sleep modes. Deep sleep might introduce some power savings that can affect an application, for instance high speed clock dependent drivers.
+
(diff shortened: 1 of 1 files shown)


0xc0170:
There's https://github.com/ARMmbed/Handbook/pull/209, similar. This one though is about API reference,  the other one is for contributing (portin sleep hal), I would assume.

0xc0170:
cc @c1728p9 @bulislaw @sg-

bulislaw:
Looks good, but that should be against the 5.5 branch as we are merging sleep manager for 5.6.

CC: @AnotherButler

AnotherButler:
@0xc0170 I've made some copy edits. Please review and make sure I didn't change the content. Also, this is the right branch.

bulislaw:
@AnotherButler that means we are switching to the new engine for 5.6?

0xc0170:
I updated the PR, removed sleep API changes taht will come later. 

@AnotherButler What branch should I use? can you please edit and switch it? I'll be away for couple of days, and this should get in. feel free to edit,  to make it integrated.

AnotherButler:
@bulislaw Yes. That is correct.

Also, we're standardizing our APIs. If this is an API and not related to porting, we need to link to the class reference Doxy and have examples in repos instead of hardcoded into the markdown. Can you help with this?

bulislaw:
I'm not sure how to reference a doxy, I'm sure I could figure it, but this wasn't released yet so it probably doesn't have a doxy online.

AnotherButler:
OK. I'll figure out how to build the Doxy, and then I'll embed the code snippet from the Doxy. Do you have time to update the example code, so it's not hardcoded into the markdown?

bulislaw:
But the example doesn't exists as a stand alone program (it's not mbed-os-example-XXX), it's just two snippets of relevant code from mbed os internals. We could potentially link to this file, but I think it wouldn't be as clear, not to mention that the file is evolving all the time so we will have issues maintaining that.

bulislaw:
The example is abstract, it's not an example app in stand alone sense. It's not something you can use out of context and you can't import it anywhere, it's just an illustration of how this sequence of calls should look like. I don't want to include specific driver's code as it's internals can change and things can be moved around, but this hardcoded abstract example will always be correct.

AnotherButler:
@sg- I think this doc should make the standardization we've been working on, or we should move it to a different location because I think our APIs should all have importable examples.

bulislaw:
I would say our examples should be useful in the first place if it makes sense to have them as programs than I agree, in this case it's example of usage from perspective of internal driver.
This PR is based on what @0xc0170 proposed, Martin what do you think? Does it make sense to create dummy example?

AnotherButler:
@bulislaw Don't worry about it in this case. In the future, please try to avoid hardcoding any examples in markdown that may require updating. Is this PR ready to merge, or does it still require content review/changes?

bulislaw:
I think it's ready.

AnotherButler:
@bulislaw We need to embed the class reference Doxy for this before I merge. Do you know the link?

bulislaw:
Search on online doxy doesn't give me back any results, does it need to be regenerated? We are looking for these headers: https://github.com/ARMmbed/mbed-os/blob/master/platform/mbed_sleep.h https://github.com/ARMmbed/mbed-os/blob/master/platform/DeepSleepLock.h

AnotherButler:
Perfect, thanks. I can find what should be the right link from there. I'll embed it and merge.

docs/refs/api/platform/sleep.md
```diff
@@ -0,0 +1,92 @@
...
+- `Timeout`
+- `Timer`
+- `SPI`
+- `I2C`
+- `CAN`
+- `SerialBase`
+
+### API
+
+[![View code](https://www.mbed.com/embed/?type=library)](https://docs.mbed.com/docs/mbed-os-api/en/mbed-os-5.6/api/DeepSleepLock_8h_source.html)
```
bulislaw: I think we should also reference this header:
https://github.com/ARMmbed/mbed-os/blob/master/platform/mbed_sleep.h

bulislaw:
Looks great, thanks!

PR #255, opened 2017-09-16
Refactor to reference book

sg-:
Lots of work to do but need this rendered on the website to see
how it fits together. TODO work should be corrected in the future
but not block this patch.


AnotherButler:
@sg- Thanks for all the PR. I'd prefer not to have TODOs on the docs site. Can you please move the TODOs into a separate PR, so they don't block the merging of this? Also, the engine is still down, so we can't see what it looks like just yet.

sg-:
I'd prefer to leave them for now so we get a sense of whats needed to be done and how it all fits together. This will only be for internal viewing and will remove before public accessibility.

PR #259, opened 2017-09-18
Update configuration section

sg-:


AnotherButler:
Note to self: Merging placeholders for now, so we can what's expected.

PR #269, opened 2017-09-25
Rewrite toolchain configuration in preference order

theotherjimmy:
And add a section about the priority between multiple toolcahin
configurations

### DIFF
--- a/docs/tools/offline/cli.md
+++ b/docs/tools/offline/cli.md
@@ -74,51 +74,61 @@ Mbed CLI requires adding the following to the system `PATH`:
 
 #### Mandatory: Toolchain selection
 
-You need to tell Mbed CLI where to find the toolchains that you want to use for compiling your source tree.
+You need to tell Mbed CLI where to find the toolchains that you want to use for compiling. The following toolchains are supported:
 
-There are multiple ways to configure toolchain locations:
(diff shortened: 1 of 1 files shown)


docs/tools/offline/cli.md
```diff
@@ -74,51 +74,61 @@ Mbed CLI requires adding the following to the system `PATH`:
...
 ```
 $ mbed config -G ARM_PATH "C:\Program Files\ARM"
 [mbed] C:\Program Files\ARM now set as global ARM_PATH
 ```
+The `-G` switch tells Mbed CLI to set this as a global setting, allowing all projects owned by this user to compile with Arm Compiler 5. You may instead set the `ARM_PATH` using a project local setting by omitting the `-G` command line switch.
+
+Mbed CLI supports a setting for each toolchain path. These setting are listed below, along with a description of what path is expected in each setting.
 
-The `-G` switch tells Mbed CLI to set this as a global setting, rather than local for the current program.
+* `ARM_PATH`: The path to the *base* directory of your Arm Compiler installation. This should be the directory containing the directory containing the binaries for `armcc` and friends.
```
AnotherButler: Query: The directory containing the binaries? Or the directory containing the directory containing the binaries?

theotherjimmy: For this one, if the binary is `<some-path>/arm/bin/armcc` then the path should be `<some-path>/arm`. Do you think an example would help here?

AnotherButler: Yes

theotherjimmy:
@AnotherButler I added the examples as requested.

PR #281, opened 2017-09-29
Add filesystem example to storage reference

sarahmarshy:
@sg- @AnotherButler 

Inspired by this question - https://os.mbed.com/questions/79128/nRF52-DK-Store-data-in-flash/

Though there is no way to store to internal flash, it still seems like there should be an example in the relevant docs page - https://os.mbed.com/docs/v5.6/reference/storage-overview.html


AnotherButler:
@sg- bump

sg-:
This file seems confused. It should be an introduction to storage but it seems to blend 
API usage, porting and example programs. I dont think the example makes this any better.

sarahmarshy:
@sg- I agree. Admittedly, this was a bit of a quick fix to at least show a working code sample in the reference section.

PR #285, opened 2017-10-09
Clarify using mbed CLI vs compiler for custom targets

sarahmarshy:
@theotherjimmy 
@AnotherButler


AnotherButler:
I made a few minor grammar changes. If they look good to you, I will merge.

PR #286, opened 2017-10-10
Add API specification for RawSerial class

kegilbert:
I didn't want to pull over and copy all of the Serial examples as that wouldn't add much. Tried to show examples closer to the use case of RawSerial, let me know if you think it's not sufficient or should be changed to something else.

@AnotherButler 
@sg-

### DIFF
--- a/docs/reference/api/drivers/RawSerial.md
+++ b/docs/reference/api/drivers/RawSerial.md
@@ -1,1 +1,34 @@
 ## RawSerial
+
+The RawSerial interface provides UART functionality without the use of streams like the Serial class. This makes it suitable for use in interrupt handlers with the RTOS.
+
+### RawSerial class reference
+
+[![View code](https://www.mbed.com/embed/?type=library)](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/classmbed_1_1_raw_serial.html)
+
+<span class="notes">**Note**: On a Windows machine, you will need to install a USB serial driver. See [Windows serial configuration](/docs/v5.6/tutorials/serial-communication.html#windows-serial-driver).</span>
+
+Serial channels have a number of configurable parameters:
+
+  - _Baud Rate_ - There are a number of standard baud rates ranging from a few hundred bits per second, to megabits per second. The default setting for a serial connection on the Mbed microcontroller is 9600 baud.
+  - _Data length_ - Data transferred can be either 7 or 8 bits long. The default setting for a serial connection on the Mbed microcontroller is 8 bits.
+  - _Parity_ - You can add an optional parity bit. The parity bit will be automatically set to make the number of 1s in the data either odd or even. Parity settings are *Odd*, *Even* or *None*. The default setting for a serial connection on the Arm Mbed microcontroller is None.
+  - _Stop Bits_ - After data and parity bits have been transmitted, one or two stop bits are inserted to "frame" the data. The default setting for a serial connection on the Mbed microcontroller is one stop bit.
+
+The default settings for the Mbed microcontroller are described as _9600-8-N-1_, a  common notation for serial port settings.
+
+### RawSerial hello, world
+
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/RawSerial_HelloWorld/)](https://os.mbed.com/teams/mbed_example/code/RawSerial_HelloWorld/file/112a40a5991a/main.cpp)
+
+### RawSerial examples
+
+#### Example one
+
+Write a message to a device at a baud rate of 19200.
+
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/RawSerial_ex_1/)](https://os.mbed.com/teams/mbed_example/code/RawSerial_ex_1/file/6a0d9cb21969/main.cpp)
+
+#### Mbed OS Example
+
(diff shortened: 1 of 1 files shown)


docs/reference/api/drivers/RawSerial.md
```diff
@@ -1 +1,34 @@
 ## RawSerial
+
+The RawSerial interface provides UART functionality without the use of streams like the Serial class. This makes it suitable for use in interrupt handlers with the RTOS.
```
AnotherButler: Query: Does this mean the Serial class is a stream, or the Serial class provides UART functionality without the use of streams?

docs/reference/api/drivers/RawSerial.md
```diff
@@ -1 +1,34 @@
...
+
+Serial channels have the following configurable parameters:
+
+  - _Baud Rate_ - Standard baud rates range from a few hundred bits per second to megabits per second. The default setting for a serial connection on the Mbed microcontroller is 9600 baud.
+  - _Data length_ - Transferred data can be either 7 or 8 bits long. The default setting for a serial connection on the Mbed microcontroller is 8 bits.
+  - _Parity_ - You can add an optional parity bit. The parity bit will be automatically set to make the number of 1s in the data either odd or even. Parity settings are *Odd*, *Even* or *None*. The default setting for a serial connection on the Arm Mbed microcontroller is None.
+  - _Stop Bits_ - After data and parity bits have been transmitted, one or two stop bits are inserted to "frame" the data. The default setting for a serial connection on the Mbed microcontroller is one stop bit.
```
AnotherButler: Query: Who or what is transmitting the data and parity bits? Also, who or what is inserting the stop bits to "frame" the data? Please clarify for active voice.

docs/reference/api/drivers/RawSerial.md
kegilbert: The RawSerial class provides UART functionality without being a stream, the Serial class inherits from the Stream class.

docs/reference/api/drivers/RawSerial.md
kegilbert: These are configurable settings for the RawSerial object, the transmitting is being done by that object on calls to write/putc. 

Would something like 
>After data and parity bits have been transmitted, the RawSerial object inserts one or two stop bits to "frame" the data.

be preferable?

AnotherButler: Yes. 
Also, do we know who or what is transmitting the data and parity bits? If not, maybe we could do "After the transmission of data and parity bits, the RawSerial object inserts one or two stop bits to "frame" the data."?

AnotherButler: In parity, please change it to "The object automatically sets the parity bit to make the number of..."

docs/reference/api/drivers/RawSerial.md
```diff
@@ -1 +1,34 @@
 ## RawSerial
+
+The RawSerial class provides UART functionality without the use of streams like the Serial class. This makes it suitable for use in interrupt handlers with the RTOS.
```
sg-: stream is out of context (and should be Stream - maybe with a link??). It doesn't re-target against the standard library. Probably better to say it partially re-implements scan and print family functions (need to fact check how true this is)

kegilbert: Serial calls into Stream's printf which calls into vfprintf to print through an output stream. RawSerial re-implements printf using vsprintf to write the formatted string to a buffer and writes to serial using _base_putc() -> target's serial_putc definition so looks like it checks out. 

Updated the phrasing and added a link to the Stream class.

docs/reference/api/drivers/RawSerial.md
```diff
@@ -1 +1,34 @@
 ## RawSerial
+
+The RawSerial class provides UART functionality without the use of [Stream's](https://github.com/ARMmbed/Handbook/blob/new_engine/docs/reference/api/platform/Stream.md)  print and scan functions as in the Serial class. Instead of using the standard library print and scan functions they are re-implemented to work using each target's underlying serial communication functions. See the porting guide for [target serial support](https://os.mbed.com/docs/v5.6/reference/contributing-target.html#serial). This makes it suitable for use in interrupt handlers with the RTOS.
```
AnotherButler: I don't think I understand the second sentence. Who or what uses the standard library print and scan functions? Who or what is reimplementing the print and scan functions to work using each target's underlying serial communication functions? Also, is Stream part of the Serial class?

kegilbert: Sorry sort of wrote that after looking through Serial and RawSerial implementations for a while and it made sense in my head at the time. I'll rephrase it.

kegilbert: Updated it to hopefully be more clear, let me know if it's not. 

The Serial class inherits from the Stream class, yes. Serial uses the Stream class's print and scan functions as a child.

AnotherButler:
@kegilbert Please review my rephrasing, and make sure the meaning is still correct.

kegilbert:
Rephrasing looks better, thanks!

sg-:
This is looking really good! I'm a little confused about the configuration section so want to make sure we use it to drive consistency. There are a few ways to configure something:
- compile time config (this isnt the place for that IMO)
- constructor (seems what we're documenting)
- run time options not in the constructor

What is the plan with this section?

kegilbert:
@sg- @AnotherButler 
Spoke briefly with Sam, he had some good points about what should go in the configuration section. Aiming to keep that section fairly stripped down and highlight what settings can be configured during construction and runtime for users trying to jump in. The class reference following this section contains much more detailed information.

Let me know your thoughts on the updated layout.

PR #287, opened 2017-10-10
Add HeapBlockDevice API docs

kegilbert:
@AnotherButler 
@sg-

### DIFF
--- a/docs/reference/api/storage/HeapBlockDevice.md
+++ b/docs/reference/api/storage/HeapBlockDevice.md
@@ -1,2 +1,20 @@
 ## HeapBlockDevice
 
+The HeapBlockDevice class provides a way to simulate block devices for software development or testing. The created blocks are nonvolatile, they will not persist across power cycles.
+
+### HeapBlockDevice Class Reference
+
+[![View code](https://www.mbed.com/embed/?type=library)](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/class_heap_block_device.html)
+
+HeapBlockDevices have some notable configurable parameters:
+
+  - _size_ - Size of the Block Device in bytes
+  - _read_size_ - Minimum read size required in bytes
+  - _program_size_ - Minimum program size required in bytes
+  - _erase_size_ - Minimum erase size required in bytes
+  
+### HeapBlockDevice Example
+
+Create a HeapBlockDevice, program it, read the block back, and cleanup.
+
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/HeapBlockDevice_ex_1/)](https://os.mbed.com/teams/mbed_example/code/HeapBlockDevice_ex_1/file/5991e7053465/main.cpp)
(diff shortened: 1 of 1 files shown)


kegilbert:
LGTM, thanks!

docs/reference/api/storage/HeapBlockDevice.md
```diff
@@ -1,2 +1,20 @@
 ## HeapBlockDevice
 
+The HeapBlockDevice class provides a way to simulate block devices for software development or testing. The created blocks are nonvolatile; they do not persist across power cycles.
+
+HeapBlockDevices have some notable configurable parameters:
+
+  - _size_ - Size of the block device in bytes.
```
sg-: How are these configured? Is this just the constructor? Why are they notable?

kegilbert: Correct, they're constructor parameters. Wanted to list them out up front, but they are listed in attached class reference links as well. The phrasing is probably a bit off, they're not so much notable parameters as much as a list of parameters used to configure the HeapBlockDevice. I can change the wording or pull that list out.

AnotherButler: We have a planned section for configuration options for storage APIs (https://jira.arm.com/browse/IOTMORF-1273). However, this planned page should be generated JSON. Would the generated JSON include this content?

kegilbert:
@sg- @AnotherButler 
Follow up changes to emulate the changes in the RawSerial API configuration section.

docs/reference/api/storage/HeapBlockDevice.md
```diff
@@ -1,2 +1,31 @@
 ## HeapBlockDevice
 
+The HeapBlockDevice class provides a way to simulate block devices for software development or testing. The created blocks are nonvolatile; they do not persist across power cycles.
+
+HeapBlockDevices have the following configurable parameters in either one of two constructors:
```
AnotherButler: Query: What are the two constructors?

kegilbert: The second block that's talking about creating a block device using a block parameter to set the erase, program, and read sizes is the second constructor. The four parameters above that refer to the first constructor.

kegilbert: @sg- Added subheaders to break out constructors, are the titles good or should that be rephrased?

sg-:
Seems ok but now I dont understand when to use one or the other or even why both exist...

kegilbert:
@sg- Rephrased it a bit, as far as I can tell the purpose of the first constructor is to just allow more control over erase, program, and read sizes. If they are all bound by the block size then you can just specify that parameter and use the second constructor.

sg-:
Why would there be 2 constructors and not just one with default parameters?

kegilbert:
@sg- Mainly for convenience, having the shortened constructor with just the block_size parameter is easier to use for a lot of block devices that have all three of those parameters restricted to said block size. Using just the verbose constructor with default parameters would be confusing if the only bit of relevant data a user had was block size.

kegilbert:
@sg- Was there anything you wanted changed in this doc?

PR #288, opened 2017-10-10
Verify that "id"s are unique in Travis CI

theotherjimmy:
You need to turn on Travis CI integration to make this run

Special character soup!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+script:
+# Assert that we have not added any duplicate "id"s. The bash snippet below
+# uses grep to extract any headers with id elements from the text, removing
+# everything that is not the contents of the id feild, and then sorting.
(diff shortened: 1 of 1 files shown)


AnotherButler (approved):

AnotherButler:
@iriark01 

@theotherjimmy wrote a script for us that will check all unique heading IDs any time a PR goes in to make sure that we don't duplicate them. He says it will only take about a minute to run. He can also do it for the Cloud docs if you like it.

iriark01:
How far up do they need to be unique? @micque01 can headers repeat across books?

theotherjimmy:
For what it's worth, this checks for duplicate ids in this repository at the moment.

micque01:
I think it is fine if a header/id is repeated in a different book. I would try to avoid duplicating it in the same book though

PR #290, opened 2017-10-12
Reorder content on Windows installer for Mbed CLI

MarceloSalazar:
This is a proposal to improve the instructions and message for Windows users that want to install Mbed CLI. Summary:

- Make users aware of Windows installer at the very beginning before they see advance/detailed instructions.
- Rename Mbed CLI for Windows, to Windows installer for Mbed CLI, as the previous looked a bit awkward to me.
- For advanced users that need to use Linux and Mac OS X, we can point them at the 'Installing Mbed CLI standalone stable version' section.

Pending
- Add links in the info banner.

@AnotherButler please have a look and let me know your thoughts


docs/tools/offline/cli-setup.md
```diff
@@ -3,6 +3,8 @@
 
 You can install Mbed CLI on Windows, Linux and Mac OS X.
 
+<span class="tips">If you're working on Windows, we recommend you use the [Windows installer for Mbed CLI](LINK TBC). For Linux and Mac OS X, refer to the section [Installing Mbed CLI standalone stable version](LINK TBC)</span>
```
AnotherButler: It's weird to me that this is before requirements.

docs/tools/offline/cli-setup.md
```diff
@@ -31,37 +33,11 @@ Mbed CLI is compatible with [Virtual Python Environment (virtualenv)](https://py
...
-$ python setup.py install
-```
-
-On Linux or Mac, you may need to run with `sudo`.
-
-#### Mbed CLI for Windows installer
+### Windows installer for Mbed CLI
 
-Mbed CLI for Windows installs [Mbed CLI](https://github.com/ARMmbed/mbed-cli) with all requirements on Windows 7 and newer (both the 32- and 64-bit versions).
+[Mbed CLI](https://github.com/ARMmbed/mbed-cli) is also bundled with a Windows installer that includes all requirements for Windows 7 and newer (both the 32- and 64-bit versions).
```
AnotherButler: Query: Who or what bundled Mbed CLI with the Windows installer? Please clarify for active voice.

docs/tools/offline/cli-setup.md
MarceloSalazar: Do users really need to know the requirements for Mbed CLI? (python, git, etc?)

I'm concerned about them being overwhelmed by so many requirements, which are resolved by using a single Windows installer package. The only requirement for the Windows installer is having Windows 7, 8 or 10.

docs/tools/offline/cli-setup.md
MarceloSalazar: The Windows installer includes Mbed CLI (and other things).
This may require a bit more discussion in terms of the name for the product.
Maybe use [Mbed CLI Windows Installer](https://github.com/ARMmbed/mbed-cli-windows-installer)
FYI @arekzaluski

docs/tools/offline/cli-setup.md
AnotherButler: I would say users installing Mbed CLI need to know any requirements or prerequisites to using it, yes.

MarceloSalazar: Amanda, IMO people that want to download the Windows Installer for Mbed CLI to get started with Mbed don't need to worry about too many details - for example, if you want to download Skype, you only care about getting the compatible version for your Windows PC, not about the libraries/apps that get installed as part of that (although it's ok to mention in release notes, etc).

The key section **Install Mbed CLI for Windows installer** containing a link to the .exe is buried in the middle of the doc and the link is not clearly visible. This is not great for new users.
I believe we need either a new section (at the same level as Setup), or bring the topic earlier with a banner as I'm suggesting. We're running many workshops and need to point people to the page for the Windows installer - again not great to have too much info for a new starter.

AnotherButler: OK. That makes sense to me. Thanks for the comparison. That helps.

AnotherButler:
@MarceloSalazar I've fixed the merge conflicts. Do you think this is ready to merge?

MarceloSalazar:
Thanks @AnotherButler! Looks good.
There is one more thing: we are mixing 'Mbed CLI for Windows installer' and 'Windows installer for Mbed CLI'; my preference is the latter as sounds more natural. Could you have a look and possibly fix this? Thanks!

AnotherButler:
Good catch.

MarceloSalazar:
Thanks! Looks good now!

PR #293, opened 2017-10-17
Fill out mbed-ls "contributing" guide

theotherjimmy:
I had to mess with the template provided as partners do not implement Mbed-ls
support, and as such it would really be hard for them to have bugs when
implementing it. The section titles need to be revised as they do not map well
into the way I would describe "making mbed-ls detect a new target".

### DIFF
--- a/docs/reference/contributing/tools/mbedls.md
+++ b/docs/reference/contributing/tools/mbedls.md
@@ -4,7 +4,7 @@ Mbed-ls is a Python module that detects and lists Mbed Enabled boards connected 
 
 ### Defined behavior
 
-Mbed-ls needs the following information to correctly detect an Mbed Enabled target:
+Mbed-ls needs the following information to correctly detect an Mbed Enabled board:
 
  - A four-digit hexadecimal identifier of a board class.
  - A vendor string for detection on Windows.
--- a/docs/reference/contributing/tools/mbedls.md
+++ b/docs/reference/contributing/tools/mbedls.md
@@ -1,4 +1,4 @@
-### mbed-ls
+### Mbed-ls
 
 Mbed-ls is a Python module that detects and lists Mbed Enabled boards connected to a host computer.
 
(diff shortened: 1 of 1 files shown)


AnotherButler:
@sg- Does this look good from a content perspective?

docs/reference/contributing/tools/mbedls.md
```diff
@@ -1,21 +1,21 @@
-### mbed-ls
+### Mbed-ls
 
-[Add description here.]
+Mbed-ls is a Python module that detects and lists Mbed Enabled boards connected to a host computer.
```
sg-: host computer or just a computer?

docs/reference/contributing/tools/mbedls.md
```diff
@@ -1,21 +1,21 @@
...
+Further, Mbed-ls only detects devices that meet the following criteria:
 
-##### Undefined behavior
+ - Devices have a mass storage device (MSD) endpoint.
+ - The MSD endpoint has a FAT file system.
+ 
+### Implementation
 
-[Add undefined behavior. A bulleted list is fine.]
+When adding a target to Mbed-ls, you typically need to add a new definition to the platform database. You can add a new target to the platform database by adding entry mapping from your platform's four-digit hexadecimal identifier to the name of the board to the `DEFAULT_PLATFORM_DB` variable in [`mbed_lstools/platform_database.py`](https://github.com/ARMmbed/mbed-ls/blob/master/mbed_lstools/platform_database.py).
```
sg-: platform replaced with board. Where do you get a board id from?

docs/reference/contributing/tools/mbedls.md
```diff
@@ -1,21 +1,21 @@
...
 
-##### Potential bugs
-
-[Add potential bugs. A bulleted list is fine.]
-
-#### Implementing mbed-ls
-
-[Add explanation of how to implement.]
+You may also need to add a new vendor string for detection on windows. If you need to add a new vendor string, add it to the `usb_vendor_list` member of the `MbedLsToolsBase` class in [`mbed_lstools/lstools_base.py`](https://github.com/ARMmbed/mbed-ls/blob/master/mbed_lstools/lstools_base.py).
```
sg-: a bit vauge. Something like if your boards interface firmware is not based on [list of supported interface firmwares] then you need to do this...

docs/reference/contributing/tools/mbedls.md
theotherjimmy: I suppose you would like to remove "host". I'll do that.

docs/reference/contributing/tools/mbedls.md
theotherjimmy: "platform database" is the thing in mbed-ls. I will change "your platform's four-digit..." to "your board's four-digit..." and add where to find the "four-digit hexadecimal identifier". It's the "Unique product code" on the os.mbed.com website which you should receive through mbed enabled.

docs/reference/contributing/tools/mbedls.md
theotherjimmy: I think it's Daplink, Daplink derivatives and Segger.

theotherjimmy: Hm, STLink too.

theotherjimmy:
@sg- Thanks for the review. have I addressed all of your comments?

PR #297, opened 2017-10-18
SocketAdress Class doc update

deepikabhavnani:
Added an example to fetch ip address and port using socket address.

@sg- Please review


deepikabhavnani:
@sg-  Please review

PR #304, opened 2017-10-24
Extend runtime exectution page

bulislaw:
@sg- @AnotherButler

### DIFF
--- a/docs/reference/runtime/Execution.md
+++ b/docs/reference/runtime/Execution.md
@@ -2,14 +2,12 @@
 
 ### Threads
 
-Your application (`main` function) will start execution in the main thread, but it's not the only thread running in Mbed OS. There's a number of live system threads at any given time. They should be transparent from the user's point of view for most use cases. All threads that exist in the system by default:
-* main - default thread that executes application's `main` function.
+Your application (`main` function) will start execution in the main thread, but it's not the only thread running in Mbed OS. There's a number of threads running system services, such as:
+* main - default thread that executes application's `main` function. The main thread has 4kB of stack space by default, it can be configured by the application in mbed_app.json by defining MAIN_STACK_SIZE parameter.
 * idle - thread that's run by the scheduler when there's no other activity in the system (e.g. all other threads are waiting for some event). It's used to make sure the board is not burning empty processor cycles, but is put to sleep for as long as possible.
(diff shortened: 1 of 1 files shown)


docs/reference/runtime/Execution.md
```diff
@@ -1,6 +1,25 @@
 ## Execution
 
-You can use the same RTOS API in ISR. The only two warnings are:
+### Threads
+
+Your application (`main` function) will start execution in the main thread, but it's not the only thread running in Mbed OS. There's a number of live system threads at any given time. They should be transparent from the user's point of view for most use cases. All threads that exist in the system by default:
+* main - default thread that executes application's `main` function.
```
sg-: main thread - ... int main void is the main system thread and has stack size of ... by default

docs/reference/runtime/Execution.md
```diff
@@ -1,6 +1,25 @@
...
+* idle - thread that's run by the scheduler when there's no other activity in the system (e.g. all other threads are waiting for some event). It's used to make sure the board is not burning empty processor cycles, but is put to sleep for as long as possible.
```
sg-: idle thread - comment about centralized sleep management?

docs/reference/runtime/Execution.md
```diff
@@ -1,6 +1,25 @@
...
+* timer - thread for handling system and user timer objects.
```
sg-: Is this really still used? Can we say its depreciated with the RtosTimer class?

bulislaw: Good point will add that.

docs/reference/runtime/Execution.md
bulislaw: By centralised sleep management you mean?

docs/reference/runtime/Execution.md
bulislaw: I've added that as a note after, but it fits here better, I'll move that info inline.

AnotherButler:
@bulislaw Please review my changes, and make sure I did not accidentally change the meaning of the content. Please pay special attention to the `main` bullet. I tried to clarify but may have misunderstood.

bulislaw:
@AnotherButler thanks for that, looks good!

PR #306, opened 2017-10-25
Add API docs for SlicingBlockDevice

kegilbert:
@AnotherButler 
@sg-

### DIFF
--- a/docs/reference/api/storage/SlicingBlockDevice.md
+++ b/docs/reference/api/storage/SlicingBlockDevice.md
@@ -1,6 +1,11 @@
 ## SlicingBlockDevice
 
-[Add description here.]
+The SlicingBlockDevice class provides a way to break up a block device into sub units without the need to manually track offsets. The SlicingBlockDevice acts as an opposite of the [ChainingBlockDevice](/docs/v5.6/reference/chainingblockdevice.html) class.
+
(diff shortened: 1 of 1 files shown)


docs/reference/api/storage/SlicingBlockDevice.md
```diff
@@ -1,11 +1,19 @@
...
+  - _end_ - End block address to mark the end of the block device. This block is not mapped; negative addresses are calculated from the end of the underlying block device.
 
 ### SlicingBlockDevice class reference
 
 [![View code](https://www.mbed.com/embed/?type=library)](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/class_slicing_block_device.html)
 
 ### SlicingBlockDevice example
 
-[Add example here.]
+This SlicingBlockDevice example partitions a [HeapBlockDevice](/docs/v5.6/reference/heapblockdevice.html) into three subunits and showcases programming and reading back data segments through both the underlying master block device and the sliced subunits.
```
sg-: not sure what the policy is on links to specific versions but in general I'm not a fan of linking but happy to be told otherwise.

kegilbert: Fair enough, I could pull the links out. I don't feel particularly strongly one way or the other.
@AnotherButler Would you rather have the links or no?

kegilbert: @AnotherButler The specific version issue should be worked around by your new release update script right?  Although we should probably be consistent on when we're linking and when we're not. If we reference every external class with a link I could see that getting out of hand.

PR #307, opened 2017-10-25
Add link to SlicingBlockDevice

kegilbert:
SlicingBlockDevice acts as an opposite for this class, add in reference to it.

# DEPENDENT ON https://github.com/ARMmbed/Handbook/pull/306

@AnotherButler 
@sg-

### DIFF
--- a/docs/reference/api/storage/ChainingBlockDevice.md
+++ b/docs/reference/api/storage/ChainingBlockDevice.md
@@ -1,6 +1,6 @@
 ## ChainingBlockDevice
 
-The ChainingBlockDevice class provides a way to chain together multiple block devices. You can then interact with the chained block devices as if they were a single block device of size equal to the sum of each substorage unit. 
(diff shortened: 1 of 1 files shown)


docs/reference/api/storage/ChainingBlockDevice.md
```diff
@@ -1,6 +1,6 @@
 ## ChainingBlockDevice
 
-The ChainingBlockDevice class provides a way to chain together multiple block devices. You can then interact with the chained block devices as if they were a single block device of size equal to the sum of each substorage unit. 
+The ChainingBlockDevice class provides a way to chain together multiple block devices. You can then interact with the chained block devices as if they were a single block device of size equal to the sum of each substorage unit. The ChainingBlockDevice acts as an opposite of the [SlicingBlockDevice](/docs/v5.6/reference/slicingblockdevice.html).
```
sg-: not sure the policy of linking and not a fan of linking but happy to be told otherwise. If our docs are well organized finding class names should be intuitive.

kegilbert: Fair enough, I could pull the links out. I don't feel particularly strongly one way or the other.

AnotherButler: I don't mind including internal links that keep our users within our documentation. As a user, I find them helpful. Even if I can go to the navigation bar and find the API easily, it's nice not to have to. However, I do think too many links on a page is problematic.

PR #308, opened 2017-10-25
Runtime memory page update

bulislaw:
@AnotherButler @sg-

### DIFF
--- a/docs/reference/runtime/Memory.md
+++ b/docs/reference/runtime/Memory.md
@@ -3,13 +3,13 @@
 This is a basic overview of the memory model in Mbed OS.
 
 ```
-+---------------------+   Last Address of RAM
-| Scheduler/ISR Stack |
++---------------------+   Last address of RAM
+| Scheduler/ISR stack |
 +---------------------+
 |          ^          |
 |          |          |
 |                     |
-|      Heap Cont.     |
+|      Heap cont.     |
 |---------------------|
 | User thread n stack |
 |---------------------|
@@ -26,18 +26,18 @@ This is a basic overview of the memory model in Mbed OS.
 | ZI: Global data     |
 |                     |
 +---------------------+
-| ZI: Idle Stack      |
+| ZI: Idle stack      |
 +---------------------+
-| ZI: Timer Stack     |
+| ZI: Timer stack     |
 +---------------------+
-| ZI: Main Stack      |
+| ZI: Main stack      |
 +---------------------+
 |                     |
 | ZI: Global data     |
 |                     |
 +---------------------+
-| RW: Vector Table    |
-+=====================+   First Address of RAM
+| RW: Vector table    |
++=====================+   First address of RAM
 |                     |   Last address of flash
 |                     |
 |     Application     |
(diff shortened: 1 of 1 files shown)


docs/reference/runtime/Memory.md
```diff
@@ -51,25 +51,25 @@ This is a basic overview of the memory model in Mbed OS.
...
 
-Inside RAM we can distinguish two logical types: static and dynamic memory. Each of them is used in different ways:
-* Static (zero initialized)
-  * Vector table
-  * Global data
-  * Static data
-  * Stacks for default threads (main, timer, idle, scheduler/ISR)
-* Dynamic
-  * Heap (dynamic data)
+Inside RAM, we can distinguish two logical types: static and dynamic memory. You can use each of them in different ways:
```
bulislaw: I think that 'you can use each of them' changes the meaning a bit. It's used by the system in this way rather than provides an option for the user.

AnotherButler: Ah. Good to know. Would "The system uses each of them in different ways:" be better?

bulislaw: yeah, that sounds great.

docs/reference/runtime/Memory.md
```diff
@@ -1,51 +1,75 @@
...
+|                     |
++---------------------+
+| RW: Vector table    |
++=====================+   First address of RAM
+|                     |   Last address of flash
+|                     |
+|     Application     |
+|                     |
+|                     |
++---------------------+
```
sg-: missing the RO vector table

docs/reference/runtime/Memory.md
```diff
@@ -1,51 +1,75 @@
...
+|                     |
+| Optional bootloader |
+|                     |
++---------------------+   First address of flash
```
sg-: missing the RO vector table??

docs/reference/runtime/Memory.md
```diff
@@ -1,51 +1,75 @@
...
-| RW                |  
-+===================+   First Address of RAM
-|                   |
-|                   |   Flash
+### RAM
 
-```
+Inside RAM, we can distinguish two logical types: static and dynamic memory. The system uses each of them in different ways:
+* Static (zero initialized):
+  * Vector table.
```
0xc0170: I am wondering if this is zero initialized?  It is usually in the own section in ROM, and once it is reallocated it is zero initalized? I belive just copied.

docs/reference/runtime/Memory.md
```diff
@@ -1,51 +1,75 @@
...
+* Dynamic:
+  * Heap (dynamic data).
+  * Stacks for user threads. Mbed OS will dynamically allocate memory on heap for user thread's stacks.
+
+Stack checking is turned on for all threads, and the kernel errors if it detects an overflow condition.
+
+### Flash
+
+Flash is a read only memory (ROM) that contains:
+* Application code.
```
0xc0170: As asked above, vector table might be here ?

docs/reference/runtime/Memory.md
bulislaw: Fixed

docs/reference/runtime/Memory.md
bulislaw: True.

docs/reference/runtime/Memory.md
bulislaw: Fixed

PR #314, opened 2017-10-27
Remove CallChain since it isn't public API

c1728p9:
Remove the page CallChain.md since this code is not part of the public API and is deprecated.


c1728p9:
Depends on https://github.com/ARMmbed/mbed-os/pull/5393

c1728p9:
Moved to #322

PR #315, opened 2017-10-27
Rework exporter contributing sections

theotherjimmy:


theotherjimmy:
Thanks for the edits. It's still accurate.

PR #317, opened 2017-10-28
Update bootstrap.md to contain more information

c1728p9:
Update the bootstrap page by adding information on linker script changes and adding newer information about the files cmsis.h, cmsis_nvic.h, mbed_rtx.h and targets.json.

### DIFF
--- a/docs/reference/contributing/target/bootstrap.md
+++ b/docs/reference/contributing/target/bootstrap.md
@@ -1,10 +1,64 @@
 ### Bootstrap
 
+#### Bring in CMSIS-Core files
+
 You need CMSIS-CORE files for startup and peripheral memory addresses, and you need linker scripts for Arm, IAR and GCC toolchains. These files are usually in the `mbed-os\targets\TARGET_VENDOR\TARGET_MCU_FAMILY\TARGET_MCUNAME\device` directory.
 
-Mbed OS requires dynamic vector relocation, which requires extensions to CMSIS-CORE. Extend CMSIS-CORE by adding an `mbed-os\targets\TARGET_VENDOR\TARGET_MCUNAME\cmsis.h` file. This header file defines the vector relocation additions and device-specific headers that include CMSIS-CORE. Next add a relocation function in `mbed-os\targets\TARGET_VENDOR\TARGET_MCUNAME\cmsis_nvic.c and .h`. This relocation function changes the contents of the Interrupt Vector Table at run time.
+#### Modify Linker scripts
 
-CMSIS-RTOS needs a few macros to initialize the SYS_TICK. These macros are usually configured in`mbed-os\targets\TARGET_VENDOR\mbed-rtx.h`.
+After adding the core files, the next step is to update the linker scripts for `mbed-os`. To do this, follow the steps below.
+
+* Reserve space for RAM vector table. For the example below this is:
+
+```
+RW_IRAM1 (0x20000000 + (VECTORS*4)) (0x30000 - (VECTORS*4))  {  ; RW data
+```
+
+* Define the start of the heap:
+ * ARM - The heap starts immediately after the region ```RW_IRAM1```.
+ * GCC_ARM - The heap starts at the symbol ```__end__```.
+ * IAR - The heap is the ```HEAP``` region.
+* Add defines for a relocatable application - ```MBED_APP_START``` and ```MBED_APP_SIZE```.
+* Add preprocessing directive ```#! armcc -E``` (ARM compiler only).
+
+Example linker script for the ARM compiler:
+
+```
+#! armcc -E
+
+#if !defined(MBED_APP_START)
+  #define MBED_APP_START 0x08000000
+#endif
+
+#if !defined(MBED_APP_SIZE)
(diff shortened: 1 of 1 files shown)


c1728p9:
@sg- can you review this?

docs/reference/contributing/target/bootstrap.md
```diff
@@ -1,10 +1,64 @@
...
+#### Bring in CMSIS-Core files
+
 You need CMSIS-CORE files for startup and peripheral memory addresses, and you need linker scripts for Arm, IAR and GCC toolchains. These files are usually in the `mbed-os\targets\TARGET_VENDOR\TARGET_MCU_FAMILY\TARGET_MCUNAME\device` directory.
 
-Mbed OS requires dynamic vector relocation, which requires extensions to CMSIS-CORE. Extend CMSIS-CORE by adding an `mbed-os\targets\TARGET_VENDOR\TARGET_MCUNAME\cmsis.h` file. This header file defines the vector relocation additions and device-specific headers that include CMSIS-CORE. Next add a relocation function in `mbed-os\targets\TARGET_VENDOR\TARGET_MCUNAME\cmsis_nvic.c and .h`. This relocation function changes the contents of the Interrupt Vector Table at run time.
+#### Modify Linker scripts
+
+After adding the core files, the next step is to update the linker scripts for `mbed-os`. To do this, follow the steps below.
+
+* Reserve space for RAM vector table. For the example below this is:
```
AnotherButler: Query: Do these steps need to happen in order?

c1728p9: Nope

c1728p9:
> We should add something about the startup file

@bulislaw, are there any modifications that need to be made to the CMSIS-Core startup file? I thought this was compatible with mbed-os.

> I would remove the Finishing section as that's not all you need to do.

Is there more to do to get the target to boot and run or are you just referring to the next porting steps?

> I would remove the point about targets.json as it doesn't belong here (and it's not helping really) we need separate section that would go in details about it.

This is required to get a new target building and booting.

c1728p9:
I rebased this PR and updated it so it has linker script templates for GCC and IAR. I kept the liker script templates simple with just one rom region and one ram region. Adding support for multiple regions and security bits is possible but will be a more involved task that is likely outside the scope of this docs update.

bulislaw:
> We should add something about the startup file
> @bulislaw, are there any modifications that need to be made to the CMSIS-Core startup file? I thought this was compatible with mbed-os.

I don't think so, but we don't even mention that it's needed or where it should go, it's required to boot.

> I would remove the Finishing section as that's not all you need to do.
> Is there more to do to get the target to boot and run or are you just referring to the next porting steps?
It's not a stand alone page but part of the porting guide, we will have a section on that, it doesn't need to be duplicated on every page.

> I would remove the point about targets.json as it doesn't belong here (and it's not helping really) we need separate section that would go in details about it.
> This is required to get a new target building and booting.

Again it's not a stand alone document, but part of porting guide.

docs/reference/contributing/target/bootstrap.md
```diff
@@ -1,10 +1,295 @@
...
+
+    /* Check if data + heap + stack exceeds RAM limit */
+    ASSERT(__StackLimit >= __HeapLimit, "region RAM overflowed with stack")
+}
+```
+
+#### Create and update files
+
+- Extend CMSIS-CORE by adding the file `mbed-os\targets\TARGET_VENDOR\TARGET_MCUNAME\cmsis.h`. This header file includes device-specific headers that include CMSIS-CORE.
+- Add a relocation function in `mbed-os\targets\TARGET_VENDOR\TARGET_MCUNAME\cmsis_nvic.h`. This contains the define `NVIC_NUM_VECTORS`, which is the number of vectors the devices has, and `NVIC_RAM_VECTOR_ADDRESS`, which is the address of the RAM vector table.
```
sg-: With vectors defined in the linker script, do we need cmsis_nvic.h anymore?

docs/reference/contributing/target/bootstrap.md
```diff
@@ -1,10 +1,295 @@
...
+- Define the initial stack pointer, `INITIAL_SP`, in `mbed_rtx.h`. This file is typically in `mbed-os\targets\TARGET_VENDOR\mbed_rtx.h`.
```
sg-: Whats the next step to get rid of mbed_rtx.h, extern a symbol from the .S startup file?

docs/reference/contributing/target/bootstrap.md
c1728p9: This is needed in both places. In the future it would be good to just use a value from the linker script.

docs/reference/contributing/target/bootstrap.md
c1728p9: If a heap + isr stack region is added to the linker script and used from the code then this could be removed.

c1728p9:
@bulislaw Sam is happy with this and I would like to get this merged. I disagree with the changes you suggested but it may be because I don't understand what you are looking for. Could you either post the changes to this PR or approve this PR and make the changes in a followup PR?

PR #318, opened 2017-10-28
Add documentation for the ConditionVariable class

c1728p9:
Add a description and example for the ConditionVariable class.


c1728p9:
This should go into 5.7 as it is a new feature. Once https://github.com/ARMmbed/mbed-os/pull/3648 is merged and dthe doxygen is available I'll update this to include the class reference.

c1728p9:
Will do. I held off on creating an example because this change isn't in mbed-os yet. @sg- can the example use master of mbed-os or does it need to be fixed to a release? If it needs to be on a release I won't be able to add an importable example until 5.7.

sg-:
It should be from a release so staging all this until 5.7 shouldn't be a problem.

PR #320, opened 2017-10-30
Add ProfilingBlockDevice API doc

kegilbert:
# DEPENDENT ON https://github.com/ARMmbed/mbed-os/pull/5395
The class reference link will be broken until the above PR is merged and the doxygen docs are regenerated.

@AnotherButler 
@sg-

### DIFF
--- a/docs/reference/api/storage/ProfilingBlockDevice.md
+++ b/docs/reference/api/storage/ProfilingBlockDevice.md
@@ -1,8 +1,8 @@
 ## ProfilingBlockDevice
 
-The ProfilingBlockDevice class provides a BlockDevice implementation to wrap around an existing block device object to log reads, writes, and erases.
+The ProfilingBlockDevice class provides a BlockDevice implementation to wrap around an existing block device object to log reads, writes and erases.
 
-ProfilingBlockDevices take in a pointer to the block device being profiled as the only configurable parameter. All block device operations that wish to be logged need to be performed through the ProfilingBlockDevice object. 
+ProfilingBlockDevices take in a pointer to the block device being profiled as the only configurable parameter. All block device operations that wish to be logged need to be performed through the ProfilingBlockDevice object.
 
 ### ProfilingBlockDevice class reference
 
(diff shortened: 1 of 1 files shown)


docs/reference/api/storage/ProfilingBlockDevice.md
```diff
@@ -1,2 +1,15 @@
 ## ProfilingBlockDevice
 
+The ProfilingBlockDevice class provides a BlockDevice implementation to wrap around an existing block device object to log reads, writes and erases.
```
AnotherButler: Content query: Is this too much like calling this a wrapper?

docs/reference/api/storage/ProfilingBlockDevice.md
```diff
@@ -1,2 +1,15 @@
...
+
+ProfilingBlockDevices take in a pointer to the block device being profiled as the only configurable parameter. All block device operations that wish to be logged need to be performed through the ProfilingBlockDevice object.
```
AnotherButler: Query: What does it mean that the operations need to be performed through the object? Please clarify. Also, please stop personifying operations. They don't like it when you do that.

kegilbert: In order for read, write, and erase calls to be counted, they need to be be called from the ProfilingBlockDevice object, not the underlying block device directly. 
i.e. given a block device object `bd`, and a profilingblockdevice object `profiler` pointing to `bd` you'd need to call 
```c++
profiler.read(....);
```
rather than
```c++
bd.read(...);
```

AnotherButler: Ah. Makes sense. Please do. Thanks. 👍

kegilbert:
@AnotherButler Rephrased the use case a bit.

docs/reference/api/storage/ProfilingBlockDevice.md
pan-: To be more precise the ProfilingBlockDevice class is a [decorator](https://en.wikipedia.org/wiki/Decorator_pattern) which adds profiling capabilities to a BlockDevice instance. 

@sg- What is our position on design pattern terminology ? Should we use it or avoid it ?

sg-: We should use it.

kegilbert:
@AnotherButler @pan- Updated the phrasing to use decorator.

PR #324, opened 2017-10-31
Add storage contributing overview doc

kegilbert:
Split the existing storage overview doc in the API chapter to form separate API and contributing overview pages. 

@AnotherButler @sg- 
Was this primarily what you were looking for? Let me know if you want more content or think anything should move between pages.

### DIFF
--- a/docs/reference/api/storage/storage.md
+++ b/docs/reference/api/storage/storage.md
@@ -7,15 +7,15 @@ The storage APIs present in Arm Mbed OS are:
 
 #### Declaring a file system
 
-The <a href="https://github.com/ARMmbed/mbed-os/blob/master/features/filesystem/FileSystem.h" target="_blank">FileSystem</a> class provides the core API for file system operations. You must provide a block device to back the file system. When you declare a file system with a name, you can open files on the file system through standard POSIX functions (see <a href="http://pubs.opengroup.org/onlinepubs/009695399/functions/open.html" target="_blank">open</a> or <a href="http://pubs.opengroup.org/onlinepubs/9699919799/functions/fopen.html" target="_blank">fopen</a>).
+The [FileSystem](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/class_file_system.html) class provides the core API for file system operations. You must provide a block device to back the file system. When you declare a file system with a name, you can open files on the file system through standard POSIX functions (see <a href="http://pubs.opengroup.org/onlinepubs/009695399/functions/open.html" target="_blank">open</a> or <a href="http://pubs.opengroup.org/onlinepubs/9699919799/functions/fopen.html" target="_blank">fopen</a>).
 
 Existing file systems:
 
-The <a href="https://github.com/ARMmbed/mbed-os/tree/master/features/filesystem/fat" target="_blank">FAT file system</a> is a standard file system.
+The [FAT file system](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/class_fat_file_system.html) is a standard file system.
 
-The <a href="https://github.com/ARMmbed/mbed-os/blob/master/features/filesystem/bd/BlockDevice.h" target="_blank">BlockDevice</a> class provides the underlying API for representing block-based storage that can be used to back a file system. Mbed OS provides standard interfaces for the more common storage media, and you can extend the BlockDevice class to provide support for unsupported storage.
+The [BlockDevice](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/class_block_device.html) class provides the underlying API for representing block-based storage that can be used to back a file system. Mbed OS provides standard interfaces for the more common storage media, and you can extend the BlockDevice class to provide support for unsupported storage.
 
-Additionally, two utility block devices give you better control over how storage is allocated. The <a href="https://github.com/ARMmbed/mbed-os/blob/master/features/filesystem/bd/SlicingBlockDevice.h" target="_blank">slicing block device</a> allows you to partition storage into smaller block devices that you can use independently, and the <a href="https://github.com/ARMmbed/mbed-os/blob/master/features/filesystem/bd/ChainingBlockDevice.h" target="_blank">chaining block device</a> allows you to chain multiple block devices together and extend the usable amount of storage.
+Additionally, two utility block devices give you better control over how storage is allocated. The [slicing block device](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/class_slicing_block_device.html) allows you to partition storage into smaller block devices that you can use independently, and the [chaining block device](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/class_chaining_block_device.html) allows you to chain multiple block devices together and extend the usable amount of storage.
 
 <span class="notes">**Note:** Some file systems may provide a format function for cleanly initializing a file system on an underlying block device or require external tools to set up the file system before the first use.</span>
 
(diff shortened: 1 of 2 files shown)


docs/reference/contributing/storage/storage.md
```diff
@@ -1,3 +1,27 @@
 <h2 id="contributing-storage">Storage</h2>
 
-The two storage APIs are BlockDevice and FileSystem. These APIs provide filesystem and file operations on a block device, as well as an interface for access to block-based storage.
+Storage support is split between file systems and their underlying block device support. The [storage API page](/docs/v5.6/reference/storage-overview.html) has more information on existing APIs in Mbed OS for both interfaces.
+
+#### Block Device
+
+Backing file systems on new hardware requires adding a block device implementation. You can extend the [BlockDevice](https://github.com/ARMmbed/mbed-os/blob/master/features/filesystem/bd/HeapBlockDevice.h) class to provide support for unsupported storage.
+
+If you want to port a new file system to Mbed OS on existing storage options, you can skip to the following section.
```
AnotherButler: Query: What do you mean by the following section? File systems two lines below? Or the next page? What should the reader do if the reader doesn't want to port a new file system to Mbed OS?

docs/reference/contributing/storage/storage.md
```diff
@@ -1,3 +1,27 @@
...
+- `file_close`.
+- `file_read`.
+- `file_write`.
+- `file_seek`.
+
+Here is the full API that a file system may implement:
+
+[![View code](https://www.mbed.com/embed/?type=library)](https://github.com/ARMmbed/mbed-os/blob/master/features/filesystem/FileSystem.h#L205)
+
+A block device must back file systems in Mbed OS. If you are using supported hardware, then you can continue. Otherwise, view the block device porting section above.
```
AnotherButler: Query: What does continue mean here? Read the next page or continue what the reader's doing? Please clarify.

docs/reference/contributing/storage/storage.md
kegilbert: Correct, the following section is referring to the file system section. 

A reader going through this section will want one of the following 3 situations:

1. Add a new physical or virtual storage device, requiring the base abilities of reading, writing, and erasing it.
2. Add a new filesystem implementation on an existing storage solution (Say on a SD card which Mbed OS supports). They do not need to port any  block device code themselves (what the last line in the Block Device section is referring to).
3. Combine the above two, a reader may want to implement a new filesystem solution on a block device that is not supported in Mbed OS.

docs/reference/contributing/storage/storage.md
kegilbert: This was worded a bit poorly, continue as in continue with your filesystem porting. Otherwise you'll need a block device port to back the filesystem and should implement that first.

kegilbert:
@sg- I can update all of the header file links. Question though, do we have a way to link to cpp files in particular? The link to https://github.com/ARMmbed/mbedos/blob/master/features/filesystem/fat/FATFileSystem.cpp was meant to show implementation details for porting as an example. 

@AnotherButler Is there a way to link to header file locations in Github from the class reference page? If a reader is being told to extend a class or look at an implementation it would be nice to know where to find said class in the source without too much manual work on their part.

sg-:
> was meant to show implementation details for porting as an example.

In that case, just mentioning a class / implementation name google will find should suffice. I dont want to have links suggest where code lives in the case of linking a repo.

kegilbert:
Fair enough.

kegilbert:
@AnotherButler @sg- Replaced the github links, note that the filesystem links will be broken until the doxy files are regenerated to suck in the filesystem changes from https://github.com/ARMmbed/mbed-os/pull/5395 . Updated the PR description to note this.

docs/reference/api/storage/storage.md
```diff
@@ -7,15 +7,15 @@ The storage APIs present in Arm Mbed OS are:
 
 #### Declaring a file system
 
-The <a href="https://github.com/ARMmbed/mbed-os/blob/master/features/filesystem/FileSystem.h" target="_blank">FileSystem</a> class provides the core API for file system operations. You must provide a block device to back the file system. When you declare a file system with a name, you can open files on the file system through standard POSIX functions (see <a href="http://pubs.opengroup.org/onlinepubs/009695399/functions/open.html" target="_blank">open</a> or <a href="http://pubs.opengroup.org/onlinepubs/9699919799/functions/fopen.html" target="_blank">fopen</a>).
+The [FileSystem](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/class_file_system.html) class provides the core API for file system operations. You must provide a block device to back the file system. When you declare a file system with a name, you can open files on the file system through standard POSIX functions (see <a href="http://pubs.opengroup.org/onlinepubs/009695399/functions/open.html" target="_blank">open</a> or <a href="http://pubs.opengroup.org/onlinepubs/9699919799/functions/fopen.html" target="_blank">fopen</a>).
```
sg-: we should not be linking to open, fopen, rather have these documented ourselves.

kegilbert: @sg- Replaced external links with short blurbs on open and fopen.

kegilbert:
@sg- Do the new changes look good to you? Did you want anything else changed?

sg-:
Can you confirm how compliant each standard library is with POSIX? If not lets reduce the use of that term. Looking better.

kegilbert:
@sg- IAR and ARM's standard libraries I don't believe are completely compliant with POSIX. I spoke with Chris a bit on the history of these docs, it's a bit murky. The reference to POSIX here was mainly pointing out the subset of POSIX functions Mbed OS has C++ wrappers for. 

If it's less ambiguous I can remove the mention of POSIX from the docs in this page.

kegilbert:
@sg- Were you thinking scratch the POSIX references from this page or leave as is?

sg-:
scratch them and just document the function name and what it does

kegilbert:
@sg- @AnotherButler Removed POSIX references.

kegilbert:
@sg- Yes it does conflict with the LittleFS PR. We merged in some of the changes with https://github.com/geky/mbed_Handbook/commit/0c4f2e612e659ded1f89005879f0e94737aab0d5.

We get the UML diagrams from the doxygen in the [class reference](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/classmbed_1_1_file_system.html), did you want those embedded in a different way/place?

sg-:
Do those render in the docs site? I didnt think so but could be wrong. They do miss the point of usability when compared to what we drew showing an example of classes that exist today not to be confused with the requirement to always keep upto date.

AnotherButler:
We don't transclude those images on the docs site. That's a feature we have to ask for, so I don't know how well they render.

kegilbert:
They're pretty hard to get to honestly. Had to do the following:

- Go to the File System API page from: https://os.mbed.com/docs/v5.6/reference/storage.html
image
- Select the FileSystem class reference (which seems unnecessary to most people as the doxygen API reference is already pulled into the page).
image

sg-:
In that case a ppt version or artistic rendering of the whiteboard ftw!

AnotherButler:
Could we screenshot the image from the Doxy?

sg-:
Its not good enough for that

kegilbert:
This isn't a diagram we intend to change frequently right? Seems pretty set. I'd be fine with a screenshot/artistic rendition.

AnotherButler:
Please see PR #358 for further changes on the contributing storage page. We've pulled the API storage page into the existing file.

PR #326, opened 2017-11-02
Remove CThunk page since this will be deprecated

c1728p9:
CThunk isn't part of mbed-os's public API and will be deprecated in the future so this patch removes it from the docs.


c1728p9:
CC @AnotherButler @sg- @0xc0170

PR #327, opened 2017-11-03
BLE: Add iBeacon reference.

pan-:
This page describes what the iBeacon technology is and what the iBeacon class do.


pan-:
@sg- @AnotherButler Could you review ?

PR #330, opened 2017-11-03
Change storage config snippet to JSON

kegilbert:
Was this more in line with what you wanted @AnotherButler, @sg-?

I was going to update the rest of the configuration docs under the https://os.mbed.com/docs/v5.6/reference/configuration-overview.html chapter if so.


kegilbert:
Gotchya, I'll revert it. What else does IOTMORF-1273 need? More general descriptive text? @sg-

kegilbert:
Clarified offline what needed to be done, closing.

PR #334, opened 2017-11-10
BLE: Add reference of Heart rate service.

pan-:
This page describes what the Bluetooth Heart Rate Service is and what the HeartService class do.


pan-:
@sg- @AnotherButler Could you review ?

PR #336, opened 2017-11-13
Add UARTSerial API

adbridge:
This adds the API doc for UARTSerial. 

To  Do

1. The actual examples need to be published on mbed_official.
2. Examples 1 and 3 tested on K64F successfully. Example 2 needs testing - requires specialized hardware such as ESP8266 and Arduino header. 
3. Links to the examples in the API document then need to be updated.

Example contents will be pasted here for review purposes.

### DIFF
--- a/docs/reference/api/drivers/UARTSerial.md
+++ b/docs/reference/api/drivers/UARTSerial.md
@@ -1,0 +1,54 @@
+## UARTSerial
+
+The UARTSerial class provides an alternative and enhanced version of the [Serial](https://os.mbed.com/docs/v5.6/reference/serial.html) interface. 
+
+
+Features:
+
+* It uses the latest FileHandle API that permits non-blocking and device operations. It is now possible to use a serial device with the C library file handling functions.
+E.g
+
+        UARTSerial serial(RX, TX);
+
+        FILE *in = fdopen(&serial, “r”);
+
+        FILE *out = fdopen(&serial, “w”);
+
+        while (1) {
+
+            fscanf(in, “%s”, buffer);
+            fprintf(out, “Hello %s\r\n”, buffer);
+        }
+  
+
+* It has built-in buffering, making input and output more efficient.
+
+* It also implements a poll() function which is useful for juggling multiple streams.
+
+
+### UARTSerial class reference
+
+[![View code](https://www.mbed.com/embed/?type=library)](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/classmbed_1_1_uartserial.html)
+
+<span class="notes">**Note**: On a Windows machine, you will need to install a USB serial driver. See [Windows serial configuration](/docs/v5.6/tutorials/serial-communication.html#windows-serial-driver).</span>
+
+
+### UARTSerial examples
+
+#### Example one
+
+Write a "Hello world" message to a device at a baud rate of 19200.
+
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/UARTSerial_HelloWorld/)](https://os.mbed.com/teams/mbed_example/code/UARTSerial_HelloWorld/file/7376f17bb36e/main.cpp)  **LINK TBC**
+
+#### Example two
+
+Provide a serial pass-through between the PC and an external UART.
+
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/UARTSerial_Example2/)](https://os.mbed.com/teams/mbed_example/code/UARTSerial_Example2/file/8d318218bac1/main.cpp) **LINK TBC**
+
+#### Example three
+
+Schedule a callback on the global event queue whenever there is a UART interrupt. The callback checks if a character has been typed and if so reads it and echoes it back to the terminal. It also toggles the LED state on each key press.
+
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/UARTSerial_Example3)](https://os.mbed.com/teams/mbed_example/code/UARTSerial_Example3/file/3b040f367dd8/main.cpp) **LINK TBC**


adbridge:
UARTSerial_HelloWorld example:

```
#include "mbed.h"
#include <stdio.h>

UARTSerial device(USBTX, USBRX);

int main() {
    device.set_baud(19200);
    FILE *output = fdopen(&device, "w");

    fprintf(output, "Hello World\r\n");
}
```

adbridge:
UARTSerial_Example2

```
#include "mbed.h"

UARTSerial pc(USBTX, USBRX);
UARTSerial device(MBED_CONF_APP_UART1_TX, MBED_CONF_APP_UART1_RX);

static void copy_some(FileHandle *out, FileHandle *in) {
    // To ensure performance, allow to read multiple bytes at once, although
    // we don't expect to read many in practice.

    // read() will return immediately, as we've already
    // checked that `in` is ready with poll()
    char buffer[32];
    ssize_t read = in->read(buffer, sizeof buffer);
    if (read <= 0) {
        error("Input error");
    }

    // Then write them all out. Assuming output port is similar speed to input,
    // this may block briefly, but not significantly.
    ssize_t written = 0;
    while (written < read) {
        ssize_t w = out->write(buffer + written, read - written);
        if (w <= 0) {
            error("Output error");
        }
        written += w;
    }
}

int main() {
    char buffer[32];
    pollfh fds[2];

    fds[0].fh = &pc;
    fds[0].events = POLLIN;
    fds[1].fh = &device;
    fds[1].events = POLLIN;

    while (1) {
        // Block until either of the 2 ports is readable (or has an error)
        poll(fds, 2, -1);

        if (fds[0].revents) {
            copy_some(fds[1].fh, fds[0].fh);
        }
        if (fds[1].revents) {
            copy_some(fds[0].fh, fds[1].fh);
        }
    }
}
```

adbridge:
UARTSerial_Example3

#include "mbed.h"

DigitalOut led1(LED1);
DigitalOut led2(LED2);

UARTSerial pc(USBTX, USBRX);

void callback_ex() {
    while (1) {
        char c;
        if (pc.read(&c, 1) != 1) {
            break;
        }
        printf("%c\n", c);
        led2 = !led2;
    }
}

int main() {
    pc.set_blocking(false);
    pc.sigio(mbed::mbed_event_queue()->event(callback_ex));

    while (1) {
        led1 = !led1;
        wait(0.5);
    }
}

adbridge:
@AnotherButler @sg- Please review

docs/reference/api/drivers/UARTSerial.md
```diff
@@ -0,0 +1,54 @@
+## UARTSerial
+
+The UARTSerial class provides an alternative and enhanced version of the [Serial](https://os.mbed.com/docs/v5.6/reference/serial.html) interface. 
+
+
+Features:
```
sg-: Dont care much for this section, dry and factual, we dont need to sell it :)

docs/reference/api/drivers/UARTSerial.md
```diff
@@ -0,0 +1,54 @@
...
+
+* It uses the latest FileHandle API that permits non-blocking and device operations. It is now possible to use a serial device with the C library file handling functions.
```
sg-: latest wont be true forever so can drop that, in fact, this was always possible with Stream so not sure why calling it out here.

docs/reference/api/drivers/UARTSerial.md
```diff
@@ -0,0 +1,54 @@
...
+E.g
+
+        UARTSerial serial(RX, TX);
```
sg-: Lets keep this as a serial example, not mixing the filehandle in

docs/reference/api/drivers/UARTSerial.md
adbridge: From what Kevin was saying, it was designed to replace the need for stream in this context....

docs/reference/api/drivers/UARTSerial.md
adbridge: AFAIU from Kevin, being able to use filehandles and thus directly call things like fprintf() to write out to the serial port was one of the main advantages of UARTSerial over Serial. If we just make this page look like Serial, why would anyone use it ?

SeppoTakalo:
Why this has been parked?

Those example are clear and simple. They demonstrate why `UARTSerial` should be used instead of old Serial.

Also, I don't thin such a simple API need examples pushed in mbed.org. Just have them inline in the documentation.

sg-:
> Why this has been parked?

It just needs to be updated. The use of file handles while important is not the whole API. The examples give a very angled use pattern which does not follow other Mbed OS API usage patterns. I agree this is a strong feature of UARTSerial but there are still ongoing work around updating Stream to manage file handles within an object rather than dropping this all on the user. If the API as encapsulated was presented and described I wouldn't have as much objection. But the state of the contribution doesn't wholly reflect the usage model of the API.

@SenRamakri

SeppoTakalo:
To my understanding, use of file handles is the most important feature of this API. That with buffering.
So It is not angled use pattern, it is THE use pattern we want it to be used.

In fact, in current Mbed OS if you turn on
```
"platform.stdio-buffered-serial": true,
```

You get `stdin` and `stdout` that are UARTSerial. And those are filehandles.

I fail to see what the use case of the Stream class is after that. The whole Stream class is bad design and undocumented. It tries to be both filehandle and still give out `printf()` and `scanf()` functions and has a name.

sg-:
> I fail to see what the use case of the Stream class is after that. The whole Stream class is bad design and undocumented. It tries to be both filehandle and still give out printf() and scanf() functions and has a name.

The posix API compatibility is a detail of the update, not the centerpiece, never was. Sorry if that has not been clear. It deserves a tutorial page to itself, no question about that. The existing design pattern encapsulates print/scan functionality in Steam. Sure we're identified problems with the implementation but we have not yet worked out how to migrate all libraries and users of the uart.printf design pattern. This is what I was hoping would be achieved by updates to Stream such that it would be compatible with refactored file handles.

SenRamakri:
@sg- @SeppoTakalo @adbridge @AnotherButler - This is a great start for UARTSerial, the doc does exhibits following use cases/APIs.

FileHandle Interface
Posix style APIs
set_baud
read/write APIs
poll
set_blocking
sigio/Event handling/Callback

I think it would be better to demonstrate the below APIs as well which I think are supported by UARTSerial and is very relevant to serial communication.

set_format
set_flow_control
writable - This is sort of encapsulated by poll but still we provide and implement this.
readable - This is sort of encapsulated by poll but still we provide and implement this.
sync 
isatty

@AnotherButler - Can you please review the UARTSerial description here? Is there any opportunity to expand on the description. I think its more on the terser side.

PR #341, opened 2017-11-20
Add prohibited word check Travis job

kegilbert:
Covered this in the earlier email mostly, but recapping here for posterity. Spoke with Senthil and decided to put up the PR now, we can discuss it more when you're back in the office.

The job works by comparing the PR revision against the current head (set to new_engine) and listing any new prohibited words found. In the case of new, renamed, or moved files all prohibited words are listed from the file in question if present. The thought being a) It’s a bit difficult to differentiate between a new file and a renamed file from the context of the script (I could try to add in some git checks but it’ll be a bit uglier) and b) If you’re touching a file, might not be a bad thing to be aware of any black listed words present so we can slowly clean up the existing docs. Line and whitespace changes will not trigger errors, the word itself is mapped to its file and compared against the two revisions. Likewise removing existing prohibited words will not trigger errors.

Sample runs:

- Expected failures: https://github.com/kegilbert/Handbook/pull/2
- Expected pass:     https://github.com/kegilbert/Handbook/pull/3

Clicking on the failed Travis job in the PR will show you the build, scroll to the bottom to get a list of errors.

The concern was with fuzzy situations where a prohibited word pops up in a phrase where it’s deemed ok to have, or if you move a file and don’t want to fix all of the errors you’ll have a failing CI when merging. I think it'd be better to at least have the list of failed words on each PR and fail the CI on any errors and submit based on manual inspection.

Thoughts?

@AnotherButler 
@iriark01

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,9 +71,9 @@ jobs:
       script:
         # Run scripts to search for added prohibited words
         - python check_tools/validate.py -c -i check_tools/prohibited.txt -f . &> check_tools/pr_black_list.log
-        - git remote add upstream https://github.com/armmbed/Handbook.git
-        - git fetch --depth=1 upstream new_engine
-        - git checkout new_engine
+        - git remote add upstream https://github.com/armmbed/mbed-os-5-docs.git
+        - git fetch --depth=1 upstream development
+        - git checkout development
         - python check_tools/validate.py -c -i check_tools/prohibited.txt -f . &> check_tools/head_black_list.log
         - python check_tools/smartdiff.py
 # Manage statuses
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,16 @@ jobs:
                 fi
             fi
 
+      env:
+        - NAME=prohibited-word-check
+      script:
+        # Run scripts to search for added prohibited words
+        - python check_tools/validate.py -c -i check_tools/prohibited.txt -f . &> check_tools/pr_black_list.log
+        - git remote add upstream https://github.com/armmbed/Handbook.git
+        - git fetch --depth=1 upstream new_engine
+        - git checkout new_engine
+        - python check_tools/validate.py -c -i check_tools/prohibited.txt -f . &> check_tools/head_black_list.log
+        - python check_tools/smartdiff.py
 # Manage statuses
 before_install:
   - |
@@ -103,4 +113,3 @@ after_success:
 # Job control
 stages:
     - name: test
-
--- a/check_tools/smartdiff.py
+++ b/check_tools/smartdiff.py
@@ -23,25 +23,33 @@ def main():
 
     for doc,words in pr_dict.iteritems():
         # Split the line number and trailing newlines from the list of found prohibited words
-        # Check for move in file location as the path is included in the key. Catch on try failure
         try:
-            stripped_pr_words = [word.split(',')[1].replace("\n","") for word in words]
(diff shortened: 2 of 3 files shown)


kegilbert:
@AnotherButler requested a manual override, looking into a way to add that.

kegilbert:
@AnotherButler This PR's gotten a bit old. Is  there still interest? Looks like the lack of a way to manually override false negatives is holding it back. That feature is likely possible but will take a good amount of additional effort that I won't be able to put into this for a good period of time. 

If manually overriding results in necessary I'll leave the PR as is and get back to you when I can poke around a bit more with adding that as a feature.

cmonr:
@kegilbert You need a rebase.

.travis.yml
```diff
@@ -66,6 +66,16 @@ jobs:
...
             fi
 
+      env:
+        - NAME=prohibited-word-check
+      script:
+        # Run scripts to search for added prohibited words
+        - python check_tools/validate.py -c -i check_tools/prohibited.txt -f . &> check_tools/pr_black_list.log
+        - git remote add upstream https://github.com/armmbed/mbed-os-5-docs.git
+        - git fetch --depth=1 upstream development
+        - git checkout development
```
cmonr: Why is this needed? Shouldn't Travis CI already have this checked out?

check_tools/validate.py
```diff
@@ -1,50 +1,63 @@
...
-    
+
     if len(options.inputfile) <= 0 or len(options.wordsfile) == None:
         parser.error("Invalid arguments")
         
     words_file = open(options.wordsfile)
     alllines = words_file.read()
-    words = alllines[:-1].replace("\n","|")
-    
+    words = alllines[:-1].replace("\n","|")
```
cmonr, line 47: Trailing whitespace

.travis.yml
```diff
@@ -66,6 +66,16 @@ jobs:
...
+        - python check_tools/validate.py -c -i check_tools/prohibited.txt -f . &> check_tools/head_black_list.log
```
cmonr, line 77: What's the reason that this will not always be empty?

check_tools/smartdiff.py
```diff
@@ -0,0 +1,52 @@
+import sys
+
+def format_map(black_list):
```
cmonr, line 3: An example of what kind of string format this is expecting would go a long way for readability and understanding what exactly this is parsing.

check_tools/smartdiff.py
```diff
@@ -0,0 +1,52 @@
...
+            # words. If a key error is found here, either the head revision contained no prohibited words, a
+            # new file was added, or an existing file was moved or renamed. Fill with empty list to
+            # compare the PR revision against
+            stripped_head_words = [word.split(',')[1].replace("\n","") for word in head_dict[doc]]
+        except KeyError:
+            stripped_head_words = []
+
+        stripped_pr_words = [word.split(',')[1].replace("\n","") for word in words]
+
+        # Do not want to trigger failure on removed prohibited words from existing docs.
```
cmonr, line 37: Can't help but think a word for this could be "whitelist", but unsure where that would be used within this PR/these scripts.

.travis.yml
cmonr: If a `check_tools` directory already exists, why not commit this `head_black_list.log` (whitelist?) with the developement branch as well?

cmonr (commented):
Questions

PR #342, opened 2017-11-21
Add UDP and TCP ping-pong examples.

janjongboom:
@AnotherButler Two new examples, as discussed


AnotherButler:
👍 This looks great. Thanks for doing this. I changed the formatting for this section to match our new formatting, so all links open in a new tab for best linking practices.

PR #349, opened 2017-11-30
Update Event docs

deepikabhavnani:
1. Ensure the Event API actually refers to Event and not EventQueue
2. Create the EventQueue API reference by following the structure of the other API references.

@AnotherButler - Please review


docs/reference/api/rtos/EventQueue.md
```diff
@@ -1,11 +1,32 @@
...
+
+### EventQueue example: posting events to the queue
+
+The code below demonstrates queueing functions to be called after a delay and queueing functions to be called periodically.
+
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/events_ex_2/)](https://os.mbed.com/teams/mbed_example/code/events_ex_2/file/488fe91e2e80/main.cpp)
+
+### EventQueue example: chaining events from more than one queue
+
+Event queues easily align with module boundaries, where internal state can be implicitly synchronized through event dispatch. Multiple modules can use independent event queues but still be composed through the `EventQueue::chain` function.
```
AnotherButler: Query: Would it be correct to say "... where event dispatch can implicitly synchronize internal state."?

docs/reference/api/rtos/EventQueue.md
```diff
@@ -1,11 +1,32 @@
...
 
 [![View code](https://www.mbed.com/embed/?type=library)](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/classevents_1_1_event_queue.html)
 
-### EventQueue example
+### EventQueue example: deferring from interrupt context
+
+The code executes two handler functions (`rise_handler` and `fall_handler`) in two different contexts:
+
+1. In interrupt context when a rising edge is detected on `SW2` (`rise_handler`).
+2. In the context of the event loop's thread function when a falling edge is detected on `SW2` (`fall_handler`). `queue.event()` is called with `fall_handler` as an argument to specify that `fall_handler` runs in user context instead of interrupt context.
```
AnotherButler: Query: Would it be correct to say "`fall_handler` calls `queue.event()` as an argument to specify..."?

deepikabhavnani: No, fall_handler function is an argument to queue.event().
We can rephrase it as, 
You can use 'fall_handler` function as argument to `queue.event()` if you want to run it in user context instead of interrupt context.

docs/reference/api/rtos/EventQueue.md
deepikabhavnani: Yes

docs/reference/api/rtos/EventQueue.md
```diff
@@ -1,11 +1,122 @@
...
+```
+
+### Shared event example: running the shared queue from main
+
+To further save RAM, if you have no other work to do in your main function after initialization, you can dispatch the global event queue from there, avoiding the need to create a separate dispatch thread.
+
+To do this, set the `mbed_app.json` configuration option `events.shared-dispatch-from-application` to true, and add a dispatch call to main, as in this example. (The prints now show the same context for startup and `fall_handler`).
+
+```
+#include "mbed.h"
```
AnotherButler: @deepikabhavnani Could you please help me add this example to the teams repo, so it's not hard-coded into the docs?

docs/reference/api/rtos/EventQueue.md
```diff
@@ -1,11 +1,122 @@
...
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/events_ex_3/)](https://os.mbed.com/teams/mbed_example/code/events_ex_3/file/fca134a32b61/main.cpp)
+
+### Shared event example: deferring from interrupt context
+
+Like the previous example, this defers from interrupt to an event queue thread. However, rather than creating its own thread, it uses the shared event queue – potentially sharing it with other system components and saving RAM.
+
+As the event queue is shared, you should limit the execution time of your event functions to avoid delaying other users’ events excessively.
+
+```
+#include "mbed.h"
```
AnotherButler: @deepikabhavnani Could you please help me add this example to the teams repo, so it's not hard-coded into the docs?

docs/reference/api/rtos/Event.md
```diff
@@ -6,7 +6,7 @@ The Event class is thread safe. The `post` and `cancel` APIs are IRQ safe.
 
 ### Event class reference
 
-[![View code](https://www.mbed.com/embed/?type=library)](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/_event_8h_source.html)
+[![View code](https://www.mbed.com/embed/?type=library)](https://os-doc-builder.test.mbed.com/docs/v5.7/mbed-os-api-doxy/_event_8h_source.html)
```
AnotherButler: Query: Is there a reason we're pointing to .h file instead of to the Event class (https://os-doc-builder.test.mbed.com/docs/v5.7/mbed-os-api-doxy/classevents_1_1_event.html)?

AnotherButler: After our in-person conversation, I think you're correct, and we should point to this .h file instead of the class page. We can always update this later when the class page is more complete.

PR #350, opened 2017-12-01
Update common_issues.md

MarceloSalazar:
I've updated the doc with tweaking a few things and adding other items.

- Added troubleshooting as the title. This is not all about "debug", which means something very specific for embedded sw developers :)
- Many issues applies in particular to users who haven't used mbed before.
- Point at the latest Mbed CLI version, so we don't need to docs the version each time.
- Point at latest & supported toolchain versions
- Note for compiler licenses
- Point at additional documentation for proper debug (using 3rd party IDEs).

@AnotherButler please have a look and let me know what you think.

### DIFF
--- a/docs/tutorials/debug/common_issues.md
+++ b/docs/tutorials/debug/common_issues.md
@@ -6,8 +6,7 @@ When you start using Mbed OS, you may find small challenges in a few cases. This
 
 #### Mbed CLI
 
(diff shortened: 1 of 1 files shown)


MarceloSalazar:
Looks good, thanks a lot @AnotherButler! :)

AnotherButler:
Thank you. I appreciate the quick review.

PR #351, opened 2017-12-04
Update Ticker and SleepManager based on feedback

c1728p9:
Remove information about the 30-minute limitation from the Ticker docs, as the newer API is not limited by this. Add a note to about the higher power consumption of ticker. Add a note to SleepManager indicating that sleep() should not need to be called normally. Also explicitly reference classes which are safe to use in deep sleep mode.

### DIFF
--- a/docs/reference/api/drivers/Ticker.md
+++ b/docs/reference/api/drivers/Ticker.md
@@ -6,11 +6,11 @@ You can create any number of Ticker objects, allowing multiple outstanding inter
 
 ### Warnings and notes
 
-* Timers are based on 32-bit int microsecond counters, so they can only time up to a maximum of 2^31-1 microseconds (30 minutes). They are designed for times between microseconds and seconds. For longer times, you should consider the time() real time clock.
-
 * No blocking code in ISR: avoid any call to wait, infinite while loop or blocking calls in general.
 
 * No printf, malloc or new in ISR: avoid any call to bulky library functions. In particular, certain library functions (such as printf, malloc and new) are not re-entrant, and their behavior could be corrupted when called from an ISR.
+
+* While an event is attached to a Ticker deep sleep is blocked to maintain accurate timing. If you don't need microsecond precision consider using the LowPowerTicker class instead since this does not block deep sleep mode.
 
 ### Ticker class reference
 
--- a/docs/reference/api/drivers/Ticker.md
+++ b/docs/reference/api/drivers/Ticker.md
@@ -6,11 +6,11 @@ You can create any number of Ticker objects, allowing multiple outstanding inter
 
 ### Warnings and notes
 
-* No blocking code in ISR: avoid any call to wait, infinite while loop or blocking calls in general.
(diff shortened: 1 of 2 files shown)


c1728p9:
@0xc0170 the user facing API of the sleep manager is primarily the lock and unlock functions. Should the idle loop info be in this page? If not do you have any recommendations on where to move it?

0xc0170:
The only doc I am aware of is https://os.mbed.com/docs/latest/reference/wait.html that can be referenced? There's a good info there ` If there are no other threads in ready state, it can put the whole microcontroller to sleep, saving energy.` 

> And I do not understand the idle_loop, there is no example for it. How do I break away from it? Why would I override it?

@c1728p9 Should wait doc add small info for how it is done - that there exists idle loop that by default sleeps, and can be overwritten by an application?

@lauri-piikivi has a point, I could not locate info about idle loop weak linkage and what is the default idle loop behavior. That would answer all his questions.

c1728p9:
@0xc0170 can this PR be merged as is?

Also, is overriding the idle loop something we want to support/encourage? It seems like this just makes bugs more likely and adds confusion to our docs. What do you think @0xc0170?
CC @deepikabhavnani

0xc0170:
> @0xc0170 can this PR be merged as is?

> You do not need to call `sleep()` directly unless you are overriding the default sleep handling of Mbed OS.

Only this is not clear to me. We have default idle hook that sleeps (tickless or not). How this would work if sleep is used also in the app? I would discourage to use sleep in the app, rather to use idle loop hook. If my understanding is not correct, then this is LGTM, should get in soon.

> Also, is overriding the idle loop something we want to support/encourage? It seems like this just makes bugs more likely and adds confusion to our docs. What do you think @0xc0170?

Good question (this relates to my above question),  we have sleep manager - an app can disable deep sleep, but not sleep or do app specific things before and after sleep (If we remove idle hook, we do not have any other hook for app to do something before sleep or after sleep. We can discuss this separately.

AnotherButler:
@lauri-piikivi Please review again at your convenience.

c1728p9:
>>You do not need to call sleep() directly unless you are overriding the default sleep handling of Mbed >OS.
>
>Only this is not clear to me. We have default idle hook that sleeps (tickless or not). How this would >work if sleep is used also in the app?

Calling sleep outside the background thread should be discouraged, as this can cause problems. Unlike `Thread::wait()` calling `sleep()` does not cause a task switch to the next available thread. Instead, it puts the core to sleep (letting no other threads run) until an interrupt wakes the processor from sleep. This isn't a problem on the background thread because it has the lowest priority and only gets to run when all the other threads are already suspended.

0xc0170:
Lets merge this, any outstanding issues can be tracked via issue or resolved via separate PR

AnotherButler:
@lauri-piikivi If you agree that we can merge this and track outstanding issues separately, could you please approve this, so we can merge it?

lauri-piikivi:
Approved -- it is going in a very good direction. Thanks to All!
(I do not seem to have any explicit approve button anymore, or I just cannot find it)

PR #352, opened 2017-12-04
Write docs page for LowPowerTicker

c1728p9:
Fill in LowPowerTicker.md with content and make a slight update to Ticker.


c1728p9:
Hi @0xc0170 I made that section for the LowPowerTicker since the regular Ticker had that warnings section as well. The doxygen  for ```attach`` does not have this information. If you want this additional information please create a separate PR.

PR #354, opened 2017-12-04
QuadSPI API documentation

SenRamakri:
This provides Quad SPI driver API documentation. Note that the QuadSPI implementation will be in a feature branch for 5.7 release.


AnotherButler:
Note to self: Merge after the updated Doxy kicks in.

0xc0170:
@AnotherButler Where do we define this is feature driver, it is on the branch. Means its under development (contributions welcome) ? I expect that as its feature branch, we receive feedback from partners, community that could potentially lead to changes.

I get an impression from reading this document that QSPI was released and it's ready for production

AnotherButler:
@0xc0170 This is a placeholder. Because we don't know when or how this will be delivered, we created it here for now. It will not be merged or added to the .json. Instead, it will be moved to the decided delivery method once we know what that is.

kegilbert:
Some questions primarily in regards to the linked example for now:

https://os.mbed.com/teams/mbed_example/code/mbed-os-example-qspi/file/cc0165946232/main.cpp/

Is this the final example we plan on using? I like the actual functionality of it rather than just creating a QSPI instance and doing nothing with it, but the example seems like a bit much at first blush compared to our other examples. While we need more examples like this, should we have a smaller stepping stone example not tied to specific hardware in addition to this? I'm not too familiar with the QSPI feature so may not really be reasonable just throwing it out there.

In case this is the final example proposal, a few minor nits:

1. Minor: Line 40: Why is the QSPI object being allocated here with new out of curiosity? Our other examples don't typically do this. Is it primarily to be able to check for memory allocation or QSPI ctor failures?

```c++
 myQspi = new QSPI((PinName)QSPI_PIN_IO0, (PinName)QSPI_PIN_IO1, (PinName)QSPI_PIN_IO2, (PinName)QSPI_PIN_IO3, (PinName)QSPI_PIN_SCK, (PinName)QSPI_PIN_CSN);
```

2. Minor: Line 72: Why is the tx_buf initiated manually character by character rather than with
```c++
tx_buf[] = "HELLO QUAD-SPI!";
```

3. Minor: Line 227: Argument comments for the final 
```myQspi->command_transfer```
aren't lined up like the other calls are.

AnotherButler:
Who can fix the example?

kegilbert:
@maciejbocianski is listed as the developer in the teams example, would you be able to clean up the example? Senthil will be out for another ~1.5 weeks.

maciejbocianski:
@kegilbert @AnotherButler sure thing, I will do it tomorrow

kegilbert:
Thanks!

maciejbocianski:
@AnotherButler @kegilbert @0xc0170 @SenRamakri 
QSPI example and README has been updated (please https://os.mbed.com/teams/mbed_example/code/mbed-os-example-qspi/)

Please review

PR #366, opened 2017-12-18
Add mbed-mesh-api content to Connectivity.md

AnotherButler:
Add Mesh configuration content to connectivity configuration section.


SeppoTakalo:
Hmm.. Do we now lose all the descriptions from these fields? Does it make sense?

AnotherButler:
This follows the configuration structure of the other APIs. Does it make sense to have only mesh following a different structure?

SeppoTakalo:
Mesh networks are basically always a bit different.
On a typical Ethernet connection, no configuration is needed at all. On a Wifi, the SSID and password are enough, and those are already part of the API.
On a mesh, we typically require channel configurations, security keys, etc. Lots of stuff that is provided by the configuration system but it's not part of the API.

I therefore strongly believe that these descriptions should be retained within the mesh sections of the document. I'm not against of having them shown in the page where all configuration field are displayed but I we should not lose the description of the fields. Name is just not enough.

AnotherButler:
It seems strange to me that we include configuration content in the networksocket API that we do not include in any of the other APIs. It goes against the templated structure we're trying to create. Is there a way to include this information in the configuration section but keep it streamlined with the other content in that section? Is there a script we can run that includes all of that information for all of the configuration, not just networksocket?

AnotherButler:
@SeppoTakalo Any further thoughts?

SeppoTakalo:
I still don't like this proposal.

If we wan't to move the mesh configuration to under configuration section of handbook why don't we move the entire page, without any modification, as a another section. Just after "Connectivity" page.

Some of the pages already have explaining texts associated with configuration variables: https://os.mbed.com/docs/v5.7/reference/configuration.html

So if we just move the configuration variable names to this common page, we lose their definition and documentation which I'm against of. We should not remove the documentation.

sg-:
The biggest problem to me seems to be that the configuration doesnt have a description alongs side the option. If the information from these tables should really be in the config json. https://github.com/ARMmbed/Handbook/pull/366/files#diff-f1a357b922459ef5dd806ab5dd0d970cL189

kjbracey-arm:
I've added a PR to mbed-os to add the help to the JSON, which would allow the auto-generated Connectivity.md to come out better.

AnotherButler:
@SeppoTakalo Now that the code has merged, could you please update this PR, so we can merge it? Thanks so much.

AnotherButler:
@SeppoTakalo Does your connectivity overhaul override this PR?

SeppoTakalo:
Yes, this is obsoleted by https://github.com/ARMmbed/Handbook/pull/502 and earlier network related PRs and can be closed.

PR #381, opened 2018-01-24
Update connectivity config options

kegilbert:
Generated via:
`mbed compile --config -m K64F -t GCC_ARM -v`

@AnotherButler


AnotherButler:
@geky Is this the right script? [It looks right to me, but I'd like an engineer to confirm.]

PR #382, opened 2018-01-25
Add pointer and references examples

### LINKED ISSUE #292: Coding guideline: Clarification on pointer position in declarations
In the coding guideline, under the `rules` section the following rule is enunciated: 

> For pointers, `*` is adjacent to a name (`analogin_t *obj`).

It raise several questions: 

* Should this rule be applied to references ? 
* What if the name is omitted ? This can be the case in function declaration. Shall it be `void foo(int*)` or `void foo(int *)`
* Is that rule valid for function return type ? Shall it be `int *foo(void)` or `int* foo(void)` ? 
* What shall be used to declare immutable pointers ? The rule cannot be applied because the `const` modifier have to go to the right of the pointer according to the C and C++ standards. 
* How pointer to pointer and reference to pointer shall be written `int* &foo` or `int *&foo` ? 

I also want to point out that C and C++ standard takes different stances on the subject (even if both constructs are valid). In the C standard there is always a space between the pointee type and the pointer declarator followed by another space if a qualifiers follow. In C++ standard pointer and references declarator are always weld to the type they point to.

It may be interesting to had a bit of background behind the choice. 

Another note, when doxygen generate a function documentation, the rule is broken and the pointer goes next to the type it points to ([example](https://os.mbed.com/docs/v5.6/mbed-os-api-doxy/classmbed_1_1_serial_base.html#abee1597dcb5e2dfcdfb42ed9fe871e37))

0xc0170:
@pan- This should fix #292 (we might want to add more, let me know). Your issue had it nicely described ,thanks for all that info there.

Reason: the current codebase (most of it) follows what is described here

### DIFF
--- a/docs/reference/contributing/guidelines/style.md
+++ b/docs/reference/contributing/guidelines/style.md
@@ -59,55 +59,57 @@ uint32_t adc_function(analogin_t *obj, uint32_t options)
 ```
 ##### Rules
 
-* Indentation - four spaces. Please do not use tabs.
+- Indentation - four spaces. Please do not use tabs.
 
-* Braces - K&R style.
+- Braces - K&R style.
 
-* One true brace style (1TBS) - use braces for statements of type `if`, `else`, `while` and `for` (exception [from K&R](http://en.wikipedia.org/wiki/Indent_style#Variant:_1TBS)).
+- One true brace style (1TBS) - use braces for statements of type `if`, `else`, `while` and `for` (exception [from K&R](http://en.wikipedia.org/wiki/Indent_style#Variant:_1TBS)).
 
-* One line per statement.
+- One line per statement.
 
-* Preprocessor macro starts at the beginning of a new line; the code inside is indented according to the code above it.
+- Preprocessor macro starts at the beginning of a new line; the code inside is indented according to the code above it.
 
-* Cases within `switch` are indented (exception from K&R).
+- Cases within `switch` are indented (exception from K&R).
 
-* Space after statements of type `if`, `while`, `for`, `switch`. The same applies to binary operators (like, `+` and `*`) and the ternary operator (`?` and `:`).
+- Space after statements of type `if`, `while`, `for`, `switch`. The same applies to binary operators (like, `+` and `*`) and the ternary operator (`?` and `:`).
 
-* Each line preferably has at most 120 characters.
+- Each line preferably has at most 120 characters.
 
-* Comments should use proper spelling and grammar.
+- Comments should use proper spelling and grammar.
 
-* For pointers, `*` is adjacent to a name (`analogin_t *obj`).
+- For pointers or references, the symbols `*` or `&` are adjacent to a name (`analogin_t *obj`. `analogin_t &obj`). If you omit the name, place the space between the type and `*` (such as `int *` or `int &`).
 
-* Don't leave trailing spaces at the end of lines.
+- For function return pointers or references, the symbols `*` or `&` are adjacent to a function name (`int *func()` or `int &func()`).
 
-* Empty lines should have no trailing spaces.
+- Don't leave trailing spaces at the end of lines.
 
-* Unix line endings are default option for files.
+- Empty lines should have no trailing spaces.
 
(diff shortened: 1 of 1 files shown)


docs/reference/contributing/guidelines/style.md
```diff
@@ -59,55 +59,57 @@ uint32_t adc_function(analogin_t *obj, uint32_t options)
...
+- Each line preferably has at most 120 characters.
 
-* Comments should use proper spelling and grammar.
+- Comments should use proper spelling and grammar.
 
-* For pointers, `*` is adjacent to a name (`analogin_t *obj`).
+- For pointers or references, the symbols `*` or `&` are adjacent to a name (`analogin_t *obj`. `analogin_t &obj`). If you omit the name, place the space between the type and `*` (such as `int *` or `int &`).
 
-* Don't leave trailing spaces at the end of lines.
+- For return pointers or references, the symbols `*` or `&` are adjacent to a return type (`int* func()` or `int& func()`).
```
0xc0170: `function return` or similar ? @AnotherButler please review

0xc0170:
Rebased ,squashed two commits into one (edit is included in the commit). Please review

geky:
Sorry, I only just saw this pr

docs/reference/contributing/guidelines/style.md
geky: This seems inconsistent with both variable declarations and type expressions.

``` cpp
int* doit(int *);
              ^--- right side?
   ^-------------- left side?
```

``` cpp
(struct { int *a; })* doit(int *);
                               ^--- right side?
                    ^-------------- left side?
              ^------------------- right side?                  
```

``` cpp
int (* doit(int *))(int *);
     ^------------------- what do we do here?
```

geky: I guess my qualms are more about how declarations are inconsistent with this rule.

In C, functions declarations are the same as other declarations, except the name has an argument list appended.

It is strange that they would have different style rules:
``` cpp
int hi;   // variable declaration
int hi(); // function declaration

int *hi;   // variable declaration
int* hi(); // function declaration
```

0xc0170: @geky  We had this discussion for critical section a while ago (2 or 3 years back). I only captured what is in the codebase, as Kevin highlighted the older code has it this way, the newer code has it differently.

Based on the feedback and closer look at the codebase (mainly in the file handles and Buss clases there are pointers and references, its very few of the older code base and not even consistent there), let's make it `right`. I'll rebase this PR

0xc0170:
Updated as I commented earlier, please review

docs/reference/contributing/guidelines/style.md
```diff
@@ -59,55 +59,57 @@ uint32_t adc_function(analogin_t *obj, uint32_t options)
...
-* Unix line endings are default option for files.
+- Empty lines should have no trailing spaces.
 
-* Use capital letters for macros.
+- Unix line endings are default option for files.
 
-* A file should have an empty line at the end.
+- Use capital letters for macros.
+
+- A file should have an empty line at the end.
```
kjbracey-arm: Hmm, do you mean "empty line at the end"? I would interpret that statement as meaning

    last line with content\n
    \n

Do you just mean "new line at the end"?

0xc0170: This was just a formatting fix done by @AnotherButler . I haven't touched.

Yes, new empty line at the end of the file. Should we clarify this one or?

kjbracey-arm: Save it for a later tweak maybe. Nothing to do with this PR.

I'm assuming you mean you want this

    last line with content\n
    |    <- cursor position

as opposed to this:

    last line with content, unterminated|   <- cursor position

and you don't really want:

    last line with content\n
    \n     <- empty line
    |      <- cursor position

0xc0170:
@AnotherButler I can't set reviewers . I would like to run this with @sg- , will discuss this update today, so should be ready within a day.

PR #384, opened 2018-01-25
events: latest event example fixes IRQ printf issue

### LINKED ISSUE #335: events_tutorial.md does printf() in ISR, this does not work
In https://github.com/ARMmbed/Handbook/blob/5.6/docs/tutorials/using_apis/events_tutorial.md we do `printf()` calls in ISRs to show how the code is ran. This is not supported. It used to work in GCC but not with latest mbed OS (throws 'not supported in interrupt context' or something similar on the serial) and never worked in ARMCC.

I'm not sure who wrote this article, but it needs to be rewritten.

@AnotherButler 

Note that I wrote https://developer.mbed.org/blog/entry/Simplify-your-code-with-mbed-events/ some time ago, maybe it fits this section too.

0xc0170:
Fixes #335


janjongboom:
Check.

0xc0170:
Rebased

PR #388, opened 2018-01-29
Update MemoryPool documentation to capture the block initialization s…

SenRamakri:
Update MemoryPool docs to clarify that the objects placed on Memory pools should be initialized manually by the user. Memory pool allocates raw memory blocks from the pool and initializes them to 0s.


AnotherButler:
@SenRamakri I've completed a copy edit on this. Please review, and make sure I didn't accidentally change the meaning of anything. Also, who should I tag as the engineering reviewer on this?

PR #390, opened 2018-01-30
Use template for sleep PG

bulislaw:
Update the Sleep PG to match the HAL template.

@AnotherButler we still need to correct the links to point out to doxy for feature branches when available.

### DIFF
--- a/docs/reference/contributing/target/sleep.md
+++ b/docs/reference/contributing/target/sleep.md
@@ -36,7 +36,7 @@ We are working on the new HAL Sleep API, which will replace the current version 
 
 To make sure your platform is ready for the upcoming changes, you need to implement the future API and submit it in a separate pull request against `feature-hal-spec-sleep` branch. You can find the API and specification for the new Sleep API in the following header file:
 
-[![View code](https://www.mbed.com/embed/?type=library)](https://github.com/ARMmbed/mbed-os/blob/feature-hal-spec-sleep/hal/sleep_api.h)
+[![View code](https://www.mbed.com/embed/?type=library)](https://os.mbed.com/docs/v5.7/feature-hal-spec-sleep-doxy/group__hal__sleep.html)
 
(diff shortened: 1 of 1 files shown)


docs/reference/contributing/target/sleep.md
```diff
@@ -1,46 +1,53 @@
...
-```c++
-void hal_deepsleep(void);
+```
+mbed test -t <toolchain> -m <target> -n "tests-mbed_hal-sleep*"
 ```
 
-To enable sleep support in Mbed OS, you need to add the `SLEEP` label in the `device_has` option of the target's section in the `targets.json` file.
+You can read more about the test cases:
+
+[![View code](https://www.mbed.com/embed/?type=library)](https://github.com/ARMmbed/mbed-os/blob/feature-hal-spec-sleep/TESTS/mbed_hal/sleep/sleep_api_tests.h)
```
AnotherButler: I can't find this in the feature branch Doxygen.

docs/reference/contributing/target/sleep.md
```diff
@@ -1,46 +1,53 @@
...
 
-- Sleep.
+[![View code](https://www.mbed.com/embed/?type=library)](https://os-doc-builder.test.mbed.com/docs/v5.7/mbed-os-api-doxy/sleep__api_8h_source.html)
 
-```c++
-void hal_sleep(void);
-```
+To make sure your platform is ready for the upcoming changes, you need to implement the future API and submit it in a separate pull request against `feature-hal-spec-sleep` branch. You can find the API and specification for the new Sleep API in the following header file:
+
+[![View code](https://www.mbed.com/embed/?type=library)](https://github.com/ARMmbed/mbed-os/blob/feature-hal-spec-sleep/hal/sleep_api.h)
```
AnotherButler: I can't find this in the feature branch Doxygen.

PR #392, opened 2018-02-01
Mbed OS 2 to 5 Migration Guide

yennster:
This PR includes a tutorial/write-up of a few errors that can occur when migrating a component or program from Mbed OS 2 to Mbed OS 5 and how to mitigate them. It also details information on how you can identify if your program uses Mbed OS 2 or Mbed OS 5.

This document was created in collaboration with @sarahmarshy


docs/tutorials/migrate/migrate_from_mbed_os_2_to_5.md
```diff
@@ -0,0 +1,308 @@
...
+
+Some repositories may have both an `mbed.bld` file and a `mbed-rtos.lib` file present. Mbed-RTOS is the precursor to Mbed OS 5, which combines the older Mbed OS 2 library with Mbed-RTOS. So, Mbed OS 5 can replace BOTH `mbed.bld` and `mbed-rtos.lib`.
+
+### Migrating to Mbed OS 5
+
+There are two possible outcomes when trying to migrate to Mbed OS 5:
+
+1. Replacing the old Mbed library with Mbed OS "just works," as the APIs in the library have not changed. In this instance, you're finished.
+2. Replacing the old Mbed library with Mbed OS produces some compilation errors. For one of two reasons:
+  - The API calls in the **library** are out of date, and you need to migrate to Mbed OS API syntax.
```
sarahmarshy: The example you chose fails to compile because of target specific code in the library, so the "migrating to Mbed OS API syntax" doesn't apply to that case.

yennster: I think this section is just talking about general cases right? Not the example yet?

sarahmarshy: We could try to find a case of out of date API calls.

docs/tutorials/migrate/migrate_from_mbed_os_2_to_5.md
```diff
@@ -0,0 +1,308 @@
...
+mbed compile -m ublox_evk_odin_w2 -t gcc_arm
+```
+
+It successfully compiles, so no changes are necessary to the Grove - Buzzer library or Hello World program.
+
+### Example component #2 - application fails to compile
+
+- Component: [SRF08-Ultrasonic-Range-Finder](https://os.mbed.com/components/SRF08-Ultrasonic-Range-Finder/).
+- Hello World repo: [SRF08HelloWorld](https://os.mbed.com/users/melse/code/SRF08HelloWorld/).
+- Platform: [FRDM-K64F](https://os.mbed.com/platforms/FRDM-K64F/).
```
sarahmarshy: I think it should be more clear that the platform/toolchain listed are what are using to compile and not the only supported target/toolchain for this example.

It is obvious later in the compile command, but I think it should be clear when they are first listed.

docs/tutorials/migrate/migrate_from_mbed_os_2_to_5.md
sarahmarshy: Yes, but it makes it seem like these are the only two cases for compilation errors. There can be target specific code in application *or* library.

docs/tutorials/migrate/migrate_from_mbed_os_2_to_5.md
yennster: Would something like this work?

```
Repositories and configuration information used for this example:

- Component: repo
- ... etc.
```

sarahmarshy: I would split it into:

Repositories
- repo1
- repo2

Compile conditions/settings/something:
- GCC 
- K64F

docs/tutorials/migrate/migrate_from_mbed_os_2_to_5.md
yennster: How about this:

> 2. Replacing the old Mbed library with Mbed OS produces some compilation errors. These errors can occur in the library or application code for the following two reasons:
>  - The Mbed OS API calls are out of date, and you need to migrate to the updated Mbed OS API syntax.
>  - The application uses target specific code, and you need to be migrate it to use code for the specific target you are compiling for.

sarahmarshy: > The application uses

Change to 'there is target specific code'

yennster:
@sarahmarshy Updated with those changes

0xc0170:
Having migration doc 👍 

Not certain all errors are related to migrating (library error or targets errors are usual errors for non generic or outdated libraries). These can be applied to any version of mbed - "troubleshooting in general".

There are two paths here: it works or compiler errors, what about run time errors?  The target code is the same (if we consider using the mbed lib version that is equal to mbed OS version (same target fixes))., then what can be wrong (rtos vs non rtos)?

yennster:
@0xc0170 I see your point, however should these other errors and mediations be detailed in a separate guide? I feel that this guide's purpose was more to get the program/library compiling to 100% rather than a debugging at run-time guide(?)

@sarahmarshy

AnotherButler:
@0xc0170 We also have this document: https://os.mbed.com/docs/v5.7/tutorials/compile-time-errors.html

Would it be helpful to add more content to it and then link @yennster 's piece to it?

yennster:
@0xc0170 Note added 👍

AnotherButler:
@yennster Could you please upload all images to the "images" folder at the top level of the docs structure?

yennster:
@AnotherButler Images fixed!

AnotherButler:
Thanks so much. @yennster Is there a way for you to easily rebase this so it's against our new_engine branch? If not, I can re-create it against that branch.

yennster:
@AnotherButler I've been trying to rebase, but I think it will be easier for you to re-create it against that branch

AnotherButler:
Please see new PR #448 for new PR against test branch.

PR #398, opened 2018-02-06
Update documentation for target definitions

drewcassidy:
Adds some much-needed documentation on how to use the `config` key when
defining a new target, as well as overriding it in other targets or in
the mbed_app.json file

### DIFF
--- a/docs/tools/mbed_targets.md
+++ b/docs/tools/mbed_targets.md
@@ -6,7 +6,7 @@ Arm Mbed uses JSON as a description language for its build targets. You can find
 
 You are not allowed to redefine existing targets in `custom_targets.json`. To better understand how a target is defined, we'll use this example (taken from `targets.json`):
 
-```
+```json
     "TEENSY3_1": {
         "inherits": ["Target"],
         "core": "Cortex-M4",
@@ -20,6 +20,7 @@ You are not allowed to redefine existing targets in `custom_targets.json`. To be
         },
         "device_name": "MK20DX256xxx7",
         "detect_code": ["0230"]
+    }
 ```
 
 The style that we use for targets is:
@@ -37,7 +38,7 @@ This section lists all the properties the Mbed build system understands. Unless 
 
 The description of an Mbed target can "inherit" from one or more descriptions of other targets. When a target, called a _child_ inherits from another target, called its _parent_, the child automatically copies all the properties from the parent. After the child has copied the properties of the parent, it may then overwrite, add or remove from those properties. In our example above, `TEENSY3_1` inherits from `Target`. This is the definition of `Target`:
 
-```
+```json
 "Target": {
     "core": null,
     "default_toolchain": "ARM",
@@ -59,7 +60,7 @@ A child target may add properties that don't exist in any of its parents. For ex
 
 It's possible, but discouraged, to inherit from more than one target. For example:
 
-```
+```json
 "ImaginaryTarget": {
     "inherits": ["Target", "TEENSY3_1"]
 }
@@ -103,7 +104,7 @@ When a target inherits, it's possible to alter the values of `macros` in the chi
 
 For example:
 
-```
+```json
     "TargetA": {
         "macros": ["PARENT_MACRO1", "PARENT_MACRO2"]
     },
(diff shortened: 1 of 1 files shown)


docs/tools/mbed_targets.md
```diff
@@ -148,6 +149,40 @@ The build system errors when you use features outside of this list.
 
 When you use target inheritance, you may alter the values of `features` using `features_add` and `features_remove`. This is similar to the `macros_add` and `macros_remove` mechanism the previous section describes.
 
+#### `config` and `overrides`
+
+_configs_ provide a more flexible way to manage macros for a target. Each configuration has a macro name, as well as a default value and an optional help value. For example, this config flag defines a configuration `lf_clock_src` that defines the source for the low frequency clock on the nRF51 target, with a default value of `NRF_LF_SRC_XTAL`. Compiling sets this value for the macro `MBED_CONF_NORDIC_NRF_LF_CLOCK_SRC`:
```
AnotherButler: Query: In "a more flexible way", what are we comparing this to? More flexible than what?

Also, could you please clarify the third sentence? When we say "this config flag," are we talking about the example starting on line 156 or something else?

Also, is this example specific to the nRF51, or does it work with other boards, too?

docs/tools/mbed_targets.md
```diff
@@ -148,6 +149,40 @@ The build system errors when you use features outside of this list.
...
+```json
+"config": {
+    "lf_clock_src": {
+        "value": "NRF_LF_SRC_XTAL",
+        "macro_name": "MBED_CONF_NORDIC_NRF_LF_CLOCK_SRC"
+    }
+}
+```
+
+_overrides_ allow a child target to change the value of a config. For example, if a child target of the nRF51 uses the internal RC clock instead of the crystal, it can add an override:
```
AnotherButler: Query: Is this example specific to the nRF51, or does it work for other boards, too?

docs/tools/mbed_targets.md
drewcassidy: I was referencing the section of the style guide below on why modules and boards shouldnt use `macros`. The example is just a config definition I pulled from the NRF51 target, but I'll replace it with something more generic. I should probably move the explanation after the example

docs/tools/mbed_targets.md
```diff
@@ -290,9 +330,9 @@ For each of these target roles, some restrictions are in place:
   - `extra_labels`.
   - `public`.
   - `config`.
+  - `override`
```
0xc0170: isn't this `overrides` (s missing in here) ?

docs/tools/mbed_targets.md
```diff
@@ -148,6 +149,45 @@ The build system errors when you use features outside of this list.
...
+"config": {
+    "clock_src": {
+        "help": "Clock source to use, can be XTAL or RC",
+        "value": "XTAL",
+        "macro_name": "CLOCK_SRC"
+    }
+}
+```
+
+ This case defines the config `clock_src`, for the `CLOCK_SRC` macro, and has a default value of `XTAL`.
```
0xc0170: wouldn't be easier to show here default naming convention for config and also mention that macro_name overrides it ?

drewcassidy: ah, I didnt even realize it defaults to the config name.

docs/tools/mbed_targets.md
```diff
@@ -151,19 +151,23 @@ When you use target inheritance, you may alter the values of `features` using `f
...
+    "clock_freq": {
+        "help": "Clock frequency in Mhz",
+        "value": "16",
+        "macro_name": "CLOCK_FREQUENCY_MHZ"
     }
 }
 ```
 
- This case defines the config `clock_src`, for the `CLOCK_SRC` macro, and has a default value of `XTAL`.
+ This case defines the config `clock_src` with the default value of `XTAL` for the macro `CLOCK_SRC`, and the config `clock_freq` with the default value of 16 for the macro `CLOCK_FREQUENCY_MHZ`.
```
0xc0170: have you tested it ? Ibelieve the default name for clock_src value will be `MBED_CONF_CLOCK_SRC` ?

drewcassidy: I didnt realize there was a standardized macro for that

drewcassidy: would something like the bluetooth advertising name be a better example maybe?

drewcassidy:
So while trying to find a better config example I found [this page from the older docs](https://docs.mbed.com/docs/mbed-os-handbook/en/5.2/advanced/config_system/), should we just reuse the same thing? it looks like its still up to date

AnotherButler:
We do have this page about the configuration system: https://os.mbed.com/docs/v5.7/tools/configuring-tools.html

Would it make sense to link to it from the mbed_targets page?

drewcassidy:
ah, yeah I just found that. That would probably make sense with just a brief description of the system in `mbed_targets`

it's a bit hard to find these documentation pages through google because of all the older versions showing up. I'm wonder if there should be a change to the docs website's `robots.txt`

AnotherButler:
@drewcassidy I've added the link to the config page. I also changed this to point to our test site (the new_engine branch), so we can test it before we make it live. Could you please rebase this to resolve the merge conflicts?

AnotherButler:
Closing because I'm copying these changes against new_engine

PR #399, opened 2018-02-06
Add documentation for NVStore

davidsaada:


docs/reference/api/storage/NVStore.md
```diff
@@ -0,0 +1,15 @@
+## NVStore
+
+NVStore is a lightweight module that stores data by keys in the internal flash for security purposes. For each item key, the NVStore module provides the ability to set the item data or get it. Newly set values are added to the flash (as in journal), with the effect of overriding the previous value for this key. The NVStore module ensures that power failures don't harm existing data during any operation.
```
AnotherButler: Query: What adds newly set values to the flash, and what does "as in journal" mean?

davidsaada: Setting the data is achieved via the set API, and same goes for getting via the get API. 
As in journal means that newly added values are added to the end of the current data, overriding the previous value that was there for the same key.

davidsaada:
Thanks Amanda, nothing wrong with your changes.
I have created the mbed-os-example-nvstore repo, and modified the README.md file to point there. main.cpp in this repo is the example file. I guess that the upload to os.mbed.com and pointing to this file is your work (otherwise please let me know how this should be done).
So, seems that all that's left is the doxygen of the nvstore APIs. Just need to know how to filter out the doxygen output of unwanted classes, and how to send it to you.

AnotherButler:
I'll merge this after the code goes through and you update the reference and example links.

dannybenor:
Seems to me that this is also related to the changes in the tools. What is the status of the work Jimmy started? In any case seems like a non-trivial doc that we will start writing. Does not seems to fit the release schedule

AnotherButler:
@theotherjimmy Can you comment?

PR #408, opened 2018-02-13
Add BL parameter definitions as new document

theotherjimmy:
@AnotherButler, where should we link this from?

### DIFF
--- a/docs/tools/bootloader.md
+++ b/docs/tools/bootloader.md
@@ -23,7 +23,7 @@ This configuration parameter conflicts with `target.bootloader_img` and `target.
 
 ### `target.mbed_app_size`
 
-This parameter defines the size of your application. You may use `target.mbed_app_start` in conjunction with this parameter to set the start address, as well as the size. When `target.mbed_app_start` is not present, the application starts at the beginning of ROM.
+This parameter defines the size of your application. You may use `target.mbed_app_start` in conjunction with this parameter to set the start address, as well as the size. When `target.mbed_app_start` is not present, the application starts at the beginning of ROM. You are responsible for the alignment of the end address of the application with respect to the flash layout and vector table size of the MCU you are using.
 
 The value of this parameter is available to C and C++ as `APPLICATION_SIZE` and to the linker as `MBED_APP_SIZE`.
 
@@ -39,7 +39,7 @@ You may use this parameter in conjunction with `target.restrict_size`. It confli
 
 ### `target.restrict_size`
 
-This parameter restricts the size of the application to be at most the specified size. When `target.bootloader_img` is present, the start of the current application's code segment is computed as above; otherwise, the start address is the beginning of ROM. The size of the application is computed by rounding the end address down to the nearest flash erase block boundary and subtracting the start address. The postbuild merge process pads the resulting bootloader binary to its end address.
+This parameter restricts the size of the application to be at most the specified size rounded down to the nearest integer multiple of flash erase blocks. When `target.bootloader_img` is present, the start of the current application's code segment is computed as above; otherwise, the start address is the beginning of ROM. The postbuild merge process pads the resulting bootloader binary to its end address.
 
 The start address of the current application, as computed above, is available to C and C++ as `APPLICATION_ADDR` and to the linker as `MBED_APP_START`. The size of the current application is available as `APPLICATION_SIZE` and `MBED_APP_SIZE`. This parameter also defines `POST_APPLICATION_ADDR` and `POST_APPLICATION_SIZE` as the start address and size of the region after the application.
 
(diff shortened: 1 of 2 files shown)


docs/tools/bootloader.md
```diff
@@ -0,0 +1,48 @@
...
+For an guide on using the managed and unmanaged boot loader modes, see [the boot loader turorial](../tutorials/bootloder.md).
+
+There are 4 configuration parameters that affect the location of an application in ROM:
+ * `target.mbed_app_start`
+ * `target.mbed_app_size`
+ * `target.bootloader_img`
+ * `target.restrict_size`
+ 
+All of these parameters are valid in `targets.json`, `mbed_lib.json` and `mbed_app.json` and may be defined for individual targets and the wildcard target, `*`. A parameter in `mbed_lib.json` overrides a parameter in `targets.json`, and a parameter in `mbed_app.json` overrides one in `mbed_app.json`.
+
```
drahnr: Is this intended? I assume one should be `mbed_lib.json`.

theotherjimmy: Ha! Good catch. Yes, the second `mbed_app.json` should be `mbed_lib.json` in the last sentence.

theotherjimmy: Fixed with a rebase.

docs/tools/bootloader.md
```diff
@@ -0,0 +1,56 @@
...
+The start address of the current application, as computed above, is available to C and C++ as `APPLICATION_ADDR` and to the linker as `MBED_APP_START` ; The size of the current application is available as `APPLICATION_SIZE` and `MBED_APP_SIZE`. This parameter also defines `BOOTLOADER_ADDR` and `BOOTLOADER_SIZE` as the start address and size of the provided boot loader.
+
+This parameter may be used in conjunction with `target.restrict_size` and conflicts with `target.mbed_app_start` and `target.mbed_app_size`. When used with `target.restrict_size`, the size off the application is defined by that parameter; Otherwise, the size is the remainder of ROM.
+
+### `target.restrict_size`
+
+This parameter restricts the size of the application to be at most the specified size. When `target.bootloader_img` is present, the start of the current application's code segment is computed as above; Otherwise, the start address is the beginning of ROM. The size of the application computed by rounding the end address down to the nearest flash erase block boundary and subtracting the start address. The post-build merge process will pad the resulting boot loader binary to it's end address.
```
marcuschangarm: I'm not completely sure what is the difference between `restrict_size` and `mbed_app_size`? It seems like you can use them for the same thing: define the size of your application?

I also don't understand the rounding down part: "The size of the application computed by rounding the end address down to the nearest flash erase block boundary and subtracting the start address"

docs/tools/bootloader.md
```diff
@@ -0,0 +1,56 @@
...
+
+The start address of the current application, as computed above, is available to C and C++ as `APPLICATION_ADDR` and to the linker as `MBED_APP_START` ; The size of the current application is available as `APPLICATION_SIZE` and `MBED_APP_SIZE`. This parameter also defines `POST_APPLICATION_ADDR` and `POST_APPLICATION_SIZE` as the start address and size of the region after the application.
+
+This parameter may be used in conjunction with `target.bootloader_img` and conflicts with `target.mbed_app_start` and `target.mbed_app_size`. When used with `target.bootloader_img`, the start off the application is defined by that parameter; Otherwise, the start is the start of ROM.
+
+### Exporter limitations
+
+While the exporters can export a project using the configuration parameters above, there are some limitations. 
+
+The exporters do not interpret Mbed OS Configuration, and any changes to configuration parameters, especially boot loader parameter, require you to re-run the `mbed export` command.
```
marcuschangarm: This problem is common to all configuration changes, right? It sounds like the bootloader is a special case.

theotherjimmy: That's true. You could, in theory, modify `mbed_config.h` for other configuration parameters. BL configs don't even allow that.

docs/tools/bootloader.md
theotherjimmy: I'll have to think about how to phrase that better. The difference is that `restrict_size` will generate the `POST_APPLICATION_*` stuff and pad your executable to the size mentioned.

marcuschangarm:
Much better! It was easy to understand and didn't leave any lingering questions behind!

docs/tutorials/bootloader.md
```diff
@@ -99,12 +99,18 @@ For an example showing how to create an application that uses a bootloader, see
...
 
-The configuration option `target.mbed_app_size` defines the size of an application image in ROM by defining the `MBED_APP_SIZE` macro for the linker script. You may only define this configuration option on a per-target basis defined within the `target_overrides` section of an Mbed application configuration, and you may not define it for the metatarget `*`. When you do not define this configuration option, it defaults to the remaining ROM. The Mbed OS tools calculate the remaining ROM by subtracting the image's offset into ROM from the total size of ROM. Together with `target.mbed_app_start`, these configuration options define a continuous region of memory that an image may use. The tools verify that this region of memory is in ROM, but the tools do not perform any other checks for consistency or validity.
+Further, the exporters do not implement the postbuild merge that bootloader builds use. 
+
+For a managed bootloader:
+After exporting a project with the `target.mbed_app_start` setting, you are responsible for ensuring that a boot loader is present, if needed. Without flashing this boot loader image, the device will not boot correctly.
```
theotherjimmy: This is an unmanaged BL parameter, and the unmanaged BL description. This should be swapped with the subsequent paragraph, without changing the title.

theotherjimmy:
Much better.

theotherjimmy:
Good to go!

PR #411, opened 2018-02-14
PR template addition

0xc0170:
We would like to show users how a good pull request look like.  I used code highlight for the template, is that fine or better solutions ?

The main points that should be coevred:
- describe your changes (what is being changed and how)
- how it was tested (targets, toolchains, etc)
- pull request type (can be one of, not a mixture!)

The template is being updated here https://github.com/ARMmbed/mbed-os/pull/6078

### DIFF
--- a/docs/reference/contributing/guidelines/workflow.md
+++ b/docs/reference/contributing/guidelines/workflow.md
@@ -51,6 +51,30 @@ Pull requests on GitHub have to meet the following requirements to keep the code
 - Prefixing your commit message with a domain is acceptable, and we recommend doing so when it makes sense. However, prefixing one's domain with the name of the repo is not useful. For example, making a commit entitled "mbed-drivers: Fix doppelwidget frobulation" to the `mbed-drivers` repo is not acceptable because it is already understood that the commit applies to `mbed-drivers`. Renaming the commit to "doppelwidget: Fix frobulation" would be better, if we presume that "doppelwidget" is a meaningful domain for changes, because it communicates that the change applies to the doppelwidget area of `mbed-drivers`.
 - All new features and enhancements require documentation, tests and user guides for us to accept them. Please link each pull request to all relevant documentation and testing pull requests.
 
+##### Pull request template
+
+A good example how a pull request template should look like.
+
+```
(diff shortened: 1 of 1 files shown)


docs/reference/contributing/guidelines/workflow.md
```diff
@@ -51,6 +51,30 @@ Pull requests on GitHub have to meet the following requirements to keep the code
...
+A good example how a pull request template should look like.
+
+```
+Fix deep sleep locking bug
+
+# Description
+
+Fix problems which could leave deep sleep locked unintentionally, along with adding tests to verify this behavior is fixed.
+
+Tested locally with two targets, all toolchains.
```
cmonr: I would recommend we be specific here, to encourage others to be specific as well.

Replace "two targets" with something like "K64F and K66F".

cmonr: I would recommend we be specific here, to encourage others to be specific as well.

Replace "two targets" with something like "K64F and K66F".

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -51,6 +51,30 @@ Pull requests on GitHub have to meet the following requirements to keep the code
...
+
+Test results can be seen [here](just an example). 
+
+# Pull request type
+
+- [x] Fix
```
cmonr: Like the checklist.

0xc0170:
@adbridge @theotherjimmy @kjbracey-arm Please review

0xc0170:
> Agree with Cruz about being specific as to which two targets were tested against. But otherwise looks like a nice simple example. Can always be refined down line, good to get something in.

I had it first (named two targets I randomly chose) , but did not want to pick up only two targets (which ones we would select), lets see how this goes and can add some info about being more specific

AnotherButler:
Is it OK to merge this before 6078 merges?

cmonr:
@AnotherButler Yup! It even seems like that PR is waiting on this! https://github.com/ARMmbed/mbed-os/pull/6078#issuecomment-365852699

PR #416, opened 2018-02-23
Mbed OS 5.8 brings support for network interface status change callback

VeijoPesonen:
registration.


kjbracey-arm:
Fine, but wondering if the example should be setting non-blocking connect mode to kind of "get the value". (Not that it would lead to a visible difference in output)

VeijoPesonen:
Updated based on Kevin's feedback.

PR #423, opened 2018-02-28
Sleep handbook improvements

bulislaw:
* Rename sleep page to power management - as we are looking into adding more related functions (like reset).
* General text improvements

@AnotherButler


docs/reference/api/platform/PowerManagement.md
```diff
@@ -0,0 +1,65 @@
...
+    // do nothing
+}
+
+void main()
+{
+    rtos_attach_idle_hook(&new_idle_loop);
+}
+```
+
+### Function reference
```
AnotherButler: Query: Should this really be at the same level as idle loop? Also, should this be moved up? Right now, it looks like it's the function reference of idle loop, but it's the function reference of sleep.

docs/reference/api/platform/PowerManagement.md
```diff
@@ -0,0 +1,65 @@
...
+
+[![View code](https://www.mbed.com/embed/?type=library)](http://os-doc-builder.test.mbed.com/docs/v5.7/mbed-os-api-doxy/mbed__sleep_8h_source.html)
+
+### Example
```
AnotherButler: See above comment.

docs/reference/api/platform/PowerManagement.md
```diff
@@ -0,0 +1,65 @@
...
+
+The system clock to the core stops until a reset or an interrupt occurs. This eliminates dynamic power that the processor, memory systems and buses use. This mode maintains the processor, peripheral and memory state, and the peripherals continue to work and can generate interrupts.
+
+You can wake up the processor by any internal peripheral interrupt or external pin interrupt.
+
+2. Deep sleep mode
+
+This mode is similar to sleep but saves more power and has a longer wakeup time. It saves additional power by turning off the high-speed clocks. Because of this, you can only enter this mode when peripherals relying on high-speed clocks are not in use. Peripherals that do not rely on high-speed clocks include the LowPowerTicker, RTC and InterruptIn APIs. This mode maintains all state.
+
+#### Sleep manager
```
AnotherButler: Query: Are there supposed to be two sections called "Sleep manager"? This seems confusing to me.

docs/reference/api/platform/PowerManagement.md
bulislaw: It'll be reference for the power management API. That's why it has the same level as the other sections.

bulislaw:
Some reshuffling:
* I've kept the Sleep, Idle loop and Function reference (and reset via second PR) at level 3.
* I've moved example to sleep section as it's sleep only related.
* I've added extra reference section to idle loop as it's part of RTOS doxy group

I'm not sure it's the right outline, but it makes sense I think. Power management has 3 functionalities, two of which live in `platform/` (reset, sleep) and one in `rtos/` (idle loop) both the code and the doxy groups. So the final toc will look like (including #424):
* Sleep
  * ...
  * Example
* Reset
* Reference
* Idle loop
  * Reference

If you have any ideas how to make it better please do let me know :)

AnotherButler:
It looks like 424 overrides this one. Can I close this one?

AnotherButler:
Never mind. I just saw the dependency note on the next one.

AnotherButler:
None of these follow our template, so I don't want to merge them.

bulislaw:
What's our template?

AnotherButler:
This is our template: https://docs.internal.mbed.com/docs/master/engineers/mbed-os-user-api-class-reference.html

Could we move idle loop to its own page and create an example for it? I think that would help this page. Also, I notice in the Doxygen, it's "idle hook" rather than "idle loop". Which is correct?

AnotherButler:
@bulislaw Could you please solve the merge conflicts?

PR #424, opened 2018-02-28
System reset docs

bulislaw:
# Dependencies
- [ ] #423 Sleep page -> power management page
- [ ] https://github.com/ARMmbed/mbed-os/pull/6230 system_reset implementation

# Changes

The actual full diff, after the #423 will be merged:

```
+### System reset
+
+Mbed OS provides a standardized call to power cycle the system:
+
+```c++
+void system_reset();
+```
+
+After the call the processor and most components will reset, but it will not affect the debug subsystem.
```
I think the https://github.com/ARMmbed/mbed-os/pull/6230 won't make it for 5.8, that's why I've decided to split out the sleep page improvements out of this PR into #423 which can be independently merged for 5.8.

@anotherbutler


docs/reference/api/platform/PowerManagement.md
```diff
@@ -0,0 +1,79 @@
...
+- `SPI`.
+- `I2C`.
+- `CAN`.
+- `SerialBase`.
+
+#### Example
+
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/SleepManager_Example_1/)](https://os.mbed.com/teams/mbed_example/code/SleepManager_Example_1/file/e85412b4147e/main.cpp)
+
+### System reset
```
AnotherButler: Should system reset be the same level as sleep, or is it part of sleep? If the former, it feels out of place to me that it would be between the sleep description and class reference.

bulislaw: It's at the same level and the function reference will point to the shared doxygen group "Power management" once doxygen is regenerated.

PR #439, opened 2018-03-06
Add script for updating notes to most recent doc pages

geky:
This script generates a note for each doc page that is out of date, with a link that takes users to the most recent doc page. If it can't find a recent doc page, it will ask the user which url to use. It uses existing notes and remembers answers in an effort to reduce the number questions forwarded to the user.

Originally I attempted to parse docs.json but could not get this to work well enough.

note: Don't merge until first successful run
cc @AnotherButler, @kegilbert

### DIFF
--- a/check_tools/point_to_new_version.py
+++ b/check_tools/point_to_new_version.py
@@ -1,0 +1,188 @@
+#!/usr/bin/env python
+
+import re
+import os
+import requests
+import urllib
+import json
+from requests.auth import HTTPBasicAuth
+from subprocess import check_output
+
+# NOTE Does not follow docs.json, instead relies heavily on
+# remembering successful urls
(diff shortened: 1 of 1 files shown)


AnotherButler:
@geky Is this OK to merge, or are we waiting on something?

geky:
Hmmm, I'm not really sure. Does the script still work?

AnotherButler:
Can I tell without merging it?

geky:
If there's no one using it, then it's probably not worth merging at this point

AnotherButler:
I think there's a use case for this. I would like to merge it if it's ready. I know at one point we were waiting on something before merging it, so I wanted to follow up and make sure it's OK to go now.

PR #442, opened 2018-03-08
Typo Fix 3

jen3andruska:
I'm not sure if 'Instantiate' was intended or if this should say 'Initiate'?


AnotherButler:
Actually, this one should be instantiate. Good eye, though.

jen3andruska:
Ah ok, sounds good.

PR #446, opened 2018-03-09
Contr API Ref Typos

jen3andruska:
Just a few typos.

I'm assuming that the phrase 'ticker times increments' is meant to be singular 'time' that increments, but they may also mean that ticker 'times' (plural) increment (without the 's'). Either way, one needs to be changed.


AnotherButler:
I updated some links in between your making this PR and my reviewing it. To avoid merge conflicts, I made those changes to this PR, too. Just FYI. I don't want you to see this and think I made a lot of changes to your PR. It looks fab.

PR #452, opened 2018-03-13
Rename LOWPOWERTICKER to LPTICKER

bulislaw:
Rename  LOWPOWERTICKER to LPTICKER HAL macro name to be consistent with USTICKER.

Depends on: https://github.com/ARMmbed/mbed-os/pull/6351


bulislaw:
@AnotherButler the code dependency was merged, can we merge that and regenerate the handbook please.

PR #460, opened 2018-03-19
TBD corrections 3

jen3andruska:
There were a few things that were unclear in this section:

In the 1st paragraph of the Security Manger section:
The phrase 'process of pairing and optionally bonding' in the 2nd sentence is a little confusing. Are these two parts of one process?, i.e. 'the pairing and optionally bonding process provides this'?  Or, are they two separate processes? - in which case it should say: 'The pairing and optionally bonding processes provide this.' (I took a guess from the context that they are two separate but related processes, and that the latter was meant?)

2nd paragraph of the Security Manager section:
I changed 'process of pairing' to 'pairing process' again (in the 3rd paragraph it is called the 'pairing process' and I thought this might sound better?)
The next sentence (2nd paragraph, 2nd sentence) - I can't tell what it is trying to say? (there must be words missing/added?) . I guessed that the Security Manager provides these keys listed, but I couldn't tell what 'handle these' at the end fo the sentence might refer to?

3rd paragraph:
I changed 'may' because it sounded like it may/may not provide this - so I changed it to 'can'

### DIFF
--- a/docs/reference/api/connectivity/bluetooth/SecurityManager.md
+++ b/docs/reference/api/connectivity/bluetooth/SecurityManager.md
@@ -2,9 +2,9 @@
 
 SecurityManager deals with authentication and encryption for the Bluetooth Low Energy link. The pairing and optionally bonding processes provide this. The SecurityManager achieves bonding by saving the pairing information and reusing it on subsequent reconnections. This saves time because the pairing does not have to be performed again.
 
-The pairing process may produce a set of keys to be used during current or later connections. The SecurityManager includes the Long Term Encryption Key (LTK), the Identity Resolving Key (IRK) and the Connection Signature Resolving Key (CSRK). The SecurityManager uses the LTK to encrypt subsequent connections without having to pair again. The Link Controller uses IRK to identify peers who use random resolvable addresses. The application uses CSRK to sign and authenticate signed data.
+The pairing process may produce a set of keys to be used during current or later connections. The SecurityManager handles these, and they include the Long Term Encryption Key (LTK), the Identity Resolving Key (IRK) and the Connection Signature Resolving Key (CSRK). The SecurityManager uses the LTK to encrypt subsequent connections without having to pair again. The Link Controller uses IRK to identify peers who use random resolvable addresses. The application uses CSRK to sign and authenticate signed data.
 
-The pairing process can provide man-in-the-middle protection (MITM). The SecurityManager achieves this through various means, including out of band communication, depending on the capabilities of the local and peer device.
+The pairing process may provide man-in-the-middle protection (MITM). The SecurityManager achieves this through various means, including out of band communication, depending on the capabilities of the local and peer device.
 
 The SecurityManager stores the keys, permanently if possible, to speed security requests on subsequent connections.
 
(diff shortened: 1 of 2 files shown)


paul-szczepanek-arm:
You can leave comments on github attached to lines, then you won't have to specify the paragraph.

docs/reference/api/connectivity/bluetooth/SecurityManager.md
```diff
@@ -1,10 +1,10 @@
 ## SecurityManager
 
-SecurityManager deals with authentication and encryption for the Bluetooth Low Energy link. The process of pairing and optionally bonding provides this. The SecurityManager achieves bonding by saving the pairing information and reusing it on subsequent reconnections in order to save time by not having to perform pairing again.
+SecurityManager deals with authentication and encryption for the Bluetooth Low Energy link. The pairing and optionally bonding processes provide this. The SecurityManager achieves bonding by saving the pairing information and reusing it on subsequent reconnections. This saves time because the pairing does not have to be performed again.
```
paul-szczepanek-arm: Pairing always happens. Bonding may happen. Bonding means: saving pairing information for later use so that you don't have to pair again. There is no extra communication between the devices. One device may just decide to "bond" with the other which means nothing else than just save the result of the pairing process which otherwise would have to be redone on subsequent reconnections.

docs/reference/api/connectivity/bluetooth/SecurityManager.md
```diff
@@ -1,10 +1,10 @@
...
 
-The process of pairing may produce a set of keys to be used during current or later connections. The SecurityManager and include the Long Term Encryption Key (LTK), the Identity Resolving Key (IRK) and the Connection Signature Resolving Key (CSRK) handle these. The SecurityManager uses the LTK to encrypt subsequent connections without having to pair again. The Link Controller uses IRK to identify peers who use random resolvable addresses. The application uses CSRK to sign and authenticate signed data.
+The pairing process may produce a set of keys to be used during current or later connections. The SecurityManager includes the Long Term Encryption Key (LTK), the Identity Resolving Key (IRK) and the Connection Signature Resolving Key (CSRK). The SecurityManager uses the LTK to encrypt subsequent connections without having to pair again. The Link Controller uses IRK to identify peers who use random resolvable addresses. The application uses CSRK to sign and authenticate signed data.
```
paul-szczepanek-arm: The first two sentences were originally:
"The process of pairing may produce a set of keys to be used during current or later connections. These are handled by the SecurityManager and include the Long Term Encryption Key (LTK), the Identity Resolving Key (IRK) and the Connection Signature Resolving Key (CSRK)."
Something must've gotten lost in copy edits.

docs/reference/api/connectivity/bluetooth/SecurityManager.md
```diff
@@ -1,10 +1,10 @@
...
 
-The pairing process may provide man-in-the-middle protection (MITM). The SecurityManager achieves this through various means, including out of band communication, depending on the capabilities of the local and peer device.
+The pairing process can provide man-in-the-middle protection (MITM). The SecurityManager achieves this through various means, including out of band communication, depending on the capabilities of the local and peer device.
```
paul-szczepanek-arm: MITM protection may/can be provided by the SM if it's possible to do so. That's not always the case, some combinations of devices simply don't have the required capabilities

docs/reference/api/connectivity/bluetooth/SecurityManager.md
jen3andruska: Ok this is what I thought - they are different processes but related. This makes sense, thanks!

docs/reference/api/connectivity/bluetooth/SecurityManager.md
jen3andruska: Ok I see, so it really is 'may' then.  This makes sense.

Issue #462, opened 2018-03-20
clarify configuration


bmcdonnell-ionx:
I don't understand how to use the Mbed OS configuration system. I think the [reference documentation](https://os.mbed.com/docs/v5.7/reference/configuration.html) regarding it is unclear. If I'm missing something(s) obvious, please point me to it. Otherwise, please clarify the documentation to answer questions such as those I pose below. (Also, I'd appreciate some quick answers here, if possible.)

For instance, the page linked above provides an example of how to configure the stack size for the main thread, by putting this in `mbed_app.json`:

```
{
    "config": {
        "main-stack-size": {
            "value": 2000
        }
    }
}
```

But how does this work? What does it do? I couldn't find "`main-stack-size`" anywhere in the `mbed-os` or `mbed-cli` code trees. (I used `grep -R`.)

On further searching (`grep -Ri "main[^ ]stack[^ ]size" .`), I find the conditional `#define MBED_CONF_APP_MAIN_STACK_SIZE` in `mbed_boot.c`, so I infer that `lower-case-hyphenated` labels in `.json` files get somehow translated to `MBED_CONF_ALL_CAPS` `#define`s - but how? Where is the code that does this so I can learn more when the documentation falls short, or I want to improve something? How/where do the custom `#define`s I configure get made visible to my code? (e.g. In makefiles, or is a new header file generated? What about when I use `mbed export`?)

Where can I find the list of configurable items?

How does the info on the [Platform](https://os.mbed.com/docs/v5.7/reference/configuration-platform.html), [Drivers](https://os.mbed.com/docs/v5.7/reference/configuration-drivers.html), [RTOS](https://os.mbed.com/docs/v5.7/reference/configuration-rtos.html), and [Connectivity](https://os.mbed.com/docs/v5.7/reference/configuration-connectivity.html) pages translate to what I should put in a `.json` file? And how do the `.json` files get named?

For instance, the Platform page begins with this:

```
Configuration parameters
------------------------

Name: platform.default-serial-baud-rate
    Description: Default baud rate for a Serial or RawSerial instance (if not specified in the constructor)
    Defined by: library:platform
    Macro name: MBED_CONF_PLATFORM_DEFAULT_SERIAL_BAUD_RATE
    Value: 9600 (set by library:platform)
```

[...]

PR #463, opened 2018-03-21
Overhaul the configuration system documentation

theotherjimmy:


theotherjimmy:
@bmcdonnell-ionx and @loverdeg-ep Could you review these changes? Could you comment on anything that you have remaining questions about? Could you comment about the ordering of the information?

docs/tools/config_system.md
```diff
@@ -34,28 +147,40 @@ For example:
...
+You define a macro the short way by using a `value` instead of the deception object, such as `param3` above. Defining a parameter the short way is equivalent to a configuration parameter defined the long way with the key `help` unset, the key `macro_name` unset, the key `required` set to `false`, and the key `value` set to the value in place of the description object. Note that a JSON object is not a valid value, and there is no ambiguity between short and long way configuration parameter definitions
 
 <span class="notes">**Note:** the name of a parameter in `config` can't contain a dot (`.`) character.</span>
 
-The configuration system automatically appends an *implicit prefix* to the name of each parameter, so you don't have to worry about a name clash if you define a parameter with the same name in a library and a target, for example. The implicit prefix is:
+The configuration system appends a prefix to the name of each parameter, so a parameter with the same name in a library does not conflict with parameters of the same name in targets or other libraries. The prefix is:
```
loverdeg-ep: good to know

docs/tools/config_system.md
```diff
@@ -34,28 +147,40 @@ For example:
...
 
-- **target.** if the parameter is defined in a target.
-- **app.** if the parameter is defined in the application.
-- The name of the library followed by a dot (.) if the parameter is defined in a library.
+| Location | Prefix |
+| -------- | ------ |
+| target | `target.` |
+| any library | The name of the library followed by a dot (.) |
```
loverdeg-ep: For my own education; how is a library name specified? 'library-name.lib'?

docs/tools/config_system.md
```diff
@@ -34,28 +147,40 @@ For example:
...
+| Section | Required | Meaning |
+| ------- | -------- | ------- |
+| `name`  | Yes      | name of the library. Must be unique. May not be `app` or `target`.|
+| `macros` | No | List of macros to define on the command line. |
+| `config` | No | Configuration parameters defined for use in this library. |
+| `target_overrides` | No | Overrides for target configuration parameters and configuration parameters of the current library. |
+
+The following is an example library, "mylib".
+
+```JSON
```
loverdeg-ep: good stuff

docs/tools/config_system.md
```diff
@@ -91,87 +217,31 @@ In this JSON file:
...
-        }
-    },
-    "overrides": {
-        "stack_size": 256
-    }
-}
-```
-
-`Derived` above defines its own configuration parameter called `my_own_config` and inherits the configuration parameters from `Base`, so its configuration parameters are `serial_console_speed`, `stack_size` and `my_own_config`. It also overrides the value of the `stack_size` parameter defined in `Base`. This means that:
+All configuration parameters defined in `mylib` have a `mylib.` prefix. Outside `mylib`, `buffer_size` is accessible using the name `mylib.buffer_size`. An application will be able to override the value of this parameter, as described in [this section](#configuration-data-in-applications).
```
loverdeg-ep: Outside myLib in another mbed_lib.json or mbed_app.json. 

'this section' probably gives the answer but two or three more words could allow me to continue reading without clicking.

docs/tools/config_system.md
```diff
@@ -91,87 +217,31 @@ In this JSON file:
...
 
-There are similarities between configuration data in applications and libraries:
 
-- Applications define their configuration parameters in the `config` section of `mbed_app.json`, as explained [in the section](#defining-configuration-parameters) about defining parameters.
-- Applications can specify target-dependent values in their `target_overrides` section, as described in the [library configuration paragraph](#configuration-data-in-libraries) (but see below for differences).
-- Applications can define macros that will be defined at compile time by declaring them in `macros`.
+### `mbed_app.json` Specification
 
-There are also a few differences:
+`mbed_app.json` may be present at the root of your project or specified as the argument to the `--app-config` parameter. Unlike library configuration, only one `mbed_app.json` is interpreted during `mbed compile` or `mbed export`. like `mbed_lib.json`, `mbed_app.json` is a JSON formatted document that contains a root JSON Object. The keys within this object are refered to as sections. The allowed sections and their meaning is listed below:
```
loverdeg-ep: Does 'project' imply application or can a project be created for a library too?

docs/tools/config_system.md
```diff
@@ -181,20 +251,17 @@ The last point above is important. The application can freely override the confi
...
-- The `*` key in `target_overrides` will match *any* target. It is possible to use the `*` key in a library's `target_overrides` too, but it'd make little sense to do so, since it will always override the values defined in the library's `config` section. In an application it might make sense to use the `*` key, since it can be used to override the configuration defined by the target or the dependent libraries, no matter which target is used for building.
+`target_overrides` works a similar to how it does in libraries. However, the application may override any configuration parameter, and it must specify configuration parameters using their prefix (like `mylib.timer_period`). If an overridden parameter doesn't have a prefix, the config system assumes that the overridden parameter is one of the parameters the application defines in its own `config` section.
```
loverdeg-ep: Wordy

docs/tools/config_system.md
```diff
@@ -181,20 +251,17 @@ The last point above is important. The application can freely override the confi
...
 
 Other than this, `target_overrides` works exactly like it does for libraries. Keys in `target_overrides` are still processed in the order they are defined, so for the example above, the `*` override is always processed first (because it matches all targets) and then `Base` is only processed for the `Base` target.
```
loverdeg-ep: kind of repetitive considering the previous addition

docs/tools/config_system.md
```diff
@@ -1,23 +1,136 @@
 ## The configuration system
 
-The Arm Mbed configuration system customizes the compile time configuration of various Mbed components (targets, libraries and applications). Each component can define a number of *configuration parameters*. The values of these configuration parameters can then be *overridden* in various ways. Configuration is defined using [JSON](http://www.json.org/). Some examples of configuration parameters:
+The Arm Mbed configuration system, a part of the Arm Mbed Build Tools, customizes compile time configuration parameters. Each component can define a number of configuration parameters in its `mbed_lib.json`. The values of these configuration parameters may be overridden by `mbed_app.json`. Configuration is defined using [JSON](http://www.json.org/). Some examples of configuration parameters:
```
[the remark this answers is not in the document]
theotherjimmy: Good point. I'll remove the use of the word component, as it does not have a clear definition in this context.

docs/tools/config_system.md
```diff
@@ -1,23 +1,136 @@
...
 
 - The sampling period for a data acquisition application.
 - The default stack size for a newly created OS thread.
 - The receive buffer size of a serial communication library.
 - The flash and RAM memory size of an Mbed target.
 
-The configuration system gathers and interprets all the configurations defined in the source tree. The output of the configuration system is a list of macros that are automatically defined when compiling the code.
+The Mbed configuration system gathers and interprets the configuration defined in the target, all `mbed_lib.json` files and the `mbed_app.json` file. The configuration system creates a single header file, `mbed_config.h`, in the `BUILD` directory that contains all of the defined configuration parameters converted into C preprocessor macros. The Mbed OS configuration system is run during `mbed compile` before invoking the compiler and during `mbed export` before creating project files.
```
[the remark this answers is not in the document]
theotherjimmy: @loverdeg-ep It's all in Mbed OS's `tools/config/` directory. This is an implementation detail that has changed in the past and I would like to avoid documenting an implementation detail. 

I'll expand on the location of `mbed_config.h`

docs/tools/config_system.md
```diff
@@ -1,23 +1,136 @@
...
+
+Do not edit `mbed_config.h` manually. It may be overwritten the next time you compile or export your project, and all your changes will be lost.
+
+### Configuration parameters in `mbed_app.json`, `mbed_lib.json`
+
+An application may have one `mbed_app.json` in the root of the project and many `mbed_lib.json` files throughout the project. When present, `mbed_app.json` may override library and target configuration parameters and define new configuration parameters. 
+
+#### Overriding configuration parameters
+
+The configuration system allows a user to override any defined configuration parameters, through the use JSON object named "target_overrides".
```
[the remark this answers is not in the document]
theotherjimmy: Thanks!

docs/tools/config_system.md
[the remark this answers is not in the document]
loverdeg-ep: Kind of off topic; Considering this how would I manage the configuration for a project that doesn't have mbedOS as a dependency?

docs/tools/config_system.md
```diff
@@ -34,28 +147,40 @@ For example:
 }
 ```
 
-The JSON fragment above defines three configuration parameters named `param1`, `param2` and `param3`. There are two ways to define a configuration parameter:
+ You define a configuration parameter by specifying it's name as the key and specifying it's value either the long way or the short way. The JSON fragment above defines three configuration parameters named `param1`, `param2` and `param3`.
```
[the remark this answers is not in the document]
theotherjimmy: I'm working on better naming here. I'll get back to this.

docs/tools/config_system.md
```diff
@@ -34,28 +147,40 @@ For example:
...
+
+You define a macro the long way using a JSON object description. Above, the configuration parameters `param1` and `param2` are defined with the long way. The following keys are supported on the description object:
+  - `help`: an optional help message that describes the purpose of the parameter.
+  - `value`: an optional field that defines the value of the parameter.
+  - `required`: an optional key that specifies whether the parameter **must** be given a value before compiling the code (`false` by default). It's not possible to compile a source tree with one or more required parameters that don't have a value. Generally, it makes sense to define a required parameter only when it doesn't have a `value` key.
```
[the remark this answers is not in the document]
theotherjimmy: :+1:

docs/tools/config_system.md
```diff
@@ -34,28 +147,40 @@ For example:
...
+  - `macro_name`: an optional name for the macro defined at compile time for this configuration parameter. The configuration system will automatically figure out the corresponding macro name for a configuration parameter, but the user can override this automatically computed name by specifying `macro_name`.
```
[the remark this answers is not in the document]
theotherjimmy: Glad to hear it. Sorry it was causing furstration.

docs/tools/config_system.md
```diff
@@ -34,28 +147,40 @@ For example:
...
-    - `required`: an optional key that specifies whether the parameter **must** be given a value before compiling the code (`false` by default). It's not possible to compile a source tree with one or more required parameters that don't have a value. Generally, it makes sense to define a required parameter only when it doesn't have a `value` key.
-    - `macro_name`: an optional name for the macro defined at compile time for this configuration parameter. The configuration system will automatically figure out the corresponding macro name for a configuration parameter, but the user can override this automatically computed name by specifying `macro_name`.
+You define a macro the short way by using a `value` instead of the deception object, such as `param3` above. Defining a parameter the short way is equivalent to a configuration parameter defined the long way with the key `help` unset, the key `macro_name` unset, the key `required` set to `false`, and the key `value` set to the value in place of the description object. Note that a JSON object is not a valid value, and there is no ambiguity between short and long way configuration parameter definitions
```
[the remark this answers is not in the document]
theotherjimmy: I meant description object; probably.

docs/tools/config_system.md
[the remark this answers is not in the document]
theotherjimmy: Working naming is "with/by a description object" and "by value"

docs/tools/config_system.md
theotherjimmy: The `name` key in the library. That's mentioned below, and I'm going to add another mention of that before you get to this point in the documentation now.

docs/tools/config_system.md
theotherjimmy: Ah. I caused some confusion. I'll correct this section.

docs/tools/config_system.md
theotherjimmy: It should imply application. I'll change project to application, across the document.

docs/tools/config_system.md
theotherjimmy: Thanks. I'll tighten it up.

loverdeg-ep: I think I mean the "works a similar to how it does in libraries" in particular.

docs/tools/config_system.md
theotherjimmy: I'll remove it!

theotherjimmy:
@loverdeg-ep I think I got to all of your comment. Thank you so much for your feedback!

docs/tools/config_system.md
```diff
@@ -34,28 +147,40 @@ For example:
...
-The configuration system automatically appends an *implicit prefix* to the name of each parameter, so you don't have to worry about a name clash if you define a parameter with the same name in a library and a target, for example. The implicit prefix is:
+The configuration system appends a prefix to the name of each parameter, so a parameter with the same name in a library does not conflict with parameters of the same name in targets or other libraries. The prefix is:
 
-- **target.** if the parameter is defined in a target.
-- **app.** if the parameter is defined in the application.
-- The name of the library followed by a dot (.) if the parameter is defined in a library.
+| Location | Prefix |
+| -------- | ------ |
+| target | `target.` |
+| any library | The name of the library, as found in the `name` section of `mbed_lib.json`, followed by a dot (.) |
```
loverdeg-ep: Since my current belief is that mbed_lib.json is optional.
What if the library doesn't use mbed_lib.json?

theotherjimmy: Then it can't define any parameters, and the prefix may remain undefined.

docs/tools/config_system.md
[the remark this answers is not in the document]
loverdeg-ep: @theotherjimmy Was hoping have the question in the previous comment resolved.

loverdeg-ep:
@theotherjimmy I think I had 2 final questions unanswered in there somewhere.

Overall seems to be a massive improvement.  Nice work!

theotherjimmy:
Thanks! I'll give your comments another pass to find those unanswered questions.

docs/tools/config_system.md
```diff
@@ -88,90 +214,34 @@ In this JSON file:
...
-        }
-    },
-    "overrides": {
-        "stack_size": 256
-    }
-}
-```
-
-`Derived` above defines its own configuration parameter called `my_own_config` and inherits the configuration parameters from `Base`, so its configuration parameters are `serial_console_speed`, `stack_size` and `my_own_config`. It also overrides the value of the `stack_size` parameter defined in `Base`. This means that:
+All configuration parameters defined in `mylib` have a `mylib.` prefix. In a `mbed_app.json`, `buffer_size` is accessible using the name `mylib.buffer_size`. An application will be able to override the value of this parameter, as described in [this section](#configuration-data-in-applications).
```
bridadan: Broken link is broken

docs/tools/config_system.md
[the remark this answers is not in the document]
theotherjimmy: > Considering this how would I manage the configuration for a project that doesn't have mbedOS as a dependency?

You could use the standalone Mbed 2 tools which contain the same config system. Would you like a mention that the config system is an Mbed OS feature, and may not work without mbed OS in your project?

docs/tools/config_system.md
loverdeg-ep: Same goes for the test framework?

loverdeg-ep: ^ ignore meant to comment on another conversation.

theotherjimmy: I'm not sure what you mean.

If you're refering to utest, It's `mbed_lib.json` is below
```JSON
{                                                                                                        
    "name": "utest",                                                                                     
    "macros": ["UNITY_INCLUDE_CONFIG_H"]                                                                 
} 
```

docs/tools/config_system.md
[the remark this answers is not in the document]
loverdeg-ep: Sure. Anything to allow me to better understand my use cases.

Test framework depends on mbedOS as well?

docs/tools/config_system.md
```diff
@@ -181,20 +251,17 @@ The last point above is important. The application can freely override the confi
...
         }
     }
 }
 ```
 
-`target_overrides` works a lot like it does in libraries, but there are a few differences:
-
-- Because the application can override any configuration parameter, it must specify configuratoin parameters using their prefix (like `mylib.timer_period`). If an overridden parameter doesn't have a prefix, the config system assumes that the overridden paramater is one of the parameters the application defines in its own `config` section.
-- The `*` key in `target_overrides` will match *any* target. It is possible to use the `*` key in a library's `target_overrides` too, but it'd make little sense to do so, since it will always override the values defined in the library's `config` section. In an application it might make sense to use the `*` key, since it can be used to override the configuration defined by the target or the dependent libraries, no matter which target is used for building.
+The application may override any configuration parameter by specifying the configuration parameters including their prefix (like `mylib.timer_period`). If an overridden parameter doesn't have a prefix, it overrides a parameter in its own `config` section.
```
bridadan: This section needs to be updated. There is no `target.serial_console_speed` introduced in the example. Also what is the `Base` target that is being referred to?

docs/tools/config_system.md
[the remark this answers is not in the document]
theotherjimmy: Yep.

docs/tools/config_system.md
theotherjimmy: XD

theotherjimmy: Superfluous sentence is superfluous. I'm just going to remove it.

docs/tools/config_system.md
```diff
@@ -120,9 +120,8 @@ cellular.use-apn-lookup = 0 (macro name: "MBED_CONF_CELLULAR_USE_APN_LOOKUP")
 
 The order in which overrides are considered is:
 
-- Configuration defined by a target overrides the configuration defined by its parent.
-- Libraries override target configuration with `mbed_lib.json`.
-- The application override target and library configuration with `mbed_app.json`
+ 1. Libraries override target configuration with `mbed_lib.json`.
+ 2. The application override target and library configuration with `mbed_app.json`
```
bridadan: Nit: override should be overrides

theotherjimmy:
Alright. I'm going to squash this history, 'cause I don't think we need to keep it.

docs/tools/config_system.md
```diff
@@ -1,23 +1,135 @@
...
+#ifndef __MBED_CONFIG_DATA__
+#define __MBED_CONFIG_DATA__
+
+// Configuration parameters
+#define MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY         0 // set by library:cellular
+#define MBED_CONF_CELLULAR_USE_APN_LOOKUP                 1 // set by library:cellular
+<file truncated for brevity>
+```
+
+<span class="notes">**Note:** A macro definition will not be generated for a parameter that doesn't have a value.</span>
```
bmcdonnell-ionx: Really? That seems to run counter to the `#ifdef` C convention for conditional compilation.

docs/tools/config_system.md
```diff
@@ -1,23 +1,135 @@
...
+
+
+The names of the macro for a configuration parameter is either a prefixed name or explicitly specified by `macro_name`. A prefixed name is constructed from the prefix `MBED_CONF_`, followed by the name of the library or `APP`, followed by the name of the parameter. The prefixed name is then capitalized and converted to a valid C macro name. For example, the `random_may_start_delay` configuration parameter in the library `cellular` is converted to `MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY`.
+
+The Mbed OS build tools instruct the compiler to process the file `mbed_config.h` as if it were the first include of any C or C++ source file, so you do not have to include `mbed_config.h` manually.
```
bmcdonnell-ionx: Via what mechanism? (I'm not sure if it's best to add that info here, but I'm curious.)

docs/tools/config_system.md
```diff
@@ -1,23 +1,135 @@
...
+
+Do not edit `mbed_config.h` manually. It may be overwritten the next time you compile or export your application, and all your changes will be lost.
+
+### Configuration parameters in `mbed_app.json`, `mbed_lib.json`
+
+An application may have one `mbed_app.json` in the root of the application and many `mbed_lib.json` files throughout the application. When present, `mbed_app.json` may override library and target configuration parameters and define new configuration parameters.
```
bmcdonnell-ionx: What are "target configuration parameters", and how/where are they specified?

docs/tools/config_system.md
```diff
@@ -1,23 +1,135 @@
...
 
 - The sampling period for a data acquisition application.
 - The default stack size for a newly created OS thread.
 - The receive buffer size of a serial communication library.
 - The flash and RAM memory size of an Mbed target.
 
-The configuration system gathers and interprets all the configurations defined in the source tree. The output of the configuration system is a list of macros that are automatically defined when compiling the code.
+The Arm Mbed OS configuration system gathers and interprets the configuration defined in the target, all `mbed_lib.json` files and the `mbed_app.json` file. The configuration system creates a single header file, `mbed_config.h`, that contains all of the defined configuration parameters converted into C preprocessor macros. `mbed compile` places `mbed_config.h` in the build directory and `mbed export` places it in the application root. The Arm Mbed OS configuration system is run during `mbed compile` before invoking the compiler and during `mbed export` before creating project files.
```
bmcdonnell-ionx: `the configuration defined in the target` - where is this? What are the filenames?

docs/tools/config_system.md
```diff
@@ -1,23 +1,135 @@
 ## The configuration system
 
-The Arm Mbed configuration system customizes the compile time configuration of various Mbed components (targets, libraries and applications). Each component can define a number of *configuration parameters*. The values of these configuration parameters can then be *overridden* in various ways. Configuration is defined using [JSON](http://www.json.org/). Some examples of configuration parameters:
+The Arm Mbed OS configuration system, a part of the Arm Mbed OS Build Tools, customizes compile time configuration parameters. Each library may define a number of configuration parameters in its `mbed_lib.json`. The values of these configuration parameters may be overridden by `mbed_app.json`. Configuration is defined using [JSON](http://www.json.org/). Some examples of configuration parameters:
```
bmcdonnell-ionx: Consider clarifying what you mean by "library" here, since you're talking about neither static nor dynamic libraries, nor `.lib` files understood by Mbed CLI. Maybe even just do it in a footnote.

docs/tools/config_system.md
```diff
@@ -1,23 +1,135 @@
...
+
+The order in which overrides are considered is:
+
+ 1. Libraries override target configuration with `mbed_lib.json`.
+ 2. The application overrides target and library configuration with `mbed_app.json`
 
 #### Defining configuration parameters
 
-The configuration system understands configuration data defined in targets, libraries and applications. Although there are some slight differences in the way the configuration system works in these cases, the configuration parameters are always defined in a JSON object called "config".
+The configuration system understands configuration parameters defined in targets, libraries and applications. The configuration parameters are always defined in a JSON object called "config".
```
bmcdonnell-ionx: "Target overrides" aren't considered "configuration parameters"?

docs/tools/config_system.md
```diff
@@ -34,28 +146,40 @@ For example:
...
-    - `help`: an optional help message that describes the purpose of the parameter.
-    - `value`: an optional field that defines the value of the parameter.
-    - `required`: an optional key that specifies whether the parameter **must** be given a value before compiling the code (`false` by default). It's not possible to compile a source tree with one or more required parameters that don't have a value. Generally, it makes sense to define a required parameter only when it doesn't have a `value` key.
-    - `macro_name`: an optional name for the macro defined at compile time for this configuration parameter. The configuration system will automatically figure out the corresponding macro name for a configuration parameter, but the user can override this automatically computed name by specifying `macro_name`.
+Above, the configuration parameters `param1` and `param2` are defined using a description object. The following keys are supported on the description object:
+  - `help`: an optional help message that describes the purpose of the parameter.
+  - `value`: an optional field that defines the value of the parameter.
```
bmcdonnell-ionx: But you said above that "A macro definition will not be generated for a parameter that doesn't have a value." If that's true, how can the value be optional? Or doesn't it make the parameter useless without a value?

docs/tools/config_system.md
bmcdonnell-ionx: Or did you mean that only for `"required"` parameters?

docs/tools/config_system.md
bmcdonnell-ionx: Related: https://github.com/ARMmbed/Handbook/pull/463/files#r176273626

docs/tools/config_system.md
```diff
@@ -88,90 +213,33 @@ In this JSON file:
...
 
-`target_overrides` is used to override the values of the parameters depending on the current compilation target. The keys in `target_overrides` are matched against toolchain *labels*. (You can find a description of Mbed targets in our documentation about [adding and configuring targets](/docs/development/tools/adding-and-configuring-targets.html).) If a key inside `target_overrides` matches one of the target labels, the parameter values are changed according to the value of the key. In the example above:
+All configuration parameters defined in `mylib` have a `mylib.` prefix. In a `mbed_app.json`, `buffer_size` is accessible using the name `mylib.buffer_size`.
+`target_overrides` is used to override the values of the parameters depending on the current compilation target. The keys in `target_overrides` are matched against target labels. (You can find a description of Mbed targets in our documentation about [adding and configuring targets](/docs/development/tools/adding-and-configuring-targets.html).) If a key inside `target_overrides` matches one of the target labels, the parameter values are changed according to the value of the key.
```
bmcdonnell-ionx: Link is broken when using "view" here. Does it work in the main repo or on the website?

docs/tools/config_system.md
```diff
@@ -88,90 +213,33 @@ In this JSON file:
...
+It is an error for `mbed_lib.json` to override configuration parameters that were not defined.
 
-Except `name`, all the above keys in the JSON file are optional. However, if `target_overrides` is defined, `config` must also be defined.
+#### Overriding target attributes
 
-As explained in the section about [defining configuration parameters](#defining-configuration-parameters), the parameters have an implicit `mylib.` prefix. Outside `mylib`, `buffer_size` is accessible using the name `mylib.buffer_size`. An application will be able to override the value of this parameter, as described in [this section](#configuration-data-in-applications).
+Target configurations contain a set of attributes that you may manipulate with configuration. You may override these attributes as if they were a normal configuration parameter. If these attributes are cumulative, you may also manipulate them with the special `attribute_add` and `attribute_remove` attributes. Find a list of the attributes that may be overwritten in our documentation about [adding and configuring targets](/docs/development/tools/adding-and-configuring-targets.html)
```
bmcdonnell-ionx: Link is broken when using "view" here. Does it work in the main repo or on the website?

bmcdonnell-ionx: > If these attributes are cumulative, you may also manipulate them with the special `attribute_add` and `attribute_remove` attributes.

I don't follow.

docs/tools/config_system.md
```diff
@@ -181,128 +249,20 @@ The last point above is important. The application can freely override the confi
...
+The application may override any configuration parameter by specifying the configuration parameters including their prefix (like `mylib.timer_period`). If an overridden parameter doesn't have a prefix, it overrides a parameter in its own `config` section.
 
-Other than this, `target_overrides` works exactly like it does for libraries. Keys in `target_overrides` are still processed in the order they are defined, so for the example above, the `*` override is always processed first (because it matches all targets) and then `Base` is only processed for the `Base` target.
+`myapp` above defines its own configuration parameter (`welcome_string`) and overrides the configuration in both the target (`target.mac_addr_high`) and its `mylib` dependency (`mylib.timer_period`):
```
bmcdonnell-ionx: Awkward to refer to a literal "`myapp`" that hasn't been named as such (above).

bmcdonnell-ionx:
@theotherjimmy Great work; thanks!

Comparatively minor stuff remains...

- I made a PR on your branch with some minor improvements. (It might include some rephrasings you don't prefer.)
- Please be sure to read all my preceding comments here on Github (not just in your email), since I edited several of them.
- I see that all your changes here are to the [Tools](https://os.mbed.com/docs/v5.7/tools/index.html) &rarr; [The Configuration system](https://os.mbed.com/docs/v5.7/tools/configuring-tools.html). What's the plan for [Reference](https://os.mbed.com/docs/v5.7/reference/index.html) &rarr; [Configuration overview](https://os.mbed.com/docs/v5.7/reference/configuration.html)?
- Is the content in the [Platform](https://os.mbed.com/docs/v5.7/reference/configuration-platform.html), [Drivers](https://os.mbed.com/docs/v5.7/reference/configuration-drivers.html), [RTOS](https://os.mbed.com/docs/v5.7/reference/configuration-rtos.html), and [Connectivity](https://os.mbed.com/docs/v5.7/reference/configuration-connectivity.html) pages generated by `mbed compile --config`? If so, and if those pages are staying, please update them to say so.

> Sections ordered by the order that a user is likely to need the information (pleas correct me if I got this wrong).

Good overall, IMO. Thoughts:
- Not necessarily reordering sections, but is it possible to have a statement closer to the beginning about the connection to the `#defines`? That would help connect what I do know (`#define`s) with what I don't yet (the Mbed OS config system, `.json` files, etc).
- Me, as an Mbed OS user - I didn't even realize this page was here. I repeatedly went to the one in the Reference section, and came away confused. (Thus bullet #3 above.)

bmcdonnell-ionx:
Also, depending on how in-depth you want to go, you could seek additional input from some users who don't yet know the config system - since @loverdeg-ep and I got to cheat by ramping-up with @geky's help via conversation in #462. 😉

docs/tools/config_system.md
```diff
@@ -1,23 +1,135 @@
...
+
+You may find that configuration naming does not always describe the parameter's purpose adequately. Include the `-v` switch to include the help text defined with the configuration parameter, where the value of the configuration parameter is defined, and other details. The command `mbed compile --config -v` in the same application as above prints:
+```
+Configuration parameters
+------------------------
+Name: cellular.random_max_start_delay
+    Description: Maximum random delay value used in start-up sequence in milliseconds
+    Defined by: library:cellular
+    No value set
+Name: cellular.use-apn-lookup
```
bmcdonnell-ionx: Tangent: I see that in cases where the macro name is not defined in the `.json` file, it is not in the verbose listing, but it still is in the brief listing. Is this desirable behavior? Consider adding "Macro name" for all, with "` (default)`" following those that aren't explicitly specified in the `.json` files.

docs/tools/config_system.md
theotherjimmy: Looking above, that note is out dated. I'll remove it.

docs/tools/config_system.md
theotherjimmy: That's `-include` (should be `-imacros`) for gcc, `-preinclude` for both ArmC5 and IAR EWARM. I don't think that's really relevant, as you mentioned. I also reserve the right to change that implementation detail.

docs/tools/config_system.md
theotherjimmy: I'll reword to "configuration parameters in libraries and targets". Does that avoid the confusion?

docs/tools/config_system.md
theotherjimmy: `mbed-os/target/targets.json`. I'll add it, and a link to it's documentation.

docs/tools/config_system.md
theotherjimmy: I'll add a Note: library in this context is anything with an `mbed_lib.json`.

It's a little circular XD

docs/tools/config_system.md
bmcdonnell-ionx: Sounds good.

docs/tools/config_system.md
theotherjimmy: "target_overrides" override "configuration parameters". All keys within a value in "target_overrides" must name a "configuration parameter", so while they are not defining config, they're part of the same system. Where could I make this more clear? alternatively, what gave you the impression that "target overrides" are not part of the config system?

docs/tools/config_system.md
theotherjimmy: Yeah, I removed that note. Does that help?

docs/tools/config_system.md
```diff
@@ -7,7 +7,9 @@ The Arm Mbed OS configuration system, a part of the Arm Mbed OS Build Tools, cus
...
+The Arm Mbed OS configuration system gathers and interprets the configuration defined in the target in its [target configuration](/docs/development/tools/adding-and-configuring-targets.html), all `mbed_lib.json` files and the `mbed_app.json` file. The configuration system creates a single header file, `mbed_config.h`, that contains all of the defined configuration parameters converted into C preprocessor macros. `mbed compile` places `mbed_config.h` in the build directory and `mbed export` places it in the application root. The Arm Mbed OS configuration system is run during `mbed compile` before invoking the compiler and during `mbed export` before creating project files.
```
bmcdonnell-ionx: "defined in the target in its target configuration" - should that just be "defined in its target configuration"?

Also, broken link when viewed from here. Do they work on the site?

docs/tools/config_system.md
theotherjimmy: @AnotherButler Correct me if I'm wrong, but I think this works on the website.

docs/tools/config_system.md
theotherjimmy: I'll reword.

docs/tools/config_system.md
theotherjimmy: Yep. Thanks!

bmcdonnell-ionx:
@theotherjimmy Consider merging my PR into your branch ASAP, to minimize merge conflicts from new changes you're making.

docs/tools/config_system.md
theotherjimmy: I'm wondering how you define a configuration parameter without giving it a name. Could you explain?

theotherjimmy:
Yeah! I just noticed that you had a PR up. I went through the comment history linearly. sorry about that.

docs/tools/config_system.md
```diff
@@ -222,7 +222,7 @@ It is an error for `mbed_lib.json` to override configuration parameters that wer
 
 #### Overriding target attributes
 
-Target configurations contain a set of attributes that you may manipulate with configuration. You may override these attributes as if they were a normal configuration parameter. If these attributes are cumulative, you may also manipulate them with the special `attribute_add` and `attribute_remove` attributes. Find a list of the attributes that may be overwritten in our documentation about [adding and configuring targets](/docs/development/tools/adding-and-configuring-targets.html)
+Target configurations contain a set of attributes that you may manipulate with configuration. You may override these attributes as if they were a normal configuration parameter. If these attributes are cumulative, they are a list of items that you may add to with the special attribute `attribute_add` and subtract from with `attribute_remove`. It is an error to both add and subtract the same value from a cumulative attribute. Find a list of the attributes that may be overwritten in our documentation about [adding and configuring targets](/docs/development/tools/adding-and-configuring-targets.html)
```
bmcdonnell-ionx: What does "If these attributes are cumulative" mean? Do you mean "Attributes may be cumulative, in which case they are a list of items..."?

What is the syntax for cumulative attributes? Where do they go? (I might've forgotten a bit of context that is clear on a read-through.)

Also, missing period at the end.

theotherjimmy: Yes. That phrasing is much better. I'll be more explicit that these attributes may be added to and subtracted from with the name of their configuration parameter + `_add` and `_subtract` respectively.

docs/tools/config_system.md
bmcdonnell-ionx: To be clear, I'm suggesting a change to the tool output in verbose mode, not just the documentation. (Then the documentation here would reflect the new output.)

Example: Since no `"macro_name"` is specified for `"cellular.random_max_start_delay"`, add a line in its verbose output, `Macro name: MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY  (default)`.

theotherjimmy:
Alrightly. Getting to you're process questions @bmcdonnell-ionx 

-----
>I see that all your changes here are to the Tools → The Configuration system. What's the plan for Reference → Configuration overview?

I'm going to rewrite that as appropriate next.

-------

Is the content in the Platform, Drivers, RTOS, and Connectivity pages generated by mbed compile --config? 

Yes.

If so, and if those pages are staying, please update them to say so.

Absolutely.

docs/tools/config_system.md
theotherjimmy: Odd. Check a macro with a value. It has the behavior you're requesting. I think the fact that it does not show for config params without a value is a bug.

bmcdonnell-ionx: That's what I was saying, though I don't know enough about the intent to say whether it's a "bug" per se.

docs/tools/config_system.md
bmcdonnell-ionx: Not sure what I was thinking here. Ignore, I guess.

theotherjimmy: If you can remember what caused that confusion, pleas don't hesitate to let me know. I would like to improve the text to avoid it.

docs/tools/config_system.md
theotherjimmy: It also prints in the short version. Definitely seems like a bug to me.

theotherjimmy:
@AnotherButler I just read the "reference section" version of the config documentation. It's not worth keeping. Should we replace it with this document that I have rewritten or kill it entirely?

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
-## Configuration overview
+## The configuration system
 
-The Arm Mbed configuration system customizes the compile time configuration of Mbed components: targets, libraries and applications.
+The Arm Mbed OS configuration system, a part of the Arm Mbed OS build tools, customizes compile time configuration parameters. Each library may define a number of configuration parameters in its `mbed_lib.json`. `mbed_app.json` may override the values of these configuration parameters. Configuration is defined using [JSON](http://www.json.org/). Some examples of configuration parameters:
```
AnotherButler: Query: Who or what defines configuration using JSON? You, the user? Please clarify for active voice.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
 
-Mbed OS provides some default memory configurations for thread memory allocation. You may want to modify these configurations to better suit your project. You can read about the default configurations and how you can modify them below.
+- The sampling period for a data acquisition application.
+- The default stack size for a newly created OS thread.
+- The receive buffer size of a serial communication library.
+- The flash and RAM memory size of an Mbed target.
 
-#### Configuring the main thread stack
+The Arm Mbed OS configuration system gathers and interprets the configuration defined in the target in its [target configuration](/docs/development/tools/adding-and-configuring-targets.html), all `mbed_lib.json` files and the `mbed_app.json` file. The configuration system creates a single header file, `mbed_config.h`, that contains all of the defined configuration parameters converted into C preprocessor macros. `mbed compile` places `mbed_config.h` in the build directory, and `mbed export` places it in the application root. The Arm Mbed OS configuration system is run during `mbed compile` before invoking the compiler and during `mbed export` before creating project files.
```
AnotherButler: Query: Who or what runs the Arm Mbed OS configuration system during `mbed compile` before invoking the compiler? Please clarify for active voice.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+#ifndef __MBED_CONFIG_DATA__
+#define __MBED_CONFIG_DATA__
+
+// Configuration parameters
+#define MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY         0 // set by library:cellular
+#define MBED_CONF_CELLULAR_USE_APN_LOOKUP                 1 // set by library:cellular
+<file truncated for brevity>
+```
+
+The name of the macro for a configuration parameter is either a prefixed name or explicitly specified by `macro_name`. A prefixed name is constructed from the prefix `MBED_CONF_`, followed by the name of the library or `APP`, followed by the name of the parameter. The prefixed name is then capitalized and converted to a valid C macro name. For example, the `random_max_start_delay` configuration parameter in the library `cellular` is converted to `MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY`.
```
AnotherButler: Query: Who or what constructs the prefixed name and then capitalizes it and converts it to a valid C macro name? You, the user? The configuration system? The build tools? Please clarify for active voice.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+
+### Configuration parameters in `mbed_app.json`, `mbed_lib.json`
+
+An application may have one `mbed_app.json` in the root of the application and many `mbed_lib.json` files throughout the application. When present, `mbed_app.json` may override configuration parameters defined in libraries and the target and define new configuration parameters.
+
+#### Overriding configuration parameters
+
+The configuration system allows a user to override any defined configuration parameter with a JSON object named `"target_overrides"`.
+
+The keys in the `"target_overrides"` section are the names of a target that the overrides apply to, or the special wildcard `*` that applies to all targets. The values within the `"target_overrides"` section are an object mapping configuration parameters, as printed by `mbed compile --conifg`, to new values. See the example `"target_overrides"` section below.
```
AnotherButler: Query: Is the second sentence in this paragraph saying the values in `target_overrides` are object-mapping configuration parameters, or is it saying they are an object that maps configuration parameters? Please clarify.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+        },
+        "param3": 10
     }
 }
 ```
 
-#### Configuring user spawned thread stacks
+You define a configuration parameter by specifying its name as the key and specifying its value either with a description object or by value. The JSON fragment above defines three configuration parameters named `param1`, `param2` and `param3`.
+
+Above, the configuration parameters `param1` and `param2` are defined using a description object. The following keys are supported on the description object:
```
AnotherButler: Query: Who or what defines the configuration parameters? Also, would it be correct to replace the second sentence with "The description object supports the following keys:", or would that change the meaning?

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+
+  - `help`: an optional help message that describes the purpose of the parameter.
+  - `value`: an optional field that defines the value of the parameter.
+  - `required`: an optional key that specifies whether the parameter must be given a value before compiling the code (`false` by default). It's not possible to compile a source tree with one or more required parameters that don't have a value. Generally, setting `required` to true is only useful when `value` is not set.
```
AnotherButler: Query: Who or what gives the parameter a value before compiling the code? You, the user? Please clarify for active voice.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+  - `macro_name`: an optional name for the macro defined at compile time for this configuration parameter. The configuration system automatically figures out the corresponding macro name for a configuration parameter, but the user can override this automatically computed name by specifying `macro_name`.
+
+You define a macro by value by using an integer or string instead of the description object, such as `param3` above. Defining a parameter by value is equivalent to a configuration parameter defined with a description object with the key `value` set to the value in place of the description object, the key `help` unset, the key `macro_name` unset, and the key `required` set to `false`.
```
AnotherButler: Query: Could we replace the second sentence with "Defining a parameter by value is equivalent to defining it with a description object with the key `value` set to the value in place of the description object, the key `help` not set, the key `macro_name` not set and the key `required` set to `false`."? Or would that change the meaning?

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
     }
-    ```
+}
+```
+
+In this JSON file:
+
+- `name` is the name of the library. **This is a required field.**
+- `config` defines the configuration parameters of the library, as the section about [defining configuration parameters](#defining-configuration-parameters) explains.
+- `macros` is a list of extra macros that are defined when compiling a application that includes this library. A macro can be defined without a value (such as `MYMOD_MACRO1` above) or with a value (such as `MYMOD_MACRO2` above).
```
AnotherButler: Query: Who or what defines macros when compiling an application that includes this library? You, the user? Also, who or what can define the macro without a value? Please clarify for active voice.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+- `target_overrides` is a dictionary with target-specific values for the configuration parameters.
 
-2. Construct the thread like this in your application:
+All configuration parameters defined in `mylib` have a `mylib.` prefix. In `mbed_app.json`, `buffer_size` is accessible using the name `mylib.buffer_size`.
 
-    `Thread thread(osPriorityNormal, 2000);`
+Use `target_overrides` to override the values of the parameters, depending on the current compilation target. The keys in `target_overrides` are matched against target labels. (You can find a description of Mbed targets in our documentation about [adding and configuring targets](/docs/development/tools/adding-and-configuring-targets.html).) If a key inside `target_overrides` matches one of the target labels, the parameter values change according to the value of the key.
```
AnotherButler: Query: Who matches the keys in `target_overrides` against target labels? Please clarify for active voice.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
 ```
-int main() {
-    uint32_t stack_on_main_stack[128 / sizeof(uint32_t)]; // 128 bytes
-    Thread thread(osPriorityNormal, sizeof(stack_on_main_stack), stack_on_main_stack);
+
+It is an error to both add and subtract the same value from a cumulative attribute. For a list of the attributes that you may overwrite, please see our documentation about [adding and configuring targets](/docs/development/tools/adding-and-configuring-targets.html)
+
+### `mbed_app.json` Specification
+
+`mbed_app.json` may be present at the root of your application or specified as the argument to the `--app-config` parameter. Unlike library configuration, only one `mbed_app.json` is interpreted during `mbed compile` or `mbed export`. Like `mbed_lib.json`, `mbed_app.json` is a JSON formatted document that contains a root JSON Object. The keys within this object are sections. The allowed sections and their meanings are below:
```
AnotherButler: Query: Who or what interprets `mbed_app.json` during `mbed compile` or `mbed export`? Please clarify for active voice.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
 ```
+
+The application may override any configuration parameter by specifying the configuration parameters including their prefix (such as `mylib.timer_period`). If an overridden parameter doesn't have a prefix, it overrides a parameter in its own `config` section.
+
+The `mbed_app.json` above defines its own configuration parameter (`welcome_string`) and overrides the configuration in both the target (`target.mac_addr_high`) and its `mylib` dependency (`mylib.timer_period`):
+
+- When compiling for `NCS36510`, `app.welcome_string` is `"Hello!"`, `target.mac_addr_high` is `"0x11223344"` (from the `NCS36510` override) and `mylib.timer_period` is 100 (from the `*` override).
+- When compiling for `LPC1768`, `app.welcome_string` is `"Hello!"` and `mylib.timer_period` is 100 (also from the `*` override).
+
+It is an error for the application configuration to override configuration parameters that were not defined.
```
AnotherButler: Query: that were not defined by whom or by what? You, the user? Please clarify for active voice.

docs/reference/configuration/configuration.md
```diff
@@ -125,7 +126,7 @@ The order in which overrides are considered is:
 
 #### Defining configuration parameters
 
-The configuration system understands configuration parameters defined in targets, libraries and applications. The configuration parameters are always defined in a JSON object called "config".
+The configuration system understands configuration parameters that targets, libraries and applications define. A JSON object called "config" always defines the configuration parameters.
```
bmcdonnell-ionx: Overall I appreciate the preference for active voice, but I find it distracting in these 2 sentences. (Just my opinion. I defer to your judgement as to whether to change it.)

docs/reference/configuration/configuration.md
theotherjimmy: The sentence above outlines where confutation comes from: libraries, targets, applications.

docs/reference/configuration/configuration.md
theotherjimmy: That's a very odd question. Also I think the answer will be irrelevant to the user. It's also circular: `mbed compile` runs the configuration system.

theotherjimmy: You can word it similar to the prior sentence:

`mbed compile` run s the config sysetm ... and `mbed export` runs the config system...

docs/reference/configuration/configuration.md
theotherjimmy: That's intentionally ambiguous: anyone can construct a prefixed name with the method outlined in those sentences.You're welcome to use "the configuration system" here.

docs/reference/configuration/configuration.md
theotherjimmy: What are object-mapping configuration parameters?

----

probably objects that map configuration parameters then. :D

docs/reference/configuration/configuration.md
theotherjimmy: > Who or what defines the configuration parameters?

They're in the example

>  Also, would it be correct to replace the second sentence with "The description object supports the following keys:", or would that change the meaning?

That's the same thing. Replace away.

docs/reference/configuration/configuration.md
theotherjimmy: how about must have a value?

docs/reference/configuration/configuration.md
theotherjimmy: That's the same to me.

docs/reference/configuration/configuration.md
theotherjimmy: This is where active voice is really difficult: the presence of `macros` defines the macros. 

----

Macros defined by this feild may omit a value, or may include a value.

docs/reference/configuration/configuration.md
theotherjimmy: The config system?

docs/reference/configuration/configuration.md
theotherjimmy: The config system.

docs/reference/configuration/configuration.md
theotherjimmy: If they were not defined, there can't be a subject. (almost by definition)

theotherjimmy: You can change that to are, "that are not defined" means the same thing.

docs/reference/configuration/configuration.md
theotherjimmy: You can use the same structure as `config` above: `macros` defines the provided list of C preprocessor macros.

docs/reference/configuration/configuration.md
theotherjimmy: That should also be singular:

It is an error for the application configuration to override a configuration parameter that is not defined.

docs/reference/configuration/Connectivity.md
```diff
@@ -1,7 +1,7 @@
 <h2 id="configuration-connectivity">Connectivity</h2>
 
 
-You can configure various parameters using either C++ APIs or the Arm Mbed configuration system. Below are the parameters you can configure using the Mbed configuration system:
+You can configure various parameters using either C++ APIs or the Arm Mbed configuration system. The following is the complete list of Connectivity configuration parameters, as generated by `mbed compile --config -v`. Please see [the Configuration system documentation](/docs/development/reference/configuration.html) for details on how you may use or override these settings.
```
bridadan: Nit: any reason why `Connectivity` and `Configuration system` are capitalized?

docs/reference/configuration/Drivers.md
```diff
@@ -1,5 +1,7 @@
 <h2 id="configuration-drivers">Drivers</h2>
 
+The following is the complete list of driver configuration parameters, as generated by `mbed compile --config -v`. Please see [the Configuration system documentation](/docs/development/reference/configuration.html) for details on how you may use or override these settings.
```
bridadan: Nit: any reason why `Configuration system` is capitalized?

docs/reference/configuration/Platform.md
```diff
@@ -1,5 +1,7 @@
 <h2 id="configuration-platform">Platform</h2>
 
+The following is the complete list of platform configuration parameters, as generated by `mbed compile --config -v`. Please see [the Configuration system documentation](/docs/development/reference/configuration.html) for details on how you may use or override these settings.
```
bridadan: Nit: any reason why `Configuration system` is capitalized?

docs/reference/configuration/RTOS.md
```diff
@@ -8,6 +8,9 @@ All threads in Mbed OS share a global heap. By default, mbed OS dynamically allo
 
 By default, there are four threads running after boot: the ISR/scheduler thread, the idle thread, the timer thread and the main application thread. Combined, the idle thread, timer thread and ISR/scheduler thread consume 2 kilobytes of RAM. The user cannot modify these. You can expand or reduce the size of the main application thread stack by using the Mbed configuration system.
 
+
+The following is the complete list of RTOS configuration parameters, as generated by `mbed compile --config -v`. Please see [the Configuration system documentation](/docs/development/reference/configuration.html) for details on how you may use or override these settings.
```
bridadan: Nit: any reason why `Configuration system` is capitalized?

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+  - `macro_name`: an optional name for the macro defined at compile time for this configuration parameter. The configuration system automatically figures out the corresponding macro name for a configuration parameter, but the user can override this automatically computed name by specifying `macro_name`.
+
+You define a macro by value by using an integer or string instead of the description object, such as `param3` above. Defining a parameter by value is equivalent to defining it with a description object with the key `value` set to the value in place of the description object, the key `help` not set, the key `macro_name` not set and the key `required` set to `false`. Defining a parameter by value is equivalent to a configuration parameter defined with a description object with the key `value` set to the value in place of the description object, the key `help` unset, the key `macro_name` unset, and the key `required` set to `false`.
```
bridadan: This paragraph contains the same sentence essentially duplicated.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
     }
-    ```
+}
+```
+
+In this JSON file:
+
+- `name` is the name of the library. **This is a required field.**
+- `config` defines the configuration parameters of the library, as the section about [defining configuration parameters](#defining-configuration-parameters) explains.
+- `macros` is a list of extra macros that are defined when compiling a application that includes this library. `macros` defines the provided list of C preprocessor macros.
```
bridadan: This bullet point is funny, seems like its repeating the same thing.

Also "a application" should be "**an** application".

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
 
-2. Construct the thread like this in your application:
+All configuration parameters defined in `mylib` have a `mylib.` prefix. In `mbed_app.json`, `buffer_size` is accessible using the name `mylib.buffer_size`.
 
-    `Thread thread(osPriorityNormal, 2000);`
+Use `target_overrides` to override the values of the parameters, depending on the current compilation target. The configuration system matches keys in `target_overrides` against target labels. (You can find a description of Mbed targets in our documentation about [adding and configuring targets](/docs/development/tools/adding-and-configuring-targets.html).) If a key inside `target_overrides` matches one of the target labels, the parameter values change according to the value of the key. 
 
-By default, mbed OS dynamically allocates the memory for the thread's stack from the global heap, though you can change it in the [Thread constructor](/docs/development/reference/thread.html). Here is an example application that uses memory allocated from the main thread's stack for the newly spawned thread stack:
+It is an error for `mbed_lib.json` to override configuration parameters that were not defined.
```
bridadan: Tense looks a bit funny at the end? Thinking it should be:

"It is an error for `mbed_lib.json` to override configuration parameters that **are** not defined."

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+Target configurations contain a set of attributes that you may manipulate with configuration. You may override these attributes as if they were a normal configuration parameter. Attributes may be cumulative, in which case they are a list of items. You may add to a cumulative attribute by overriding a configuration parameter with the name of the cumulative attribute suffixed with `_add` and remove from a cumulative attribute with the suffix `_remove`. When you override, add to or subtract from a cumulative attribute, the value must be a list of items to replace the definition with, add or remove. For example, add the value `IPV4` to a target's features list with the syntax:
+
+```JSON
+"target.features_add": ["IPV4"]
 ```
-int main() {
-    uint32_t stack_on_main_stack[128 / sizeof(uint32_t)]; // 128 bytes
-    Thread thread(osPriorityNormal, sizeof(stack_on_main_stack), stack_on_main_stack);
+
+It is an error to both add and subtract the same value from a cumulative attribute. For a list of the attributes that you may overwrite, please see our documentation about [adding and configuring targets](/docs/development/tools/adding-and-configuring-targets.html)
```
bridadan: Missing a period (`.`) at the end of the last sentence.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+
+### `mbed_app.json` Specification
+
+`mbed_app.json` may be present at the root of your application or specified as the argument to the `--app-config` parameter. The configuration system interprets only one `mbed_app.json` during `mbed compile` or `mbed export`, unlike library configuration. Like `mbed_lib.json`, `mbed_app.json` is a JSON formatted document that contains a root JSON Object. The keys within this object are sections. The allowed sections and their meanings are below:
```
bridadan: I think it'd be worth mentioning `mbed compile` and `mbed export` with regards to the `--app-config` parameter.

Might take a little rewording though, something like: 

> `mbed_app.json` may be present at the root of your application or specified as the argument to the `--app-config` parameter when using `mbed compile` and `mbed export`. The configuration system interprets only one `mbed_app.json`, unlike library configuration.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+| -------- | ------ |
+| Target | `target.` |
+| Any library | The name of the library, as found in the `name` section of `mbed_lib.json`, followed by a dot (.) |
+| Application | `app.` |
 
-The default stack size of a user spawned thread is 4 kilobytes. You can configure the stack size of user spawned threads by using the [Mbed config system](/docs/development/tools/index.html#the-configuration-system) or passing in the [Thread constructor](/docs/development/reference/thread.html). You can reduce the default stack size for user spawned threads from 4K to 2K in the following two ways:
+### `mbed_lib.json` format specification
 
-1. Create an `mbed_app.json` file for your project with the following content:
+`mbed_lib.json` is a JSON formatted document that contains a root JSON Object. The keys within this object are sections. See the allowed sections and their meanings below:
```
bridadan: Not required simplification: Can we not call the keys in the object "sections"? Seems like an unnecessary new "concept" to introduce. Could the above:

> The keys within this object are sections. See the allowed sections and their meanings below:

be reworded to:

> The keys within this object and their meanings are found below:

And then change the "Section" header of the table below to "Key".

I notice you use "section" throughout this doc, but I think those uses are fine as it is informally referring to the "section" of the JSON document. I might be overanalyzing this a bit, but I thought I had missed a key config concept when I read the whole "section" bit.

docs/reference/configuration/configuration.md
bridadan: Also, same deal about "sections" here.

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
-    ```JSON
-    {
-        "config": {
-            "thread-stack-size": {
-                "value": 2000
-            }
+| Section | Required | Meaning |
+| ------- | -------- | ------- |
+| `name`  | Yes      | Name of the library. Must be unique. May not be `app` or `target`.|
+| `macros` | No | List of macros to define on the command-line. |
```
bridadan: Are they actually set on the command-line and not `mbed_config.h`? If so, are they are any important considerations other than just to look for them in a different place when building? (if no, might be helpful to just mention that distinction so understand its just a simple relocation vs behavior change)

docs/reference/configuration/configuration.md
bmcdonnell-ionx: I'm on the fence about this, but maybe it would be good to maintain the concept. I understand what a "section" is in general, and now I'm being told that the topmost JSON keys are used to divide the root JSON object into sections.

bmcdonnell-ionx: Plus then you have a word to distinguish between your topmost keys, and those lower in the hierarchy. For better or worse.

docs/reference/configuration/configuration.md
bmcdonnell-ionx: If you decide not to keep the "section" idea, be sure to change it in this table column title, too.

Though its use here makes me question if I understood correctly above that a "section" necessarily refers to the topmost keys in the JSON file.

docs/reference/configuration/configuration.md
theotherjimmy: That's probably a merge/rebase mistake.

docs/reference/configuration/configuration.md
theotherjimmy: Sorry I missed this. They're in `mbed_config.h`

theotherjimmy: @bmcdonnell-ionx Section is used as the topmost keys in the JSON file here too. Is there something that made you question that notion?

bmcdonnell-ionx: I was thinking topmost within the whole app hierarchy, i.e. within `mbed_app.json`. I guess you mean topmost within any JSON file, then?

theotherjimmy: You're guess is correct. How could I make that more clear?

docs/reference/configuration/configuration.md
bridadan: A fair point! Definitely required, mostly just my knee-jerk reaction.

docs/reference/configuration/configuration.md
bmcdonnell-ionx: I'm not sure that it needs clarification. I hadn't looked at it in context for a while.

theotherjimmy:
@AnotherButler You may have to do another round of editing. I made some changes.

docs/reference/configuration/configuration.md
```diff
@@ -232,16 +232,16 @@ Target configurations contain a set of attributes that you may manipulate with c
...
 
 ### `mbed_app.json` Specification
 
-`mbed_app.json` may be present at the root of your application or specified as the argument to the `--app-config` parameter. The configuration system interprets only one `mbed_app.json` during `mbed compile` or `mbed export`, unlike library configuration. Like `mbed_lib.json`, `mbed_app.json` is a JSON formatted document that contains a root JSON Object. The keys within this object are sections. The allowed sections and their meanings are below:
+`mbed_app.json` may be present at the root of your application or specified as the argument of the `--app-config` parameter to `mbed compile` and `mbed e}xort`. The configuration system interprets only one `mbed_app.json` during `mbed compile` or `mbed export`, unlike library configuration. Like `mbed_lib.json`, `mbed_app.json` is a JSON formatted document that contains a root JSON Object. The keys within this object are sections. The allowed sections and their meanings are below:
```
bmcdonnell-ionx: Typo: `mbed e}xort`

docs/reference/configuration/configuration.md
```diff
@@ -1,48 +1,276 @@
...
+events.present = 1 (macro name: "MBED_CONF_EVENTS_PRESENT")
+events.shared-dispatch-from-application = 0 (macro name: "MBED_CONF_EVENTS_SHARED_DISPATCH_FROM_APPLICATION")
+events.shared-eventsize = 256 (macro name: "MBED_CONF_EVENTS_SHARED_EVENTSIZE")
+events.shared-highprio-eventsize = 256 (macro name: "MBED_CONF_EVENTS_SHARED_HIGHPRIO_EVENTSIZE")
+events.shared-highprio-stacksize = 1024 (macro name: "MBED_CONF_EVENTS_SHARED_HIGHPRIO_STACKSIZE")
+events.shared-stacksize = 1024 (macro name: "MBED_CONF_EVENTS_SHARED_STACKSIZE")
+<output truncated for brevity>
+```
+
+You may find that configuration naming does not always describe the parameter's purpose adequately. Use the `-v` switch to include the help text defined with the configuration parameter, where the value of the configuration parameter is defined, and other details. The command `mbed compile --config -v` in the same application as above prints:
```
AnotherButler: Query: Do we create the configuration naming? If so, it should always describe the parameter's purpose adequately. Is it OK to delete the first sentence and start with "Use the..."?

bridadan: > Do we create the configuration naming? If so, it should always describe the parameter's purpose adequately. 

Not always, and ideally yes, but the behavior can sometimes be hard to fit in a short string, so the help text is used to explain more thoroughly.

> Is it OK to delete the first sentence and start with "Use the..."?

Honestly, yeah that's a much better solution 😄 Let's do that!

PR #471, opened 2018-03-26
Added logging page

deepikabhavnani:
Add description of mbed logging module.

Dependent on PR: https://github.com/ARMmbed/mbed-os/pull/5965

@SenRamakri @AnotherButler  - Please review.

### DIFF
--- a/docs/reference/api/platform/Logger.md
+++ b/docs/reference/api/platform/Logger.md
@@ -154,9 +154,6 @@ Usage:
 #define tr_cmdline(...)     MBED_LOG(TRACE_LEVEL_CMD, TRACE_GROUP, ##__VA_ARGS__)
 ```
 
-### Log Helper functions
-`mbed_log_ipv6` , `mbed_log_ipv6_prefix`, `mbed_log_array` are helper functions for logging special type of data. Helpers functions are not ISR friendly and should not be used in ISR context.
-
 Note: 
 1. Old Mbed OS logging APIs `mbed_trace`, `error`, `debug`, `debug_if`, are deprecated.
 2. All API's are available in the debug and develop build profiles but not in the release build profile.
--- a/docs/reference/api/platform/Logger.md
+++ b/docs/reference/api/platform/Logger.md
@@ -5,8 +5,6 @@ API's in logging module are printf-style API's which take module name and format
 
 By default all messages till LOG_LEVEL_DEBUG are enabled. If you want to enable Trace and Info level messages, then set `MBED_CONF_MAX_LOG_LEVEL` accordingly.
 
-Log messages are buffered in circular buffer and you can configure buffer size by setting `MBED_CONF_LOG_MAX_BUFFER_SIZE` config parameter. Default is 1024 bytes.
-
 All logging API's are ISR safe.
 
 ### Debug Log Levels
@@ -14,7 +12,7 @@ Below are various log levels supported and recommended usage.
 
 #### LOG_LEVEL_ERR_CRITICAL
 In case of critical errors when system recovery is not possible, you should use `MBED_CRIT` API. `MBED_CRIT` is equivalent to `ASSERT`.
-`MBED_CRIT` will log all the previous messages captured in log buffer and then application will terminate.
+`MBED_CRIT` will log message and then application will terminate.
 
 Usage:
 ```C
--- a/docs/reference/api/platform/Logger.md
+++ b/docs/reference/api/platform/Logger.md
@@ -3,9 +3,20 @@
 Mbed OS provides a set of API's that you can use to output different log level messages to STDIO at runtime. mbed_logger.h declares these functions, which are available only in debug builds.
 API's in logging module are printf-style API's which take module name and format string followed by arguments.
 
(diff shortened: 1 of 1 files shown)


docs/reference/api/platform/Logger.md
```diff
@@ -0,0 +1,164 @@
...
+Mbed OS provides a set of API's that you can use to output different log level messages to STDIO at runtime. mbed_logger.h declares these functions, which are available only in debug builds.
+API's in logging module are printf-style API's which take module name and format string followed by arguments.
+
+By default all messages till LOG_LEVEL_DEBUG are enabled. If you want to enable Trace and Info level messages, then set `MBED_CONF_MAX_LOG_LEVEL` accordingly.
+
+### Debug Log Levels
+Below are various log levels supported and recommended usage.
+
+#### LOG_LEVEL_ERR_CRITICAL
+In case of critical errors when system recovery is not possible, you should use `MBED_CRIT` API. `MBED_CRIT` is equivalent to `ASSERT`, it will dump previous debug log and application will terminate.
```
SenRamakri: "it will dump previous debug log and application will terminate" - can you please clarify on this? what is previous debug log here? Is that the messages captured in log buffer?

docs/reference/api/platform/Logger.md
```diff
@@ -0,0 +1,164 @@
...
+```C
+void dev_init() {
+    ...
+    if (DEV_ID != id) {
+        MBED_WARN ("Driver", "Incorrect Device ID");
+    }
+    ...
+}
+```
+
```
SenRamakri: Can you please add more examples which shows more arguments?

docs/reference/api/platform/Logger.md
```diff
@@ -0,0 +1,164 @@
+## Logger
+
+Mbed OS provides a set of API's that you can use to output different log level messages to STDIO at runtime. mbed_logger.h declares these functions, which are available only in debug builds.
+API's in logging module are printf-style API's which take module name and format string followed by arguments.
+
+By default all messages till LOG_LEVEL_DEBUG are enabled. If you want to enable Trace and Info level messages, then set `MBED_CONF_MAX_LOG_LEVEL` accordingly.
```
SenRamakri: Is LOG_LEVEL_DEBUG a global setting? Where is it configured?

docs/reference/api/platform/Logger.md
```diff
@@ -0,0 +1,164 @@
...
+    ...
+    MBED_TRACE("Driver", "Init successful");
+    return SUCCESS;
+}
+```
+You can alse use `MBED_TRACE_IF` function for trace level logs. It is similar to `MBED_TRACE` except that it takes an additional argument for condition, the message is logged only if condition evaluates to true.
+
+Usage:
+```C
+#define ENABLE_DRIVER_MSG   0
```
SenRamakri: Is it possible to enable all levels of messages from a specific driver( may be a  single .c file ) but disable everything else?

docs/reference/api/platform/Logger.md
deepikabhavnani: https://os.mbed.com/teams/mbed_example/code/logging/file/10b2ce72c7ac/main.cpp/
We can add more samples in this example file

docs/reference/api/platform/Logger.md
deepikabhavnani: MBED_CONF_MAX_LOG_LEVEL - MBED_CONF are all config setting globals.

docs/reference/api/platform/Logger.md
deepikabhavnani: Please read the note section below, if all drivers implement it right way then yes.

deepikabhavnani: You cannot do that for error / critical and warn level though

docs/reference/api/platform/Logger.md
deepikabhavnani: Yes the messages which are already captured in log buffer, since we push them to circular buffer. We need to dump them before performing mbed_die() operation. 
You want me to clarify more in document? In that case i will have to explain entire design/implementation of logging module

SenRamakri: You can say exactly as you mentioned - "all the messages which are already captured in log buffer".

docs/reference/api/platform/Logger.md
SenRamakri: Ok, Thanks for clarifying.

docs/reference/api/platform/Logger.md
```diff
@@ -0,0 +1,164 @@
+## Logger
```
SenRamakri: Can you please add information on how big the buffer size is and how it can be configured?

docs/reference/api/platform/Logger.md
```diff
@@ -154,9 +154,6 @@ Usage:
 #define tr_cmdline(...)     MBED_LOG(TRACE_LEVEL_CMD, TRACE_GROUP, ##__VA_ARGS__)
 ```
 
-### Log Helper functions
```
SenRamakri: Are we removing these helper functions?

deepikabhavnani: We are not removing helper functions, but helper functions are not API's. They will stay as internal functions and are subject to change.

PR #475, opened 2018-03-28
workflow: add PR types explanation

### LINKED ISSUE #429: Workflow: describe pull request type
Based on latest PR after we introduced PR type, I noticed that sometimes it is not clear what type a pull request belongs to. This might be due to not specifying each of these types.

Questions that I would like to answer and point to when not clear in contributions:

What is a fix or feature or refactor? When is it a breaking change?

We will send a proposal for clarification

0xc0170:
Every PR type should be clear to a user.
It is a user responsibility to specify PR type and accept the versioning based on it.

Fixes #429

Proposal to define PR types. I would like to reference this in every pull request so all of us are aware of the responsibility to select the proper type and the release label selection based on that.

Needs more work to be more clear and specify more details, expecting lot of comments 🔥 

@sg- @cmonr @adbridge @theotherjimmy @kjbracey-arm Please review

### DIFF
--- a/docs/reference/contributing/guidelines/workflow.md
+++ b/docs/reference/contributing/guidelines/workflow.md
@@ -94,7 +94,7 @@ We consider the following pull request types.
 
 ##### Fix
 
-A bug fix should be backward-compatible internal change that fixes incorrect behavior. The fix should provide a description what and how it fixes the issue. The test should be included to catch this issue in the future and confirm the correct behavior.
+A bug fix is backward-compatible internal change that fixes incorrect behavior. The fix provides a description what and how it fixes the issue. Tests are included to catch the issue in the future and confirms the correct behavior.
 
 Release: patch
 
@@ -115,7 +115,7 @@ Release: patch
 
 New features targets feature releases. It can be integrated only if the feature supports most of the targets (if it requires new target HAL implementation).
 
-Adding a new functionality is considered to be a feature. It does not matter if it is C++ or C.
+Adding a new functionality is considered to be a feature. It does not matter if it is C++, C or Python.
 
 Release: feature
 
--- a/docs/reference/contributing/guidelines/workflow.md
+++ b/docs/reference/contributing/guidelines/workflow.md
@@ -61,13 +61,13 @@ Pull requests on GitHub have to meet the following requirements to keep the code
 - Avoid merging commmits. (Always rebase when possible.)
 - Pull requests should fix a bug, add a feature or refactor.
 
-#### Mbed OS release versioning
+#### Release versioning
 
-Version number MAJOR.FEATURE.PATCH where:
+The Mbed OS version number follows the format MAJOR.FEATURE.PATCH. We use the:
 
-- MAJOR version for incompatible API changes
-- FEATURE version for adding functionality in backward-compatible manner
-- PATCH version for backward-compatible bug fixes
+- MAJOR version for incompatible API changes.
+- FEATURE version for adding functionality in backward-compatible manner.
+- PATCH version for backward-compatible bug fixes.
 
 ### Pull request categories
 
@@ -94,33 +94,33 @@ We consider the following pull request types.
 
 ##### Fix
 
-A bug fix is backward-compatible internal change that fixes incorrect behavior.
+A bug fix is a backward-compatible internal change that fixes incorrect behavior.
 
 Release: patch
 
 ##### Refactor
 
-Refactors are intended for feature releases if they are not fixing specific issue as they can introduce new issues.
+Refactors are intended for feature releases if they are not fixing specific issues because they can introduce new issues.
 
 Release: feature
 
 ##### New target
 
-Adding a new target targets the patch release as it updating targets folder implementation.
+Adding a new target is a change for a patch release because it updates the targets folder implementation.
 
 Release: patch
 
 ##### Feature
 
-New features targets feature releases. It can be integrated only if the feature supports most of the targets (if it requires new target HAL implementation).
+New features target feature releases. A new feature can be integrated only if the feature supports most of the targets (if it requires new target HAL implementation).
 
-Adding a new functionality is considered to be a feature. It does not matter if it is C++, C or Python.
+We consider adding a new functionality to be a feature. It does not matter if it is C++, C or Python.
 
 Release: feature
 
 ##### Breaking change
 
-Any change that results in breaking user space. It should have strong justification to be considered. Often the change could be made backward compatible, for example deprecate the old functionality and introduce the new replacement.
+A breaking change is any change that results in breaking user space. It should have strong justification for us to consider it. Often, such changes can be backward compatible, for example, deprecating the old functionality and introducing the new replacement.
 
 Release: major
 
--- a/docs/reference/contributing/guidelines/workflow.md
+++ b/docs/reference/contributing/guidelines/workflow.md
@@ -63,7 +63,7 @@ Pull requests on GitHub have to meet the following requirements to keep the code
 
 #### Release versioning
 
-Mbed OS versioning can be found at [How We Release Arm Mbed OS](https://os-doc-builder.test.mbed.com/docs/development/introduction/how-we-release-arm-mbed-os.html).
+You can find Mbed OS versioning at [How We Release Arm Mbed OS](/docs/development/introduction/how-we-release-arm-mbed-os.html).
 
 ### Pull request categories
 
--- a/docs/reference/contributing/guidelines/workflow.md
+++ b/docs/reference/contributing/guidelines/workflow.md
@@ -63,11 +63,7 @@ Pull requests on GitHub have to meet the following requirements to keep the code
 
 #### Release versioning
 
-The Mbed OS version number follows the format MAJOR.FEATURE.PATCH. We use the:
-
-- MAJOR version for incompatible API changes.
-- FEATURE version for adding functionality in backward-compatible manner.
-- PATCH version for backward-compatible bug fixes.
+Mbed OS versioning can be found at [How We Release Arm Mbed OS](https://os-doc-builder.test.mbed.com/docs/development/introduction/how-we-release-arm-mbed-os.html).
 
 ### Pull request categories
 
--- a/docs/reference/contributing/guidelines/workflow.md
+++ b/docs/reference/contributing/guidelines/workflow.md
@@ -94,14 +94,13 @@ We consider the following pull request types.
 
 ##### Fix
 
-A bug fix is backward-compatible internal change that fixes incorrect behavior. The fix provides a description what and how it fixes the issue. Tests are included to catch the issue in the future and confirms the correct behavior.
+A bug fix is backward-compatible internal change that fixes incorrect behavior.
 
 Release: patch
 
 ##### Refactor
 
-Refactors are intended for feature releases if they are not fixing specific issue as they 
-can introduce new issues.
+Refactors are intended for feature releases if they are not fixing specific issue as they can introduce new issues.
 
 Release: feature
 
--- a/docs/reference/contributing/guidelines/workflow.md
+++ b/docs/reference/contributing/guidelines/workflow.md
@@ -122,7 +122,7 @@ Release: feature
 
 Any change that results in breaking user space. It should have strong justification to be considered. Often the change could be made backward compatible, for example deprecate the old functionality and introduce the new replacement.
 
-Release: feature
+Release: major
 
 ##### Pull request template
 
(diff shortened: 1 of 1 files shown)


docs/reference/contributing/guidelines/workflow.md
```diff
@@ -79,6 +79,43 @@ Each feature has a tech lead. This person is responsible for:
...
+
+Adding a new functionality is considered to be a feature. It does not matter if it is C++ or C.
+
+Release: minor version
+
+##### Breaking change
+
+Any change that results in breaking user space. It should have strong justification to be considered. Often the change could be made backward compatible, for example deprecate the old functionality and introduce the new replacement.
+
+Release: minor version
```
geky: Shouldn't this be major version?

adbridge: Nope, major version is the 5. , next one will be 6. in a year's time

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -79,6 +79,43 @@ Each feature has a tech lead. This person is responsible for:
...
 
+##### Pull request types
+
+We consider the following pull request types.
+
+##### Fix
+
+A bug fix should be backward-compatible internal change that fixes incorrect behavior. The fix should provide a description what and how it fixes the issue. The test should be included to catch this issue in the future and confirm the correct behavior.
+
+Release: patch version
```
geky: Do we go over versions somewhere in docs? maybe we should link there in this page so these terms are defined? @AnotherButler?

docs/reference/contributing/guidelines/workflow.md
0xc0170: major would be 6.x , we do 5.x - minor.. .sometimes called feature release. 

@AnotherButler what shall we define here ?

geky: But we don't making breaking changes on minor releases.

(If we do those should be painful exceptions)

Looking at the current rules, what's the motivation for a contributor to make a change "feature" vs "breaking"? They both get released at the same time.

adbridge: It is just a classification. If someone rewrites an algorithm which totally changes the behaviour of the system, then that is a breaking change but not a new feature....

0xc0170: > But we don't make breaking changes on minor releases.

Yes, this should not be seen often and thus should really provide strong reason to be considered. there are usually ways to do it backward compatible, but might not always be the case.

> It is just a classification. If someone rewrites an algorithm which totally changes the behaviour of the system, then that is a breaking change but not a new feature....

Proposals for better specification of a breaking change ? It can be fix, feature or refactor if it results in breaking a user space. we can provide better examples here?

adbridge: A breaking changes is any change, other than the introduction of a new feature, which changes the user experience of the functionality.

0xc0170:
I'll add the versioning and update this

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -79,6 +79,43 @@ Each feature has a tech lead. This person is responsible for:
 - Rebasing often to track master development.
 - Reviewing any addition to the feature branch (approval required by the feature tech lead or another assigned person).
 
+##### Pull request types
+
+We consider the following pull request types.
+
+##### Fix
+
+A bug fix should be backward-compatible internal change that fixes incorrect behavior. The fix should provide a description what and how it fixes the issue. The test should be included to catch this issue in the future and confirm the correct behavior.
```
cmonr: Replace "The test should...." with either "A test should..." or "Tests should..."

docs/reference/contributing/guidelines/workflow.md
cmonr: :+1:

docs/reference/contributing/guidelines/workflow.md
cmonr: A seperate suggestion: replace suggestive text with active text.

An example: "A bug fix should be backwards-compatible..." would be transformed to "A big fix is backwards-compatible....".

Imo, this makes our enforcement job easier by making reviews stricter, and leaving negotiation leverage with us instead of suggesting that PR authors can barter for leeway.

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -79,6 +79,43 @@ Each feature has a tech lead. This person is responsible for:
...
+
+Release: patch version
+
+##### Refactor
+
+Refactors are intended for minor versions if they are not fixing specific issue as they
```
cmonr: Is it assumed that a refactor is common knowledge here? This described what they're targeted for, but not -what- they are.

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -79,6 +79,43 @@ Each feature has a tech lead. This person is responsible for:
...
+
+##### New target
+
+Adding a new target targets the patch release as it updating targets folder implementation.
+
+Release: patch version
+
+##### Feature
+
+New features targets minor releases. It can be integrated only if the feature supports most of the targets (if it requires new target HAL implementation).
```
cmonr: Echoing @geky comment about what is a major/minor/patch release, if it isn't already documented.

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -79,6 +79,43 @@ Each feature has a tech lead. This person is responsible for:
...
+
+Adding a new functionality is considered to be a feature. It does not matter if it is C++ or C.
```
cmonr: What about how the PR relates to tools? Python?

0xc0170:
Updated. It now contains versioning section + responded to all @cmonr comments.

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -79,6 +87,43 @@ Each feature has a tech lead. This person is responsible for:
...
+
+Adding a new functionality is considered to be a feature. It does not matter if it is C++, C or Python.
+
+Release: feature
+
+##### Breaking change
+
+Any change that results in breaking user space. It should have strong justification to be considered. Often the change could be made backward compatible, for example deprecate the old functionality and introduce the new replacement.
+
+Release: feature
```
0xc0170: As I added versioning here, this should be major version to be correct.

kjbracey-arm: I don't believe mbed OS 5 has ever attempted to truly follow semantic versioning, as mbed OS 3 components did - there's been some flexibility on this one. My understanding is we don't deliberately totally break stuff, like binning complete APIs, but we do allow incompatible changes at 5.X to correct API flaws if the deprecation alternative isn't straightforward. (Eg errno mapping changes).

My understanding was that there was an expectation that apps may need to make some changes for a feature release, and as such inclusion version references should be explicit, rather than automatically "take latest" like Yotta, but it should be straightforward.

We're a bit Linux kernely with version number in that regard, rather than user-space shared library. The patch boundary is relatively clear, but when to increment the top digit less so. I have a suspicion the top digit is reserved for marketing.

cmonr: > I have a suspicion the top digit is reserved for marketing.

**ಠ_ಠ**

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -60,6 +60,14 @@ Pull requests on GitHub have to meet the following requirements to keep the code
 - Avoid merging commmits. (Always rebase when possible.)
 - Pull requests should fix a bug, add a feature or refactor.
 
+#### Mbed OS release versioning
+
+Version number MAJOR.FEATURE.PATCH where:
+
+- MAJOR version for incompatible API changes
+- FEATURE version for adding functionality in backward-compatible manner
```
cmonr: Interesting. So we're abandoning the `MINOR` release terminology?

cmonr: "functionality in backward-compatible manner" to "functionality in a backwards-compatible manner"

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -79,6 +87,43 @@ Each feature has a tech lead. This person is responsible for:
 - Rebasing often to track master development.
 - Reviewing any addition to the feature branch (approval required by the feature tech lead or another assigned person).
 
+##### Pull request types
+
+We consider the following pull request types.
+
+##### Fix
+
+A bug fix is backward-compatible internal change that fixes incorrect behavior. The fix provides a description what and how it fixes the issue. Tests are included to catch the issue in the future and confirms the correct behavior.
```
cmonr: "is backward-compatible" to "is a backwards-compatible"

cmonr: "The fix provides a description what and how it fixes the issue."

I'm not sure what's trying to be said here.

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -79,6 +87,43 @@ Each feature has a tech lead. This person is responsible for:
...
+
+Release: patch
+
+##### Refactor
+
+Refactors are intended for feature releases if they are not fixing specific issue as they
```
cmonr: Is code cleanup considered a refactor?

cmonr: Also, is the newline between 102 and 103 intended?

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -79,6 +87,43 @@ Each feature has a tech lead. This person is responsible for:
...
+
+Adding a new target targets the patch release as it updating targets folder implementation.
+
+Release: patch
+
+##### Feature
+
+New features targets feature releases. It can be integrated only if the feature supports most of the targets (if it requires new target HAL implementation).
+
+Adding a new functionality is considered to be a feature. It does not matter if it is C++, C or Python.
```
cmonr: 👍

docs/reference/contributing/guidelines/workflow.md
cmonr: That sounds correct. I don't think we do our major releases quick enough to strictly enforce semanic versioning, specifically in regards to deprecation warnings.

docs/reference/contributing/guidelines/workflow.md
0xc0170: Yes, we have feature releases.

language will fix

docs/reference/contributing/guidelines/workflow.md
0xc0170: > I'm not sure what's trying to be said here.

I'll remove that one, it is in git guidance above

> "is backward-compatible" to "is a backwards-compatible"

@AnotherButler I believe both are correct, aren't they? If not, let me know and I'll do edits

docs/reference/contributing/guidelines/workflow.md
0xc0170: > Is code cleanup considered a refactor?

It really depends on the scope and changes made.

docs/reference/contributing/guidelines/workflow.md
0xc0170: @kjbracey-arm Agree. We were flexible (we keep backward compatibility as long as we can, but some breaking changes were done and were part of feature releases). 

What goes to major release then?

0xc0170:
I rebased to resolve conflicts, up to date!

docs/reference/contributing/guidelines/workflow.md
geky: Marketing strikes again!

theotherjimmy: One of my least favorite "we renamed a standard thing" instances"

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -80,7 +88,43 @@ Each feature has a tech lead. This person is responsible for:
...
+
+Release: feature
+
+##### Breaking change
+
+Any change that results in breaking user space. It should have strong justification to be considered. Often the change could be made backward compatible, for example deprecate the old functionality and introduce the new replacement.
```
0xc0170: @SenRamakri asked good question. As this is quite serious type of pull request, we might want to clarify more.

Shall I add `breaking user space (an application requires changes to continue working)` or similar.  This also leads to another good question is - what can actually break (what is considered public vs internal) ?

0xc0170:
Another round for review ! 📲

AnotherButler:
We have most/some of this content here: https://os-doc-builder.test.mbed.com/docs/development/introduction/how-we-release-arm-mbed-os.html

Would it make sense to move that content from the intro to here?

0xc0170:
> Would it make sense to move that content from the intro to here?

I can move there the versioning. And reference it here?

0xc0170:
@AnotherButler I updated the PR. As the document you shared with me already contains the info, I kept the section but just a reference  to the release document (this way people can get versioning info that we reference below). Review please

PR #477, opened 2018-03-28
5.8 links update

TeemuKultala:
@AnotherButler please review


AnotherButler:
The live 5.8 docs live on the 5.8 branch. This branch still goes to the internal test site, so we like to keep all of the links pointing to that. Because these links are correct on the 5.8 branch, I'm going to close this PR. Please feel free to reopen if this does not make sense to you.

PR #470, opened 2018-03-26
Technology/Connectivity refactoring

SeppoTakalo:
## DONE
* ~~New `Reference/Technology/Connectivity` section that gathers all, not just IP based, connectivity options.~~
* ~~LoRa will move under Reference/Technology/Connectivity/Lora~~
* ~~LoRa API will stay under Reference/API~~
* ~~Network interfaces will have new page and move away from Socket API.~~
* ~~Mesh tutorial will move to under Technology (partially already is)~~
* ~~Mesh content from Socket page will move to Reference/API/Mesh~~

## To be done in another PR
* New page about Mesh configuration under Reference/Configuration/Mesh
* Add more text to networking configuration section. Currently it is just a big blob about fields without any textual description of effects.

CC: @AnotherButler Please comment, is this going to right direction?
Any other feedback?

NOTE: Most of the bullets from that list are not yet done. I just wanted to make visible of what I'm going to do, as the PR will eventually be way too big to review. 

# **NOTE: ONLY MERGE INTO `new_engine`. DO NOT MERGE INTO 5.8.**

PR #481, opened 2018-04-03
Expand Ethernet interface intro

SeppoTakalo:
@AnotherButler Please review.

### DIFF
--- a/docs/reference/api/connectivity/networksocket/EthInterface.md
+++ b/docs/reference/api/connectivity/networksocket/EthInterface.md
@@ -1,19 +1,16 @@
 ## Ethernet
 
-The `EthInterface` provides a C++ API for connecting to the internet over Ethernet.
-By default, this class does not require any configuration. It is able to pick up the default
-Ethernet driver for the target and select correct network stack.
-
+The `EthInterface` provides a C++ API for connecting to the internet over Ethernet. By default, this class does not require any configuration. It is able to pick up the default Ethernet driver for the target and select the correct network stack.
 
 ### Usage
 
-For statically initializing the driver, just create an object without passing any parameters
+To statically initialize the driver, create an object without passing any parameters:
 
 ```cpp
 EthernetInterface eth;
 ```
 
-Then, if default configuration is enough, bringing the interface up:
+Then, if the default configuration is enough, bring up the interface:
 
 ```cpp
 nsapi_error_t status = eth.connect();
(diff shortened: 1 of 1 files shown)


docs/reference/api/connectivity/networksocket/EthInterface.md
```diff
@@ -1,19 +1,45 @@
...
+// Open a UDP socket
+UDPSocket socket;
+socket.open(&eth);
+```
 
-1. Instantiate the `EthInterface` class.
-1. Call the `connect` function.
-1. Once you connect the EthInterface, you can use it as a
-target for opening [network sockets](/docs/development/reference/network-socket.html).
+### Configuration
```
AnotherButler: I think we should move the content from this section into the connectivity configuration section.

docs/reference/api/connectivity/networksocket/EthInterface.md
```diff
@@ -1,19 +1,45 @@
 ## Ethernet
 
-The EthInterface provides a C++ API for connecting to the internet over Ethernet.
+The `EthInterface` provides a C++ API for connecting to the internet over Ethernet. By default, this class does not require any configuration. It is able to pick up the default Ethernet driver for the target and select the correct network stack.
```
AnotherButler: Is there anything else developers should know about what this API is and what it does? This section seems short.

docs/reference/api/connectivity/networksocket/EthInterface.md
SeppoTakalo: I don't see that it fits there.
That configuration section is about build time configuration.

This however, refers to the network configuration, how to get the IP address. It is just that you either call the `set_network()` or not... There is nothing to configure through JSON settings.

Should I change the tittle?
Maybe `IP address configuration` would be better to reflect this.

docs/reference/api/connectivity/networksocket/EthInterface.md
SeppoTakalo: Its hard to see what should be inserted here.

Networking architecture in general is described in #470 and Ethernet MAC drivers will be explained on their [own page](https://github.com/ARMmbed/mbed-os/blob/feature-emac/features/netsocket/emac-drivers/README.md) (to be moved to Handbook later on).

This whole API usage is concised in the 2 line example below in the Usage section.

```
EthernetInterface eth;
eth.connect();
```

SeppoTakalo:
Actually I'll merge this with #470 easier to work when I don't need to wait for PR to be merged.

PR #482, opened 2018-04-03
Platform docs cleanup

theotherjimmy:
- Break platform into 3 major sections
 - Introcude relevent concetps in those sections
 - Move detailed exampes into Callback docs


theotherjimmy:
@AnotherButler I think I got to all of the cleanup I needed to.

theotherjimmy:
@sg- Could you take a look?

docs/reference/api/platform/Callback.md
```diff
@@ -27,3 +27,156 @@ Here is an example that uses everything discussed in the [introduction to callba
 
 - [Platform overview](/docs/development/reference/platform.html).
```
sg-: This and Intro + related content should now be deleted

docs/reference/api/platform/Callback.md
```diff
@@ -27,3 +27,156 @@ Here is an example that uses everything discussed in the [introduction to callba
...
+Callback<int(float)> cb;          // A callback that takes in a float and returns an int
+Callback<void(float)> cb;         // A callback that takes in a float and returns nothing
+Callback<int()> cb;               // A callback that takes in nothing and returns an int
+Callback<void(float, float)> cb;  // A callback that takes in two floats and returns nothing
+```
+
+You can create a Callback directly from a C function or function pointer with the same type:
+
+``` c++
+void dosomething(int) {
```
sg-: Great to move the content around but this still violates the no foo/bar rule imo. Make sure to keep another task to make the example useful or tangible

docs/reference/api/platform/Callback.md
theotherjimmy: Thanks! I'll do that.

docs/reference/api/platform/Callback.md
theotherjimmy: I think I'd rather just do that now.

PR #493, opened 2018-04-12
Hide headers from connectivity page.

SeppoTakalo:
These were confusing users, as they have same text as full pages
about those technologies.

CC: @AnotherButler Please review & merge.

### DIFF
--- a/docs/reference/technology/connectivity/connectivity.md
+++ b/docs/reference/technology/connectivity/connectivity.md
@@ -6,7 +6,7 @@ Mbed OS supports a wide range of connectivity options. This section helps you to
 
 Below is a brief introduction of the different connectivity technologies that we support.
 
-### Bluetooth low energy
+#### Bluetooth low energy
 
 <span class="images">![](https://s3-us-west-2.amazonaws.com/mbed-os-docs-images/bluetooth.png)<span>Bluetooth</span></span>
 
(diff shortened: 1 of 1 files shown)


AnotherButler:
@SeppoTakalo @MarceloSalazar This change is now built on the test site. Please let me know when your teams feel comfortable with my merging the connectivity reorganization changes to the live 5.8 site.

SeppoTakalo:
Please don't merge.
I indicated in the original pull-request that this is targeting 5.9.
See here: https://github.com/ARMmbed/Handbook/pull/470

One of the APIs I'm documenting will be released only in 5.9, because its now under refactoring.

MarceloSalazar:
Is looking very good (besides minor things) - it's a shame we can't release this soon... so we'll have to wait a bit.

SeppoTakalo:
I understand. But I made a choice that I will not be documenting APIs that are going to be dropped in Mbed OS 5.9, so I started to documenting only what the outcome will be.

Marcelo, those "minor things", can you send me and email describing what you think is missing and what you would like to see here. This documentation is not yet finished, and I have allocated my time on working on it. So I'm very much welcoming any suggestion that might improve our user APIs.

Next step I'm going to do is to write/transfer porting guides regarding Ethernet & Wifi. After that I will concentrace more on user APIs (network Interfaces, sockets).

PR #497, opened 2018-04-19
Refactor Socket API links.

SeppoTakalo:
+ Use descriptive link ID
+ Use relative links, not absolute. Makes it cleaner.

For future, when I link to Socket API, I want to use link id which is `socket-api.html`, easier to remember, and cleaner.

I believe this is now according to `Publishing guide` -> `Text styling - the technical details`

CC: @AnotherButler Please review

### DIFF
--- a/docs/reference/api/api.md
+++ b/docs/reference/api/api.md
@@ -5,7 +5,8 @@ The [APIs](/docs/development/introduction/glossary.html) in this document are or
 - [Platform](/docs/development/reference/platform.html): platform modules that provide consistent user experience.
 - [Drivers](/docs/development/reference/drivers.html): analog and digital input and outputs and digital interfaces.
 - [RTOS](/docs/development/reference/rtos.html): handling tasks and events in Mbed OS.
-- [Network Socket](/docs/development/reference/network-socket.html): network sockets, Ethernet, Wi-Fi, cellular and mesh networking.
+- [Socket API](socket-api.html): network socket API for TCP/IP.
+- [Network Interfaces](network-interfaces.html): Network interfaces, Ethernet, Wifi, Cellular & Mesh.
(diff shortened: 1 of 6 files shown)


docs/reference/api/api.md
```diff
@@ -5,7 +5,8 @@ The [APIs](/docs/development/introduction/glossary.html) in this document are or
 - [Platform](/docs/development/reference/platform.html): platform modules that provide consistent user experience.
 - [Drivers](/docs/development/reference/drivers.html): analog and digital input and outputs and digital interfaces.
 - [RTOS](/docs/development/reference/rtos.html): handling tasks and events in Mbed OS.
-- [Network Socket](/docs/development/reference/network-socket.html): network sockets, Ethernet, Wi-Fi, cellular and mesh networking.
+- [Socket API](socket-api.html): network socket API for TCP/IP.
```
iriark01: A flat socket-api.html link won't work. It has to have /docs/development etc.

SeppoTakalo: Why it does not work?
Documentation tells that it should work.
https://docs.internal.mbed.com/docs/master/publishing/text-styling-the-technical-details.html#linking

iriark01:
I think you misunderstood something. Which line exactly?

SeppoTakalo:
I think you told that this is not working `[Socket API](socket-api.html)`

But document says that it works. Also, looking at the site, all documents are rendered at same directory level, so technically it should work.

iriark01:
I don't think it works. You're welcome to try.

SeppoTakalo:
OK. Lets try, and I can then fix the links if it does not.

If it works, I'm going to refactor most of my links to this, as it is much cleaner for document writer. You only need to know the generated link id, and add `.html` to end. No need to check the directory structure.

AnotherButler:
The socket-api links didn't work, but the rest did

PR #502, opened 2018-04-23
Move 6LoWPAN mesh configuration to its own page under Reference/Config

SeppoTakalo:
@AnotherButler Please review.


AnotherButler:
I like the intro section on the config page. Please use the command-line to generate the configuration parameters, so they match the format of the other config pages.

SeppoTakalo:
I did not spot any instructions to for any script.
How do I run it?

Also, I don't really like the way other configuration pages are structured. The script generated lists only contain name of the configuration key, no description of its use, not range of acceptable values.

Can I still keep the tables and append the script generated lists after them?

I originally expressed my concerns here: https://github.com/ARMmbed/Handbook/pull/366
So by just listing the configuration keys, it is not enough to instruct developers to use them.

Therefore, I wan't to have intro sections and tables listing the keys, acceptable values/ranges and use cases for keys.

theotherjimmy:
@SeppoTakalo See https://github.com/ARMmbed/Handbook/blob/new_engine/docs/reference/configuration/configuration.md#examining-available-configuration-parameters 

>The script generated lists only contain name of the configuration key, no description of its use,

If there is no description of it's use, then the `"help"` key in the configuration file is empty.

SeppoTakalo:
@theotherjimmy Thanks. And yes, you are correct. Help fields are empty.

I'll move the stuff to help section and create PR to the Mbed OS

SeppoTakalo:
Refactored the page to look like other configuration pages.
Actually most of the "help" fields were already there. I just have missed those earlier.

SeppoTakalo:
Publishing the changed also in Mbed OS https://github.com/ARMmbed/mbed-os/pull/6745

theotherjimmy:
:smiley: Thanks @SeppoTakalo

AnotherButler:
@SeppoTakalo Can I make these changes to 5.8, or are they only for 5.9?

SeppoTakalo:
These are valid for 5.8 as well.

PR #503, opened 2018-04-24
Fix Socket API link by using same ID in the "title" field in docs.json

SeppoTakalo:
* Fix the render engine bug found in https://github.com/ARMmbed/Handbook/pull/497

@iriark01 @AnotherButler Please merge in, so Socket-API links will be fixed.


AnotherButler:
I don't like changing the title to "Socket API". None of the other section in the navigation bar (side panel) have "API" in their names. I don't like the inconsistency.

SeppoTakalo:
Changed so that all links are pointing to `network-socket.html` instead of `socket-api.html`

SeppoTakalo:
And header is now unchanged.

AnotherButler:
Awesome. Thanks @SeppoTakalo

Issue #494, opened 2018-04-12
EventQueue examples not ordered correctly


bmcdonnell-ionx:
On this page

https://os.mbed.com/docs/v5.8/reference/eventqueue.html

in the example "Shared event example: deferring from interrupt context", it says, "Like the previous example, this defers from interrupt to an event queue thread." However, that's not what either of the two preceding examples do.

I think it should immediately follow the first example, which is titled "EventQueue example: deferring from interrupt context". The titles differ by only 1-2 words, and the code only by a few lines.

bmcdonnell-ionx:
Also, it would be helpful if you could provide an example using the `EventQueue::background()` function. 😄

AnotherButler:
@bmcdonnell-ionx Thanks for raising this issue. We'll take care of this at once. 

@SenRamakri @kegilbert I can move the example up to remove the confusion in the first comment. Could one of you or someone on your team please create an example using the `EventQueue:background()` function? I can create a ticket if you like.

AnotherButler:
@bmcdonnell-ionx Thanks again for raising this issue. 

We have moved the example up to prevent further confusion. We've also created an internal ticket to ensure completion of an example using `EventQueue:background()`.

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTDOCS-754*

PR #507, opened 2018-05-01
Create eventqueue-background.md

AnotherButler:
Create tutorial to fulfill issue #494

### DIFF
--- a/docs/tutorials/using_apis/eventqueue-background.md
+++ b/docs/tutorials/using_apis/eventqueue-background.md
@@ -1,0 +1,206 @@
+## The `EventQueue.background` method
+
+An [event queue](/docs/development/tutorials/the-eventqueue-api.html) moves tasks from one **execution context** to another **execution context**.
+
+In the embedded space, developers use the event queue to move IRQs from high-priority interrupt context to a low-priority threaded context:
+
+``` cpp
+EventQueue equeue;
+
+// example function
+void print_callback() {
+    printf("hi!\n");
+}
+
+// high priority interrupt context
+void irq() {
+    // we use the "call" method to pass the callback_function to a different execution context
+    equeue.call(print_callback);
+}
+
+// low priority threaded context
+int main() {
+    // hardware timer, can only execute code in interrupt context
+    Timer timer;
+    timer.attach_ms(irq, 100);
+
+    // dispatch events in our current context
+    equeue.dispatch();
+}
+```
+
+This example assigns the execution context for events by calling the `dispatch` function. `dispatch` runs the event queue in the current execution context, in this case the low-priority thread.
+
+However, `dispatch` has a downside: It consumes the whole thread. While you're running `dispatch`, you can't execute anything else.
+
+Passing a timeout to `dispatch` lets you get your execution context back for periodic events:
+
+``` cpp
+// low priority threaded context
+int main() {
+    // hardware timer, can only execute code in interrupt context
+    Timer timer;
+    timer.attach_ms(irq, 100);
+
+    while (true) {
+        // dispatch events in our current context for 100 ms
+        equeue.dispatch(100);
+
+        // super fancy status update system
+        led = !led;
+    }
+}
+```
+
+You can even pass 0 ms to execute any pending events and return immediately, which lets you spend all of your time blinking the LED.
+
+However, that's still rather limiting. If you want to do stuff in your thread, execute events _and_ sleep, you need to learn how long to sleep from the event queue. However, because any IRQ can update the event queue, the time to sleep can change, so you also need a way to receive notifications when a new sleep time has been decided.
+
+This is the use case for the event queue `background` function.
+
+``` cpp
+/** Background an event queue onto a single-shot timer-interrupt
+ *
+ *  When updated, the event queue calls the provided update function
+ *  with a timeout indicating when the queue should be dispatched. A
+ *  negative timeout is passed to the update function when the
+ *  timer-interrupt is no longer needed.
+ *
+ *  Passing a null function disables the existing update function.
+ *
+ *  The background function allows an event queue to take advantage of
+ *  hardware timers or other event loops, allowing an event queue to be
+ *  run in the background without consuming the foreground thread.
+ *
+ *  @param update   Function called to indicate when the queue should be
+ *                  dispatched
+ */
+void background(mbed::Callback<void(int)> update);
+```
+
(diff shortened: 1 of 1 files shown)


docs/tutorials/using_apis/eventqueue-background.md
```diff
@@ -0,0 +1,206 @@
...
+    blinky_equeue.call_every(100, blink);
+
+    // dispatch our root event queue
+    blinky_equeue.dispatch();
+}
+```
+
+You want to tightly control the memory consumption of each event loop.
+
+Consider an example Sonar class driven by a complicated FSM that I'm too lazy to write:
```
AnotherButler: Could an engineer please create an example to replace this one?

docs/tutorials/using_apis/eventqueue-background.md
```diff
@@ -0,0 +1,206 @@
...
+        // super fancy status update system
+        led = !led;
+    }
+}
+```
+
+You can even pass 0 ms to execute any pending events and return immediately, which lets you spend all of your time blinking the LED.
+
+However, that's still rather limiting. If you want to do stuff in your thread, execute events _and_ sleep, you need to learn how long to sleep from the event queue. However, because any IRQ can update the event queue, the time to sleep can change, so you also need a way to receive notifications when a new sleep time has been decided.
+
```
AnotherButler: Query: Could we summarize lines 6-58 with "Performing concurrent functions with an event queue can become complicated. For example, to concurrently perform operations in a thread, execute events and sleep, you need to..."?

I ask because we're spending 50+ lines to show a user how not to solve a problem. User documentation should tell users what to do, not what not to do. Similarly, user documentation can tell users why we chose the features we did, but it shouldn't say why we didn't choose the features we didn't.

AnotherButler (commented):
@kjbracey-arm @geky Could one of you please address my two comments?

docs/tutorials/using_apis/eventqueue-background.md
geky: These examples build up the problem and also demonstrate alternative solutions that a user may choose if they don't have the exact problem we're demonstrating a solution for.

> Performing concurrent functions with an event queue can become complicated. For example, to concurrently perform operations in a thread, execute events and sleep

What if the user doesn't care about sleeping? Where do they go? Should they stop reading this tutorial?

AnotherButler: If this is a tutorial about EventQueue.background, it should tell people what this is and how and when to use it. It should not give lots of solutions for lots of problems that aren't related to using this method. 

When I read this, what I got out of it is that you need EventQueue.background if "you want to do stuff in your thread, execute events _and_ sleep" because you need to learn how long to sleep from the event queue. Are there other use cases? Is this just a specific example of a much broader use? If so, it's not obvious.

That's why I'm trying to summarize a general use case for this method. That's why I'm asking if adding a sentence such as "Performing concurrent functions with an event queue can become complicated." would help. It doesn't make sense to spend 50 lines telling users how other solutions to a specific situation don't work instead of telling people when to use the method we're describing.

geky: This doesn't match what I thought the tutorials are. Why would this documentation go here and not in the reference page?

Where do we put documentation that tells users how to solve problems?

I'm happy if this is the wrong place for that to go, but I want to know where that should go.

AnotherButler: A tutorial about using the dispatch function should be a step-by-step guide about how and when to use the dispatch function. A tutorial about how to use dispatch without consuming the whole thread should be a step-by-step guide to using dispatch without consuming the whole thread and should talk about passing a timeout for 0 ms. A tutorial about EventQueue.background should be a step-by-step guide about how and when to use EventQueue.background. A tutorial about EventQueue.background should not be a guide about how to solve problems without using EventQueue.background. Does that make sense? 

If the purpose of this tutorial is how to EventQueue.background, other information about how to solve unrelated problems without using EventQueue.background don't belong on this page. They could be great information we could include somewhere else, but it seems distracting.

geky: Ah so this tutorial should be renamed to "how to run event queues without extra threads"?

Considering it ends with "here's how you use EventQueue.chain" it sounds like this document doesn't fit here at all.

(I'm still confused how "how to EventQueue.background" is not the same as the reference page)

AnotherButler: Yes, that works. I thought community developers were asking for how to use EventQueue.background, but if they're really asking how to run event queues without extra threads, then we can use that as a title and keep the beginning information here.

AnotherButler:
@geky @kjbracey-arm ping

docs/tutorials/using_apis/eventqueue-background.md
kegilbert: Petition to replace this with a link to the sonar example as per our discussion @geky @AnotherButler

cmonr:
Fyi, @kjbracey-arm is still on vacation.
Dunno about @geky ;)

docs/tutorials/using_apis/eventqueue-background.md
kegilbert: @geky @AnotherButler  Thoughts on the sonar example link?

geky: Sounds reasonable to me. That or we could just remove the sentence about laziness. The example does have the part that's important to users of the event queue.

I'd remove it but I don't think I can push to this branch.

geky:
Looks good to me except the lazy sentence

cmonr:
@kjbracey-arm Should be back now ;)

AnotherButler:
@yennster FYI
We created this PR to resolve the issue linked at the top, but feel free to close this PR if it's no longer relevant after your edits.

deepikabhavnani:
We still don;t have event queue background information in docs, this PR is relevant. Please add this to docs so we can close https://github.com/ARMmbed/mbed-os/issues/6759.

Not sure what is pending here

AnotherButler:
Because we don't have plans to resolve this PR in the near future, I'm closing this PR. We can reopen it at a later date if necessary.

40Grit:
@aglass0fmilk how do you feel about this closing?

0xc0170:
I don't understand the closure here neither. How can we fix  this one? Should this be taken over by someone else?

PR #509, opened 2018-05-03
Change warning

AnotherButler:
Change warning as requested by engineering in mbed-os PR #6645

### DIFF
--- a/docs/reference/api/platform/Wait.md
+++ b/docs/reference/api/platform/Wait.md
@@ -25,7 +25,7 @@ void wait_us(int us);
 
 ### Avoiding OS delay
 
-When you call wait, your board's CPU will sleep in the RTOS for the whole number of milliseconds and then spin to make the necessary fraction of a millisecond. However, it blocks the platform deep sleep for the entire duration.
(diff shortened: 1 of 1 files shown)


docs/reference/api/platform/Wait.md
```diff
@@ -25,7 +25,7 @@ void wait_us(int us);
 
 ### Avoiding OS delay
 
-When you call `wait`, your board's CPU is in a loop waiting for the required time to pass. Using the [Arm Mbed RTOS](/docs/development/reference/rtos.html), you can make a call to `Thread::wait` instead.
+When you call wait, your board's CPU will sleep in the RTOS for the whole number of milliseconds and then spin to make the necessary fraction of a millisecond. However, it blocks the platform deep sleep for the entire duration.
```
adbridge: think it should say "for the whole number of milliseconds and then spin as necessary to make up the remaining fraction of a millisecond"

To clarify the meaning of that, the wait api takes microseconds but presumably the RTOS sleep works in milliseconds, so if the number of microseconds is not an even multiple of milliseconds there will be a fraction left over after the RTOS sleep....

AnotherButler:
@adbridge Please rereview at your convenience.

PR #513, opened 2018-05-08
Update network socket APIs overview page

SeppoTakalo:
* Remove obsolete and outdated examples
* Add diagram from Technology page
* Explain the API a bit
* Extend the non-blocking section


AnotherButler:
This feels more like a typical API page than an overview page. Is that what you're going for?

SeppoTakalo:
It should feel like it.
UDPSocket and TCPSocket are both sharing similar behaviours. This should describe what is common for both so that we don't need to copy&paste the documentation.

So describing API behaviour here is on purpose.

Later on, I might even add more data here when it affects both of our socket types.

SeppoTakalo:
@AnotherButler Dropped the example, as you suggested.

This is now good to go from my point of view.

SeppoTakalo:
@AnotherButler Looks good. Thank you.

PR #514, opened 2018-05-08
Add documentation for BufferedBlockDevice

davidsaada:


docs/reference/api/storage/BufferedBlockDevice.md
```diff
@@ -0,0 +1,23 @@
+## BufferedBlockDevice
+
+<span class="images">![](https://os-doc-builder.test.mbed.com/docs/development/mbed-os-api-doxy/class_slicing_block_device.png)<span>BufferedBlockDevice class hierarchy</span></span>
```
AnotherButler: When the code merges, we'll need to update this link, as well as the class reference and example links.

docs/reference/api/storage/BufferedBlockDevice.md
```diff
@@ -0,0 +1,23 @@
...
+
+The BufferedBlockDevice class is a block device adpator, whose purpose is to reduce the read and program sizes of the underlying block device to 1. Large read and program sizes may make life difficult for block device users, so BufferedBlockDevice reduces both sizes to the minimum, where reads and writes to the underlying block device use an internal buffer. Calling the `sync` API ensures writes are flushed to the underlying block device.
```
dannybenor: adapter
e underlying block device to 1 byte.
I think we better use the word "write" instead of "program" because "program sizes" may be understood as program==code. I know the interface of write in BD is "to program" but still it will be more clear

docs/reference/api/storage/BufferedBlockDevice.md
```diff
@@ -0,0 +1,23 @@
...
+
+  - _bd_ -  Block device to back the BufferedBlockDevice.
+
+### BufferedBlockDevice class reference
+
+[![View code](https://www.mbed.com/embed/?type=library)](http://os-doc-builder.test.mbed.com/docs/development/mbed-os-api-doxy/class_slicing_block_device.html)
+
+### BufferedBlockDevice example
+
+This BufferedBlockDevice example takes a [HeapBlockDevice](/docs/development/reference/heapblockdevice.html), whose read size is 256 and program size is 512, and shows how one can read or program this block device with much smaller read and program sizes by using BufferedBlockDevice.
```
dannybenor: 256 bytes and 512 bytes

docs/reference/api/storage/BufferedBlockDevice.md
davidsaada: The term `program` is the one used by block devices, hence its usage.

docs/reference/api/storage/BufferedBlockDevice.md
davidsaada: Thanks, will change.

docs/reference/api/storage/BufferedBlockDevice.md
davidsaada: Sure. Had to push this PR first, as it was a requirement for the code PR to be merged.

docs/reference/api/storage/BufferedBlockDevice.md
davidsaada: Changed.

davidsaada:
Pushed a commit with an inline code snippet instead of the non-existing link.

AnotherButler:
@dannybenor Have your concerns been addressed?

dannybenor:
Yes. Good for me

AnotherButler:
I'm going to merge this for 5.9, but in the future, we'll need to see that the example makes it into teams, so that it renders with an import to compiler button.

PR #517, opened 2018-05-11
Update Arm compiler version

AnotherButler:
Change Arm compiler version to that of 5.9.


jeromecoutant:
Hi
I don't find any 5.9 version in https://developer.arm.com/products/software-development-tools/compilers/arm-compiler/downloads/version-5 ?
Thx

AnotherButler:
@jeromecoutant The Arm Mbed OS version will be 5.9, but the Arm compiler version is 5.06 update 6. Does that answer your question?

PR #520, opened 2018-05-21
Adds Network Status Callback registration page

VeijoPesonen:
Replaces already existing documentation about the subject. Explains
a bit further what is required from the application POV. @SeppoTakalo, @jarlamsa have a look and please let me know how to improve it. @AnotherButler Is it somehow possible to render the documentation(this PR) so I would see that all the links are working etc?

### DIFF
--- a/docs/reference/api/connectivity/networksocket/NetworkStatus.md
+++ b/docs/reference/api/connectivity/networksocket/NetworkStatus.md
@@ -1,0 +1,76 @@
+<h2 id="network-status">Network status</h2>
+
+Purpose of this interface is to provide a way to inform about connection state changes asynchronously. This is accomplished
+by providing a method to register a callback function to a socket. The callback gets triggered each time the network interface's state changes. 
+
+### Usage
+
+
+Possible network states the callback needs to be able to handle
+ 
+```cpp
+/** Enum of connection status types
+ *
+ *  Valid error codes have negative values.
+ *
+ *  @enum nsapi_connection_status
+ */
+ typedef enum nsapi_connection_status {
+    NSAPI_STATUS_LOCAL_UP           = 0,        /*!< local IP address set */
+    NSAPI_STATUS_GLOBAL_UP          = 1,        /*!< global IP address set */
+    NSAPI_STATUS_DISCONNECTED       = 2,        /*!< no connection to network */
+    NSAPI_STATUS_CONNECTING         = 3,        /*!< connecting to network */
+    NSAPI_STATUS_ERROR_UNSUPPORTED  = NSAPI_ERROR_UNSUPPORTED
+} nsapi_connection_status_t;
+```
+
+An interface which is to be monitored - ethernet used here only as an example
+
+```cpp
+EthernetInterface eth;
+```
+
+There is need to provide the callback function itself
+
+```cpp
+void status_callback(nsapi_event_t status, intptr_t param)
+{
+    printf("Connection status changed!\r\n");
+    switch(param) {
+        case NSAPI_STATUS_LOCAL_UP:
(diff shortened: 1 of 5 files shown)


jarlamsa (changes requested):

docs/reference/api/connectivity/networksocket/NetworkStatus.md
```diff
@@ -0,0 +1,76 @@
...
+            break;
+        default:
+            printf("Not supported");
+            break;
+    }
+}
+```
+
+Now, the callback function is to be registered to the interface. The socket needs to be set in
+non-blocking mode for the asynchronous behavior.
```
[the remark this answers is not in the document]
SeppoTakalo: NetworkInteface MAY be set to non-blocking mode to achieve asynchronous behaviour. But it is not forced.

It only blocks the first `connect()` call. Then disconnection and others will cause callback to be triggered.

So `set_blocking()` only affects the `connect()` behaviour.

Can you clarify those?

SeppoTakalo (changes requested):
Add this to the `docs.json` as well.

SeppoTakalo:
Other that those that I commented, It looks good.

SeppoTakalo (approved):
Update the documentation. This is now good to go from my point of view.

docs/reference/api/connectivity/networksocket/NetworkStatus.md
```diff
@@ -0,0 +1,82 @@
...
+
+Optionally, the application might want to set the `connect()` method to nonblocking mode and wait until connectivity is fully established.
+
+```cpp
+    eth.set_blocking(false);
+    eth.connect();              // Return immediately
+```
+
+By default, the `connect()` call blocks until `NSAPI_STATUS_GLOBAL_UP` state is reached. Some applications might require only link-ocal connectivity and therefore do not need to block that long. In those case monitoring the state changes is the preferred behavior.
+
```
AnotherButler: Query: Could you please link to the class reference in the Doxygen? I think it should be in Data Structures here, but I don't see it. Could it be somewhere else?

https://os-doc-builder.test.mbed.com/docs/development/mbed-os-api-doxy/annotated.html

AnotherButler (commented):
@VeijoPesonen Thanks for the PR 👍 Please review my copy edits to make sure I didn't accidentally change the meaning of anything, and please address my comment.

AnotherButler:
@jarlamsa Are your concerns addressed?

docs/reference/api/connectivity/networksocket/NetworkStatus.md
SeppoTakalo: I cannot find it either..
However, I don't see much benefit on Doxygen linking as it only need this `nsapi_connection_status_t` which is basically already shown here.

SeppoTakalo:
jarlamsa is travelling this week, but I would assume that yes, concerns have been addressed, as those we same as mine.
I did change the documentation to fix these.

SeppoTakalo:
Fixed conflicts.
@AnotherButler This is now ready to go in.
Jarno is travelling this week, so we cannot wait him to click "Approve".

javiermelendezc:
Documentation on connect() and disconnect() says: 

> This blocks until connection is established, but asynchronous operation can be enabled by calling NetworkInterface::set_blocking(false).

Then in set_blocking() parameters description it says:

> virtual nsapi_error_t set_blocking(bool  blocking)	
> ....
> blocking | Use true to set NetworkInterface in asynchronous mode.

I think it should be: **Use false to set NetworkInterface in asynchronous mode.**

PR #522, opened 2018-05-24
Remove UVISOR references from Handbook

danny4478:
Remove UVISOR references from Handbook since it is being deprecated from Mbed OS

* Note: An entire paragraph of semihosting was removed, since AFAIK, the UVISOR was the only module which implemented it.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -439,12 +439,7 @@
                         "sources": [{
                                 "type": "markdown",
                                 "url": "https://github.com/ARMmbed/Handbook/blob/new_engine/docs/reference/api/security/TLS.md"
-                            },
-                            {
-                                "type": "markdown",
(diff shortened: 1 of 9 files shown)


alzix:
CC @ndevillard

AnotherButler:
I thought we were putting up a deprecation notice in the Handbook as opposed to removing all references from the Handbook. @iriark01 @ndevillard @ChiefBureaucraticOfficer Please advise

danny4478:
@AnotherButler Do you mean the warning message on build (already merged) is enough?

AnotherButler:
@danny4478 We'll need to add a deprecation warning to the Handbook in addition to the one in the code. So we'll leave the uVisor handbook pages up online, but we'll add a notice to the top of them. It's crucial that product approves the message for both the Handbook and code notes.

PR #523, opened 2018-05-24
LoRaWAN API section update

hasnainvirk:
API documentation for LoRaWAN now contains information regarding events and callbacks. 
A separate section is added in configuration section for LoRaWAN.


hasnainvirk:
@SeppoTakalo @AnotherButler Please review.

docs/reference/api/connectivity/lorawan/lorawanconfig.md
```diff
@@ -1,4 +1,4 @@
-<h2 id="lorawan-configuration">Configuring Mbed LoRaWAN Stack</h2>
+<h2 id="lorawan-configuration">LoRaWAN Configuration</h2>
```
AnotherButler: We have this information in the configuration page: https://os-doc-builder.test.mbed.com/docs/development/reference/configuration-connectivity.html

docs.json
```diff
@@ -457,17 +457,21 @@
                     },
                     {
                         "title": "LoRaWAN",
-                        "intro": {
```
AnotherButler: For consistency with the other sections, I would like the section overview page to have general concepts about the technology and be followed by the API references. Because of this, I don't think we should start the section with the API reference.

docs/reference/api/connectivity/lorawan/lorawanconfig.md
hasnainvirk: The information in configuration page is not useful to anyone. At least I can't read it or make sense of it. Can you @SeppoTakalo ? I would like to see the configuration page for LoRaWAN alongside API docs.

docs.json
hasnainvirk: Other sections do start the page with API docs. And I was told that there is no requirement to have an intro. We could re-write the LoRaWANInterface section to reflect a bit about technology. But then why there is a separate Technology section dedicated only for this which you are suggesting here ?

docs/reference/api/connectivity/lorawan/lorawanconfig.md
hasnainvirk: @kjbracey-arm FYI

docs.json
AnotherButler: @hasnainvirk 

Every user API section (platform, drivers, RTOS, network socket, network interfaces, bluetooth, security and storage) starts with an overview page instead of going straight to the user API class reference. The individual class references within those sections don't require their own overview pages. Because LoRa is its own section, it requires an overview page. If it were nestled under network interfaces or something, it wouldn't require an overview page because there's already one for the section.

The section overview and technology pages are similar in that they both include reference material about our APIs and technology. However, our section overview pages include general concepts about their sections. The [platform page](https://os-doc-builder.test.mbed.com/docs/development/reference/platform.html) is a really good example of this. The technology pages are only for our stack technologies and include a deep-dive view of the technology's technical details and architecture.

docs/reference/api/connectivity/lorawan/lorawanconfig.md
AnotherButler: @hasnainvirk 

"The information in configuration page is not useful to anyone. At least I can't read it or make sense of it."

The [config section overview page](https://os-doc-builder.test.mbed.com/docs/development/reference/configuration.html) tells users how to use `mbed compile --config -v` to get configuration parameters. Each of the pages within that section show the configuration parameters for those groupings. For example, we have platform configuration parameters, RTOS configuration parameters and storage configuration parameters. I would like to add some context above the list of parameters for each page, like we see with the [mesh configuration page](https://os-doc-builder.test.mbed.com/docs/development/reference/configuration-mesh.html). 

The only difference in content between the config page in your PR and the existing connectivity config page's LoRa content is that your PR doesn't include the lora.automatic-uplink-message function.

As your page doesn't add new information, I don't understand how you find your suggested addition useful but not the existing page.

Because platform, drivers, RTOS, connectivity, mesh and storage configuration parameters are all located within the configuration section at the request of engineering, it doesn't make sense for LoRa configuration to lie elsewhere.

hasnainvirk:
@SeppoTakalo @peknis01 Can you please review the changes ?

hasnainvirk:
@AnotherButler Can you please review ?

hasnainvirk:
@SeppoTakalo @peknis01 @AnotherButler Please review.

PR #531, opened 2018-05-25
Low power timer / timeout docs updated

deepikabhavnani:
https://jira.arm.com/browse/IOTMORF-1302
https://jira.arm.com/browse/IOTMORF-1303

CC @AnotherButler @sg- @c1728p9

### DIFF
--- a/docs/reference/api/drivers/LowPowerTimeout.md
+++ b/docs/reference/api/drivers/LowPowerTimeout.md
@@ -2,13 +2,12 @@
 
 <span class="images">![](https://os-doc-builder.test.mbed.com/docs/development/mbed-os-api-doxy/classmbed_1_1_low_power_timeout.png)<span>LowPowerTimeout class hierarchy</span></span>
 
-Use the LowPowerTimeout interface to set up an interrupt to call a function after a specified delay. You can create any number of LowPowerTimeout objects, allowing multiple outstanding interrupts at the same time.
+Use the LowPowerTimeout interface to set up an interrupt to call a function after a specified delay. You can create any number of LowPowerTimeout objects. This allows multiple outstanding interrupts at the same time.
 
-### Warnings and notes
+#### Notes
 
-* No blocking code in ISR: avoid any call to wait, infinite while loop or blocking calls in general.
-
-* No printf, malloc or new in ISR: Avoid any call to bulky library functions. In particular, certain library functions (such as printf, malloc and new) are not re-entrant, and their behavior could be corrupted when called from an ISR.
+- No blocking code in ISR: Avoid any call to wait, infinite while loop or blocking calls in general.
+- No printf, malloc or new in ISR: Avoid any call to bulky library functions. In particular, certain library functions (such as printf, malloc and new) are not re-entrant, and their behavior could be corrupted when called from an ISR.
 
 ### LowPowerTimeout class reference
 
--- a/docs/reference/api/drivers/LowPowerTimeout.md
+++ b/docs/reference/api/drivers/LowPowerTimeout.md
@@ -2,7 +2,13 @@
 
 <span class="images">![](https://os-doc-builder.test.mbed.com/docs/development/mbed-os-api-doxy/classmbed_1_1_low_power_timeout.png)<span>LowPowerTimeout class hierarchy</span></span>
 
-[Add description here.]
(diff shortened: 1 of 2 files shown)


docs/reference/api/drivers/LowPowerTimer.md
```diff
@@ -2,12 +2,14 @@
 
 <span class="images">![](https://os-doc-builder.test.mbed.com/docs/development/mbed-os-api-doxy/classmbed_1_1_low_power_timer.png)<span>LowPowerTimer class hierarchy</span></span>
 
-[Add description here.]
+LowPowerTimer inherits from the Timer Class. The timer in this case continues operating even in deep sleep mode. It relies on `lp_ticker`, which is the low power ticker.
+
+You can use the LowPowerTimer interface to create, start, stop and read a timer for measuring small times (between microseconds and seconds). You can independently create, start and stop any number of LowPowerTimer objects.
```
AnotherButler: Query: What's the benefit of independently creating, starting and stopping LowPowerTimer objects? What's the alternative?

deepikabhavnani: No special benefit or alternative, its essentially how timers work or are expected to work

docs/reference/api/drivers/LowPowerTimer.md
```diff
@@ -2,10 +2,9 @@
 
 <span class="images">![](https://os-doc-builder.test.mbed.com/docs/development/mbed-os-api-doxy/classmbed_1_1_low_power_timer.png)<span>LowPowerTimer class hierarchy</span></span>
 
-LowPowerTimer inherits Timer Class, timer in this case will continue operating even in deep sleep mode. It relies on lp_ticker which is low power ticker.
+LowPowerTimer inherits from the Timer Class. The timer in this case continues operating even in deep sleep mode. It relies on `lp_ticker`, which is the low power ticker.
```
deepikabhavnani: @AnotherButler - I was not sure if we should add link to "Timer Class" here or not, your call please

docs/reference/api/drivers/LowPowerTimeout.md
```diff
@@ -2,12 +2,19 @@
 
 <span class="images">![](https://os-doc-builder.test.mbed.com/docs/development/mbed-os-api-doxy/classmbed_1_1_low_power_timeout.png)<span>LowPowerTimeout class hierarchy</span></span>
 
-[Add description here.]
+Use the LowPowerTimeout interface to set up an interrupt to call a function after a specified delay. You can create any number of LowPowerTimeout objects. This allows multiple outstanding interrupts at the same time.
```
AnotherButler: Query: When would a user want to create LowPowerTimeout objects? When would a user want to call a function after a specified delay? What's the use case here?

docs/reference/api/drivers/LowPowerTimer.md
SenRamakri: @deepikabhavnani - We also have Timer/Timeout classes. I think it would be good to capture hows LowPower versions different from generic Timer/Timeout classes, and when people should use this vs others.

MarceloSalazar:
This looks OK as an initial overview of the Low Power timer APIs and I'm happy for you to move forward with the PR.

However, I believe we need to provide context to developers on what 'Low Power' actually means when referring to Low Power modes in general (Run, Sleep, Deep-Sleep, Hibernation / Shutdown) when working with Mbed, regardless of the specific SiP implementations.

I don't think we should be providing this information here, but for sure linking to the right doc (if we have). I'd suggest a new 'Technology' section including this topic.

@AnotherButler let me know if there is something available or you'd like to raise a ticket for this.

AnotherButler:
@MarceloSalazar We cover some of that in our power management APIs: https://os-doc-builder.test.mbed.com/docs/development/reference/power-management.html

I think some of this content should eventually move into the platforms overview page under a power management concept section. Because it's not a stack technology, I hesitate to put it in Technology.

MarceloSalazar:
@AnotherButler thanks for the link!

Indeed the content here is related to power management on the platforms section.
Probably a link to that section should be enough.

AnotherButler:
@MarceloSalazar Awesome. I'll go ahead and add that.

SenRamakri:
@AnotherButler @MarceloSalazar - That's perfect, if we could create a link to PM doc from this doc, it achieves what I was looking for. Thanks for pointing that out.

PR #537, opened 2018-05-29
TLS Corrections 1

jen3andruska:
Removed 'Mbed TLS' from the Mbed TLS Entropy page.  You will probably want to remove this from the title of the page in the left column (contents) as well, but I do not think I can edit these.


jen3andruska:
So Irit has just said that we can't have 'Mbed' alone (without TLS), so what do you want 'Mbed TLS' replaced with?  Let me know and I'll change all of these.

PR #541, opened 2018-05-29
Add proper sections to Connectivity configuration.

SeppoTakalo:
* Also add some missing configuration fields.
* LoRA dropped from this list because it will be supplied in #523 

@AnotherButler Please review.


docs/reference/configuration/Connectivity.md
```diff
@@ -1,39 +1,23 @@
 <h2 id="configuration-connectivity">Connectivity</h2>
 
+This page describes build-time configurable parameters for connectivity in Mbed OS. There are a few categories of configurable connectivity parameters, so we have structured this page between different connectivity methods:
 
-You can configure various parameters using either C++ APIs or the Arm Mbed configuration system. The following is the complete list of connectivity configuration parameters, as generated by `mbed compile --config -v`. Please see [the configuration system documentation](/docs/development/reference/configuration.html) for details on how you may use or override these settings.
```
AnotherButler: @SeppoTakalo I'd like to include a line about using `mbed compile --config -v` to show users how to get these parameters and to match the other pages. Does that sound right with you?

SeppoTakalo: I removed the section, because it is quite clearly explained in "Configuration" section: https://os.mbed.com/docs/v5.8/reference/configuration.html in chapter "Examining available configuration parameters"

However, I agree with the view of having one sentence about using `mbed compile --config -v` just to give a hint for user to check current values.

I'll add that section back.

SeppoTakalo:
Added back the removed section.

Looks ok now.

PR #548, opened 2018-05-31
Adding a note about ESP8266.

peknis01:
The warning was first changed to note and then removed from the Cloud docs, as the proper home for it would be Mbed OS as suggested by Alon Shamir.

### DIFF
--- a/docs/reference/contributing/connectivity/NetworkStack.md
+++ b/docs/reference/contributing/connectivity/NetworkStack.md
@@ -130,6 +130,8 @@ When adding a new connectivity class, you can use `mbed test` to verify your imp
 
 This example ports a driver for the ESP8266 Wi-Fi module to the NSAPI.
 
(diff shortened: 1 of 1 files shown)


SeppoTakalo:
Can we add this into the ESP driver page in Github?
(With a note that we don't even support the HW flow control in software)

peknis01:
@SeppoTakalo, feel free to suggest an appropriate location for the note. We may as well just leave it out from the docs totally, if there is no sense having it.

SeppoTakalo:
The driver already has "Serial port does not have hardware flow control enabled. " section under limitations. I feel like it should be enough.

Having the word "not suitable for production usage" is just assumption that customer is stupid and does not understand limitations without very verbal guiding.

peknis01:
Fair enough. I think @AnotherButler can close this PR then. Note not needed, right?

PR #554, opened 2018-06-01
Introduced docs changes to indicate Python 3 compatability

cmonr:


MarceloSalazar:
@cmonr thanks for the update on docs.

As discussed with @BlackstoneEngineering, it's important that we clarify the backwards compatibility problem, as Python 3 tools won't work with OS < v5.8.

cmonr:
@MarceloSalazar Did you have something more specific in mind aside from https://github.com/ARMmbed/Handbook/pull/554/commits/e388e15de8588b14b8a82ef12a27c76c6e2ffe30 ?

MarceloSalazar:
@cmonr the last update is great! - I was reviewing an old version :)

What would happen if someone attempts to use Python 3 with a previous Mbed OS?
Would it should a clear error to the user?

cmonr:
@MarceloSalazar mbed-cli would most likely error with something cryptic, similar to what @ashok-rao first got when he tried to test on Windows this week.

However, I don't like the idea of introducing something that would do a live version check in mbed-cli, since mbed-cli is often used with detached versions of mbed-os, making version confirmation much more nuanced and time consuming.

MarceloSalazar:
> making version confirmation much more nuanced and time consuming.

This is a serious UX problem and not acceptable.

We can't release something that is known to cause serious python traceback errors for many Mbed users.

If we don't have time to fix the tools, then let's delay the update on the docs until it's done, and set expectations accordingly.

cmonr:
@MarceloSalazar Should be doing a similar check with the pip version compatibility?

https://github.com/ARMmbed/Handbook/pull/549/files

MarceloSalazar:
If we decide to go ahead, I'd like to have a known issue open with the release and start working to deliver a fix in a patch release

@adbridge FYI

theotherjimmy:
A fix for what exactly @MarceloSalazar?

MarceloSalazar:
@theotherjimmy see error and proposed solution in https://github.com/ARMmbed/mbed-cli/issues/686

AnotherButler:
So do we have a code dependency for this document?

cmonr:
@AnotherButler I think so? I'm not sure what a code dependency implies in Handbook PRs.

There are a couple of moving pieces still, but here's the remaining list:

- https://github.com/ARMmbed/mbed-cli/pull/689: Adding a message to warn users of Py3 backwards-incompatabiltiy with Mbed OS <=5.8
  - Discussion still ongoing about if anything else is needed: https://github.com/ARMmbed/mbed-cli/issues/686
  - Needs to be merged, and Mbed CLI released to 1.7.3.
- https://github.com/ARMmbed/mbed-os/pull/7092: Fixes for Windows execution, specific for Py3.
  - PR moving through CI, should be merged soon.
- https://github.com/ARMmbed/mbed-os/issues/7101: Issue in the process of being resolved. No PR as of yet.

AnotherButler:
Should we wait for any of the mbed-os PRs to merge before merging this doc PR?

cmonr:
@AnotherButler I think we're fine to move forward with this PR outside of the above requirements. In any case, the above notes are all making quick progress on being resolved.

PR #560, opened 2018-06-04
Add descriptions to the storage configuration

geky:
Storage config descriptions. Note this is specific to LittleFS and NVStore.

@davidsaada, @offirko, I added a little section for NVStore. Let me know if it's wrong or anything needs to be added. Feel free to add commits to this PR if it helps.


docs/reference/configuration/Storage.md
```diff
@@ -1,21 +1,47 @@
...
+
+LittleFS provides several configuration options that you can use to tweak the performance of the file system on different hardware. By default, this file system finds the optimal configuration from the underlying block device's geometry, but you can override this to optimize special situations. For example, if your device has a large amount of RAM, you can increase the `read_size` and `prog_size` configuration options for a minor speed improvement.
+
+Note that LittleFS has 4 levels of debug logging. By default, all logging is enabled except for `enable_debug`. Setting `enable_debug` to `true` makes the log output very verbose, and the output is useful for bug reports.
+
 ```
 Configuration parameters
 ------------------------
 
-Name: filesystem.present
```
AnotherButler: Query: Does this config parameter still exist? If so, we should document it, and I can create an additional section for it.

AnotherButler:
@davidsaada @offirko Could one of you please answer my query in Chris's absence when you review this file?

docs/reference/configuration/Storage.md
```diff
@@ -64,10 +64,9 @@ Name: littlefs.enable_warn
     No value set
 ```
 
-
 ### NVStore configuration
 
-NVStore does not need much configuration, relying only on the regions of internal flash specified in the `area_*_address` and `area_*_size` for the two areas. Additionally, `max_keys` can be used to manage the amount of RAM needed for NVStore keys. Note that `max_keys` will default to the number of keys needed by Mbed OS and only needs to be modified if an application uses NVStore directly.
+NVStore does not need much configuration. It relies only on the regions of internal flash specified in the `area_*_address` and `area_*_size` for the two areas. Additionally, you cna use `max_keys` to manage the amount of RAM NVStore keys needs. Note that `max_keys` defaults to the number of keys Mbed OS needs. You only need to be modify it if an application uses NVStore directly.
```
offirko: Typo: "can" instead of "cna"
Typo: "You only need to modify it" instead of "You only need to be modify it"

AnotherButler: This is why everyone needs an editor. Thank you for catching these.

geky: Oh wow, what moron wrote this?

docs/reference/configuration/Storage.md
geky: It's not a parameter that can be configured. It's really a workaround to get the build tools to indicate which libraries are available.

I noticed this nsapi.present config option was omitted from the network config, so I did the same here.

docs/reference/configuration/Storage.md
geky: Sorry, that comment was very inappropriate. I thought it was my own writing that was being corrected and tried to make a joke.

Either way, it wasn't an acceptable comment to write. @AnotherButler thanks for all the help editting these docs, I'm sure I haven't made it easy for you to catch all of the writing mistakes.

AnotherButler: @geky No worries. It's always a pleasure editing your docs.

AnotherButler:
@offirko Is this OK to merge?

offirko:
Its ok with me, I guess if others have no more comments then its fine.

PR #564, opened 2018-06-04
Devicekey updates

dannybenor:
According to https://jira.arm.com/browse/IOTDOCS-506


dannybenor:
@iriark01  @ashok-rao  please review

iriark01:
Yes yes, all right. Not allowed to use etc :)

PR #571, opened 2018-06-05
Add NVStore example content to Handbook

AnotherButler:
Add requested content from the NVStore example to the Handbook.


dannybenor:
@davidsaada I think Ashok is right. Can you double check? The issue is because I copied from NVStore documentation

davidsaada:
> I think Ashok is right. Can you double check? The issue is because I copied from NVStore documentation

Indeed he's right.

AnotherButler:
If I change "mbed_lib.json" to "mbed_app.json", will this fix the problem?

ashok-rao:
Thanks @dannybenor and @davidsaada for the confirmation.
@AnotherButler : Yes, please change the _lib.json to _app.json. Rest all LGTM!

PR #589, opened 2018-06-13
workflow: add triage comment

0xc0170:
I would like to address new triage comments in Github issues. To inform users we triage, and leave some comments in the issues + label. Please review

This is addressing questions like this one https://github.com/ARMmbed/mbed-os/issues/7190#issuecomment-396563501


0xc0170:
cc @ARMmbed/mbed-os-maintainers

PR #591, opened 2018-06-15
Clarification of CRC API

MarceloSalazar:
- Propose to remove 'Mbed' from the side panel, as it seems redundant.
https://os.mbed.com/docs/v5.9/reference/mbedcrc.html

- Clarified that the CRC API can take advantage of the HW CRC feature provided by the platform.

@AnotherButler @bulislaw


MarceloSalazar:
If that's the agreement, that's fine. I've reverted to the original name.

PR #603, opened 2018-06-20
Update Bootstrap.md

AnotherButler:
Add content from old docs.


AnotherButler:
I was tasked with going through some old documentation and adding anything still relevant to our current docs. Could you please review this and let me know if it's still applicable?

bulislaw:
That's a bit outdated I think. @c1728p9 will be refactoring bootstrap soon, it would be good if we could document the new process here.

PR #610, opened 2018-06-27
Update socket documentation based on current refactored design

SeppoTakalo:
New design contain Abstract Socket class that can be used directly by
applications. Casting 'Socket' pointers back to 'TCPSocket' or 'UDPSocket'
is no longer necessary.

New design also emphasis usage of 'SocketAddress' for holding the IP
addresses, instead on textual format.

New design also deprecates the TCPServer completely, moving its
functionality directly info 'TCPSocket' itself.


SeppoTakalo:
@kjbracey-arm Please review from technical point of view.

@AnotherButler Please review the language

docs/reference/api/connectivity/networksocket/networksocket.md
```diff
@@ -43,15 +43,69 @@ sock.recv(buf, 100);
...
-### Network socket classes
+### Changes in Mbed OS 5.10
+
+Mbed OS Socket API has gone under refactoring for Mbed OS 5.10 release. For most of the applications, these changes are not noticeable as legacy behaviour is still emulated in `TCPSocket` and `UDPSocket` classes.
+
+New design contain Abstract Socket interface that can be used directly by
+applications. Casting `Socket` pointers back to `TCPSocket` or `UDPSocket`
+is no longer necessary.
+
+Downcasting any protocol specific class to `Socket` has no side effect and is recommended API design. Knowing the exact type is only required when socket is created.
```
kjbracey-arm: That's actually upcasting.

SeppoTakalo: Good catch. I always thought these terms as a opposite direction.
Down -> towards the base, as base is..bottom..
but actually in UML diagrams, it is up.

kjbracey-arm: I didn't actually know for sure - was doubtful and had to look it up before commenting.

docs/reference/api/connectivity/networksocket/socket.md
```diff
@@ -0,0 +1,40 @@
+## Socket
+
+<span class="images">![](https://os-doc-builder.test.mbed.com/docs/development/mbed-os-api-doxy/class_socket.png)<span>Socket class hierarchy</span></span>
+
+Socket is the abstract base class for all of the protocol-specific socket types. It defines all of the functions that consist of the Mbed OS Socket API. You cannot directly create a Socket object because it is abstract, but you can upcast any protocol-specific object to an abstract Socket object.
```
AnotherButler: Query: When we say "It defines all of the functions that consist of the Mbed OS Socket API", do we mean all of the functions that make up the Socket API?

docs/reference/api/connectivity/networksocket/networksocket.md
```diff
@@ -43,15 +43,63 @@ sock.recv(buf, 100);
 sock.close();
 ```
 
-### Network socket classes
+### Changes in Mbed OS 5.10
```
AnotherButler: We should probably have Product review the wording of this section. @iriark01 Do you agree?

docs/reference/api/connectivity/networksocket/socket.md
kjbracey-arm: Yes.

docs/reference/api/connectivity/networksocket/TCPSocket.md
```diff
@@ -6,22 +6,28 @@ The TCPSocket class provides the ability to send a stream of data over TCP. TCPS
...
-Here is a TCP client example of HTTP transaction over Ethernet interface.
+Here is a client example of HTTP transaction over TCPSocket:
 
-[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/TCPSocket_Example/)](https://os.mbed.com/teams/mbed_example/code/TCPSocket_Example/file/6b383744246e/main.cpp)
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed-os-examples/code/mbed-os-example-sockets/)](https://os.mbed.com/teams/mbed-os-examples/code/mbed-os-example-sockets/file/e0496f3424a8/main.cpp/)
 
 ### Related content
 
-- [TCPServer](/docs/development/reference/tcpserver.html) API reference.
+- [Socket](socket.html) API reference.
```
iriark01: Why was this URL changed? Will that work?

docs/reference/api/connectivity/networksocket/networksocket.md
iriark01: Probably, yes. It seems to be a pretty big change in its own way.

AnotherButler: @maansaari Could you please review this section to make sure we're positioning this change properly from a Product perspective?

docs/reference/api/connectivity/networksocket/TCPSocket.md
AnotherButler: This PR deletes the TCPserver page and adds a new socket page. The link change will work.

SeppoTakalo: Short links do work. This is also mentioned in the documentation.
Much cleaner to maintain those.

SeppoTakalo: URL was changed because we want to deprecate `TCPServer` class, so I'm linkin only to the main Socket API documentation that describes the common behaviour between TCP and UDP protocols.

iriark01: I think they work *within a book*; the danger is getting into the habit of keeping short links and then not knowing what to change when the structure changes, or not using full links between books (out of habit).

AnotherButler: These pages are all in the same book [APIs]. These links should be fine.

docs/reference/api/connectivity/networksocket/networksocket.md
maansaari: These changes are along direction we want to go with the Socket interface, especially with secure sockets and "non-IP" sockets, so this will make the Socket higher level abstraction, which servers to future needs.

PR #612, opened 2018-06-28
Updates to AnalogIn.md

mrcoulter45:
Included information about the successive-approximation ADC type and resolution in AnalogIn.md

### DIFF
--- a/docs/reference/api/drivers/AnalogIn.md
+++ b/docs/reference/api/drivers/AnalogIn.md
@@ -2,6 +2,12 @@
 
 Use the AnalogIn API to read an external voltage applied to an analog input pin. `AnalogIn()` reads the voltage as a fraction of the system voltage. The value is a floating point from `0.0`(VSS) to `1.0`(VCC). For example, if you have a 3.3V system and the applied voltage is 1.65V, then `AnalogIn()` reads `0.5` as the value.
 
+ One of the most common types of analog to digital converters used in microcontrollers today is called the successive-approximation ADC. Successive-approximation is a popular choice in modern microcontrollers because of the fact that it is accurate, low-power, and takes up a small amount of space inside of the microcontroller. 
+ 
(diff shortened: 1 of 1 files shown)


docs/reference/api/drivers/AnalogIn.md
```diff
@@ -2,6 +2,12 @@
 
 Use the AnalogIn API to read an external voltage applied to an analog input pin. `AnalogIn()` reads the voltage as a fraction of the system voltage. The value is a floating point from `0.0`(VSS) to `1.0`(VCC). For example, if you have a 3.3V system and the applied voltage is 1.65V, then `AnalogIn()` reads `0.5` as the value.
 
+ One of the most common types of analog to digital converters used in microcontrollers today is called the successive-approximation ADC. With successive-approximation, the analog input to the ADC is essentially guessed in a binary search-like fashion by the ADC hardware, which consists of a voltage comparator, a DAC, and a successive approximation register (SAR), which will hold the result of the conversion. For a 12-bit ADC, after 12 clock cycles, the value in the SAR will converge on the value of the analog input, thus completing the voltage sampling process.
```
SenRamakri: Overall the document looks great, particularly the resolution section. It would be nicer if we can make the DAC description even more shorter. The focus here is to explain the API and not necessarily how the DAC works.

SenRamakri:
@AnotherButler - Can you please add Kevin G as well to this review?

kegilbert:
Really like the step size resolution information since it adds some context to the first segment. I'm OK with the general description as well for the bit about how many clock cycles it takes to sample the ADC. Some of the information goes a bit far though for the API page. Do we have a technologies page for drivers @AnotherButler like the network stack was looking to add? Not sure if we have enough info to fill that out for drivers but some basic technology docs like this might be nice to have somewhere.

mrcoulter45:
I went ahead and just removed the super-technical information about how successive-approximation works, and just gave some general information about it.

PR #613, opened 2018-06-28
Updates to Error docs to capture the new config names and added example

SenRamakri:
Updates to Error docs to capture the new config names and added example


docs/reference/api/platform/Error.md
```diff
@@ -69,7 +69,7 @@ The error handling system in Mbed OS automatically captures the thread context f
 - Thread stack size – Size of stack; this information is captured only if the system is built with RTOS support.
 - Thread stack mem - Top of stack; this information is captured only if the system is built with RTOS support.
 - Thread current SP – SP value (PSP or MSP based on what’s active); this information is captured only if the system is built with RTOS support.
-- File name and line number - By default, capturing of file name and line number is disabled. If you need this information to be captured, you need to enable it by using the **MBED_CONF_ERROR_FILENAME_CAPTURE_ENABLED** macro.
```
kegilbert: Should we rephrase this and the other renamed changes from "...you need to enable it by using the **X** macro" to something like "...you need to set the configuration option **X** to true." ?

PR #616, opened 2018-07-01
Rename LOWPOWERTIMER to LPTICKER

c1728p9:
Update the name in the docs to reflect the code.

PR #617, opened 2018-07-01
Rename LOWPOWERTIMER to LPTICKER

c1728p9:
Update the name in the docs to reflect the code.


c1728p9:
This is the same as #616 but is for the development branch

PR #619, opened 2018-07-02
Migration: New porting guide from Cloud team

theotherjimmy:
Original text from @jen3andruska:
> New porting guide for porting Mbed OS before porting Mbed Cloud

PR #625, opened 2018-07-10
Add content from PR #619

AnotherButler:
Add content to this page, so all porting target content is in one place.

### DIFF
--- a/docs/reference/contributing/target/target.md
+++ b/docs/reference/contributing/target/target.md
@@ -1,10 +1,46 @@
-<h2 id="target-port">Targets</h2>
+<h2 id="target-port">Porting targets</h2>
 
-Adding a new microcontroller to Arm Mbed OS 5 depends on CMSIS-CORE and CMSIS-Pack. Please make sure that the microcontroller already has these available.
+This document discusses requirements to port Mbed OS to a derivative target of any one of the Mbed OS Enabled targets. To port the MCU to Mbed OS, as well, along with a custom board based on the MCU, please follow our [more detailed porting guide](docs/development/reference/porting-targets.html). 
 
-#### Adding a new microcontroller and board
+Adding a new microcontroller to Arm Mbed OS 5 depends on CMSIS-CORE and CMSIS-Pack. The microcontroller must have these available.
 
-First fork the `mbed-os` repository on GitHub into your own user account. We will use the placeholder `USERNAME` to refer to your username in the following documentation, `MCU_NAME` to refer to the new microcontroller you are adding and `BOARD_NAME` to refer to the new board you are adding. Import an Mbed OS example, and add your fork of `mbed-os` using:
+[Mbed Cloud Client](https://cloud.mbed.com/docs/current) provides reference implementation for three Mbed Enabled targets, the K64F, NUCLEO_F429ZI and the UBLOX_EVK_ODIN_W2. You do not need to port either Mbed OS or the Cloud Client if you are using one of these three boards.
+
+### Summary of steps
+
+1. Add your target to `targets.json`
+   1. Identify the your target’s features.
+   1. Identify any existing parent targets.
+   1. Identify macros required for your target's compilation.
+   1. Identify supported compilers and the Mbed OS release version.
+   1. Add the target entry in `targets.json`.
+   1. Verify that you can now pass your target to `mbed compile`.
+1. Add your target port.
+   1. Create the directory structure.
+   1. Add an entry for the target in `mbed_rtx.h`.
+   1. Add startup code and CMSIS headers, NVIC headers and **all** relevant drivers from CMSIS specifications.
+   1. Implement APIs, HAL, the system clock configuration and any other additional clocks.
+   1. Add toolchain-specific linker descriptions.
+1. Add peripherals and pin names for the target.
+1. Compile with CLI on a supported compiler.
+
+### Target properties
+
+You can identify our sample target by the following:
+
+- Target build name `MY_BOARD_1`.
+- Vendor `MY_VENDOR`, device family `MY_FAMILY` and the device on the board `MY_DEVICE`.
+- A Cortex-M4F based core.
+- Porting for Mbed OS v5.x, which requires support for Arm Compiler 5, GCC Arm Embedded and IAR EWARM.
+- A serial (UART) interface connected to the CMSIS-DAP implementation.
+- Requirement that the macro `CPU_DEVICE_1` compiles the vendor-provided HAL for the device mounted on the board.
+- Requirement that the macro `FAMILY_MY_FAMILY` compiles the vendor-provided HAL for the family of the device.
+
+All of these requirements directly map to relevant tags in the `targets.json` entry for our target. This is shown in the following step.
+
+### Adding a new microcontroller and board
+
+First fork the `mbed-os` repository. This example uses the placeholder `USERNAME` to refer to your username in the following documentation, `MCU_NAME` to refer to the new microcontroller you are adding and `BOARD_NAME` to refer to the new board you are adding. Import an Mbed OS example, and add your fork of `mbed-os` using:
 
 ```
 mbed import mbed-os-example-blinky
@@ -17,7 +53,7 @@ git branch my-new-target -u USERNAME
 cd ..
 ```
 
-#### Target description
+### Target description
 
 Add the target description to `mbed-os\targets\targets.json` using keys that the [Adding and configuring targets section](/docs/development/tools/adding-and-configuring-targets.html) describes. We recommend that you run the [targets lint script](/docs/development/tools/adding-and-configuring-targets.html#style-guide) on your target hierarchy before submitting your pull request:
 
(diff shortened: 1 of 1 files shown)


jen3andruska:
@AnotherButler Finished reviewing & copy-editing.

docs/reference/contributing/target/target.md
```diff
@@ -1,10 +1,46 @@
...
+   1. Add the target entry in `targets.json`.
+   1. Verify that you can now pass your target to `mbed compile`.
+1. Add your target port.
+   1. Create the directory structure.
+   1. Add an entry for the target in `mbed_rtx.h`.
+   1. Add startup code, CMSIS headers, NVIC headers and **all** relevant drivers from the CMSIS specifications.
+   1. Implement APIs, HAL, the system clock configuration and any other additional clocks.
+   1. Add toolchain-specific linker descriptions.
+1. Add peripherals and pin names for the target.
+1. Compile with CLI on a supported compiler.
```
ashok-rao: CLI -> Mbed CLI

docs/reference/contributing/target/target.md
```diff
@@ -34,9 +70,173 @@ Add the target description to `mbed-os\targets\targets.json` using keys that the
...
+1. Ensure that `Objects.h` declares peripherals and is available in the `/api` directory.
+
+### Linker files
+
+Each supported toolchain listed in your new target's `supported_toolchains` configuration needs a linker file. Each compiler supports a different style of linker files with a different extension:
+
+- Add a scatter file, with the extension `.sct`, for Arm Compiler 5.
+- Add a linker script, with the extension `.ld`, for GCC Arm Embedded.
+- Add an IAR Linker file, with the extension `.icf`, for IAR EWARM.
+
```
ashok-rao: @AnotherButler : Can you please add: "You will also need to add the startup codes (*.s) in the respective directories as well for ARMCC, GCC and IAR."

docs/reference/contributing/target/target.md
```diff
@@ -34,9 +70,173 @@ Add the target description to `mbed-os\targets\targets.json` using keys that the
...
+}
+#endif
+
+#endif
+```
+
+### Compiling with a supported compiler
+
+You can verify that the new board port compiles by using an [example application](/docs/development/tutorials/mbed-os-quick-start.html), and checking out the branch containing your port in the `mbed-os` subdirectory.
+
```
ashok-rao: @AnotherButler : Can you please add: "With the Mbed CLI, the command to use in the example's root directory is: mbed compile -t <TOOLCHAIN> -m <YOUR_NEW_TARGET's_NAME> -c . And if you face any build errors at this point, please add any missing headers, add/remove peripherals as required, ensure the linker scripts & startup code are in place. For simplicity, you can start with only the serial peripheral and add other peripherals once the Mbed OS port builds successfully."

theotherjimmy: @AnotherButler, @ashok-rao  Don't add the `-c` it's not needed.

docs/reference/contributing/target/target.md
```diff
@@ -1,10 +1,46 @@
...
+This document discusses requirements for porting Mbed OS to a derivative target of any one of the Mbed Enabled&trade; targets. To port the MCU to Mbed OS as well, along with a custom board based on the MCU, please follow our [more detailed porting guide](docs/development/reference/porting-targets.html). 
 
-#### Adding a new microcontroller and board
+Adding a new microcontroller to Arm&reg; Mbed OS 5 depends on CMSIS-CORE and CMSIS-Pack. The microcontroller must have these available.
 
-First fork the `mbed-os` repository on GitHub into your own user account. We will use the placeholder `USERNAME` to refer to your username in the following documentation, `MCU_NAME` to refer to the new microcontroller you are adding and `BOARD_NAME` to refer to the new board you are adding. Import an Mbed OS example, and add your fork of `mbed-os` using:
+[Mbed Cloud Client](https://cloud.mbed.com/docs/current) provides reference implementation for three Mbed Enabled targets: the K64F, NUCLEO_F429ZI and the UBLOX_EVK_ODIN_W2. You do not need to port Mbed OS or the Cloud Client if you are using one of these three boards.
```
theotherjimmy: This is not related to porting for Mbed OS (it's mbed cloud client only) and should be moved elsewhere.

docs/reference/contributing/target/target.md
```diff
@@ -1,10 +1,46 @@
...
+   1. Identify your target’s features.
+   1. Identify any existing parent targets.
+   1. Identify macros required for your target's compilation.
+   1. Identify supported compilers and the Mbed OS release version.
+   1. Add the target entry in `targets.json`.
+   1. Verify that you can now pass your target to `mbed compile`.
+1. Add your target port.
+   1. Create the directory structure.
+   1. Add an entry for the target in `mbed_rtx.h`.
+   1. Add startup code, CMSIS headers, NVIC headers and **all** relevant drivers from the CMSIS specifications.
```
theotherjimmy: I don't think a bold "all" here does anyone any good.

docs/reference/contributing/target/target.md
```diff
@@ -1,10 +1,46 @@
...
+- Porting for Mbed OS v5.x, which requires support for Arm Compiler 5, GCC Arm Embedded and IAR EWARM.
+- A serial (UART) interface connected to the CMSIS-DAP implementation.
+- Requirement that the macro `CPU_DEVICE_1` compiles the vendor-provided HAL for the device mounted on the board.
+- Requirement that the macro `FAMILY_MY_FAMILY` compiles the vendor-provided HAL for the family of the device.
+
+All of these requirements directly map to relevant tags in the `targets.json` entry for our target. This is shown in the following step.
+
+### Adding a new microcontroller and board
+
+First fork the `mbed-os` repository. This example uses the placeholder `USERNAME` to refer to your username in the following documentation, `MCU_NAME` to refer to the new microcontroller you are adding and `BOARD_NAME` to refer to the new board you are adding. Import an Mbed OS example, and add your fork of `mbed-os` using:
```
theotherjimmy: "your username" -> "your github username" 

Without that we could be refering to os.mbed.com or something else.

docs/reference/contributing/target/target.md
theotherjimmy: @ashok-rao There is no reason that startup code has to be assembly, and that startup code cannot be shared between compilers. Assembly must have a capitol `.S` suffix.

theotherjimmy:
@bulislaw The only cloud related stuff is that single paragraph in the opening. Everything else is strictly Mbed OS.

docs/reference/contributing/target/target.md
ashok-rao: @theotherjimmy :This is coming from the Cloud Client porting. There will be a section in the Cloud docs as well that points to this porting guide (correct?? @jen3andruska ). This is intended for folks who plan to (ultimately) port Cloud Client, but need Mbed OS running on their custom platform in the first place. Once they have Mbed OS running on their custom platform, users will link back to Cloud docs for the "rest" of the MCC porting (@AnotherButler : can you link to the MCC porting doc from here that @jen3andruska is working on?). This doc creates a bridge between Cloud porting and OS porting - which are both required for a custom platform.

theotherjimmy: @ashok-rao I dont' think that it's worded well if that's the case.

-------------------

I also think that linking from the mbed-os docs to the cloud docs implies that to get mbed-os support you also need cloud support. I think it has much more potential to do harm in this way than it has the potential to do good by helping those porting cloud.

AnotherButler:
@SenRamakri Could you please review this and make the changes requested by @bulislaw in his absence?

docs/reference/contributing/target/target.md
SenRamakri: I have to agree with @theotherjimmy. I'm not sure why we need to mention the fact that MCC implements porting for k64f or other platforms here. This is a document about porting Mbed-OS and I don't think we need to mention MCC ports available for Mbed-OS, it actually confuses me to read that in introduction. What I would suggest is create a section named "Related Links or Other reference" and may be we can capture things like "Also look into... MCC ports, HAL porting guides".

docs/reference/contributing/target/target.md
```diff
@@ -34,9 +70,173 @@ Add the target description to `mbed-os\targets\targets.json` using keys that the
...
+      └── mbed_rtx.h
+```
+
+### HAL porting
 
-There are many more APIs to implement. You enable the following APIs by adding a `device_has` attribute to the MCU_NAME target definition in `targets.json` and providing an implementation of the API declared in the API header.
+There are many more APIs to implement. You enable the following APIs by adding a `device_has` attribute to the `MCU_NAME` target definition in `targets.json`, and providing an implementation of the API declared in the API header.
 
 device_has       |   API header
 -----------------|------------------
```
SenRamakri: It would be useful to create a link for each of the below header files to actual files in mbed-os repo. Not a big deal, but would be convenient.

docs/reference/contributing/target/target.md
```diff
@@ -1,10 +1,46 @@
...
+   1. Identify macros required for your target's compilation.
+   1. Identify supported compilers and the Mbed OS release version.
+   1. Add the target entry in `targets.json`.
+   1. Verify that you can now pass your target to `mbed compile`.
+1. Add your target port.
+   1. Create the directory structure.
+   1. Add an entry for the target in `mbed_rtx.h`.
+   1. Add startup code, CMSIS headers, NVIC headers and **all** relevant drivers from the CMSIS specifications.
+   1. Implement APIs, HAL, the system clock configuration and any other additional clocks.
+   1. Add toolchain-specific linker descriptions.
```
SenRamakri: It would be good to mention the HAL tests we have under mbed-os\TESTS\mbed_hal which can be used to verify HAL porting.

docs/reference/contributing/target/target.md
```diff
@@ -1,10 +1,46 @@
...
+1. Compile with CLI on a supported compiler.
+
+### Target properties
+
+You can identify our sample target by the following:
+
+- Target build name `MY_BOARD_1`.
+- Vendor `MY_VENDOR`, device family `MY_FAMILY` and the device on the board `MY_DEVICE`.
+- A Cortex-M4F based core.
+- Porting for Mbed OS v5.x, which requires support for Arm Compiler 5, GCC Arm Embedded and IAR EWARM.
```
SenRamakri: This might change, as we are adding ARMv8M platforms which only need to support ARMC6 for now. So, depending on Core Archicture we may support all of these compilers or ARMC6 only if its ARMv8M based core.

ashok-rao: Agree. Let's add a line here for v8 only requirements.

docs/reference/contributing/target/target.md
theotherjimmy: We could at least make them a link to their respective porting pages.

ashok-rao:
@AnotherButler : Any news on this please? Thanks.

AnotherButler:
@bulislaw Once @ashok-rao addresses the comments from @SenRamakri are you OK with this merging?

docs/reference/contributing/target/target.md
```diff
@@ -34,9 +70,173 @@ Add the target description to `mbed-os\targets\targets.json` using keys that the
...
+### Sample directory structure
+
+Below is an example of all of the files that you would add for a target from vendor `MY_VENDOR`, with an MCU `MY_DEVICE_1`, from device family `MY_FAMILY_1`, mounted on the board `MY_BOARD_1` and supporting the `SERIAL` device:
+
+```
+mbed-os
+└── targets
+   └── TARGET_MY_VENDOR
+      ├── TARGET_MY_FAMILY
+      │  ├── TARGET_MY_DEVICE_1
```
0xc0170: I like this tree, I was explaining the structure this week to at least 2 issues regarding not having `MCU_`. 

Can we use `TARGET_MCU_DEVICE_1` rather than `MY_DEVICE_1` ? vendor - family - mcu - board. This is how we do it for couple of targets.

Same is mentioned in directory structure above as `TARGET_<specific_MCU>` . I would keep there `MCU` word

docs/reference/contributing/target/target.md
```diff
@@ -1,10 +1,46 @@
-<h2 id="target-port">Targets</h2>
+<h2 id="target-port">Porting targets</h2>
 
-Adding a new microcontroller to Arm Mbed OS 5 depends on CMSIS-CORE and CMSIS-Pack. Please make sure that the microcontroller already has these available.
+This document discusses requirements for porting Mbed OS to a derivative target of any one of the Mbed Enabled&trade; targets. To port the MCU to Mbed OS as well, along with a custom board based on the MCU, please follow our [more detailed porting guide](docs/development/reference/porting-targets.html).
```
bulislaw: That doesn't make any sense. This document is about porting the MCU... We are trying to merge two pages talking about slightly different thing. Oh and the `docs/development/reference/porting-targets.html` doesn't exists as far as i can tell.

docs/reference/contributing/target/target.md
bulislaw: Why are we mentioning MCC here, we are talking about porting the OS, after the OS is ported we can start porting the MCC.

docs/reference/contributing/target/target.md
```diff
@@ -1,10 +1,46 @@
...
 
-#### Adding a new microcontroller and board
+Adding a new microcontroller to Arm&reg; Mbed OS 5 depends on CMSIS-CORE and CMSIS-Pack. The microcontroller must have these available.
 
-First fork the `mbed-os` repository on GitHub into your own user account. We will use the placeholder `USERNAME` to refer to your username in the following documentation, `MCU_NAME` to refer to the new microcontroller you are adding and `BOARD_NAME` to refer to the new board you are adding. Import an Mbed OS example, and add your fork of `mbed-os` using:
+[Mbed Cloud Client](https://cloud.mbed.com/docs/current) provides reference implementation for three Mbed Enabled targets: the K64F, NUCLEO_F429ZI and the UBLOX_EVK_ODIN_W2. You do not need to port Mbed OS or the Cloud Client if you are using one of these three boards.
+
+### Summary of steps
+
+1. Add your target to `targets.json`.
```
bulislaw: That assumes cloud porting, not the system.

docs/reference/contributing/target/target.md
ashok-rao: I think the confusing part is "To port the MCU to Mbed OS as well, along with a custom board based on the MCU, please follow our [more detailed porting guide](docs/development/reference/porting-targets.html). " ...

@AnotherButler : Would it be clear if we just mention "To port an MCU itself (rather than a derived target of the MCU), please refer to the detailed porting guide <link to porting MCU porting guide> here" ?
Maybe we also need to mention clearly somewhere, what "porting an MCU" vs "porting a derived target" means as well.. before the start of this document.

@theotherjimmy .. thoughts??

docs/reference/contributing/target/target.md
ashok-rao: @bulislaw : That's the ultimate goal. In order to reach the top floor (Pelion Cloud Client), we need to clear the ground floor (OS port running successfully on whatever board the customer has) .. and no probs if they use K64 / F429 / ODIN as these are already working in the examples.. however, the problem comes when they use an F429 MCU on a custom PCB (not NUCLEO) or a K64 (not FRDM) or ODIN (not EVK / MTB..) .. this guide should help them get Mbed OS running on those custom boards 1st. Then, they can move to PCC..

docs/reference/contributing/target/target.md
ashok-rao: Doesn't a new target needs to be added here anyways for building from CLI ?

ashok-rao:
@donatieng : I know you are quite occupied with the release, but could you please have a look at this PR? I've addressed previous comments & now that DM is becoming a part of the OS, I feel this doc needs to be absolutely put in place as well.. could you please let me know? Thanks.

docs/reference/contributing/target/target.md
donatieng: Where's the "more detailed porting guide"? Also, this page should address porting any new boards/MCU - not just a derivative.

docs/reference/contributing/target/target.md
donatieng: Not happy with the current wording which is confusing; as said above this page is about porting the OS.
However you could add a note somewhere: "Note: This page describes the porting process of Mbed OS which is a Pelion Cloud Client dependency; to port Cloud Client, go to this page..."

docs/reference/contributing/target/target.md
donatieng: Agree with @theotherjimmy

docs/reference/contributing/target/target.md
donatieng: Yup, a link would be valuable too

docs/reference/contributing/target/target.md
donatieng: Is that requirement in the right place? Not sure whether you're documenting terminology or requirements

docs/reference/contributing/target/target.md
```diff
@@ -1,10 +1,46 @@
...
+This document discusses requirements for porting Mbed OS to a derivative target of any one of the Mbed Enabled&trade; targets. To port the MCU to Mbed OS as well, along with a custom board based on the MCU, please follow our [more detailed porting guide](docs/development/reference/porting-targets.html). 
 
-#### Adding a new microcontroller and board
+Adding a new microcontroller to Arm&reg; Mbed OS 5 depends on CMSIS-CORE and CMSIS-Pack. The microcontroller must have these available.
 
-First fork the `mbed-os` repository on GitHub into your own user account. We will use the placeholder `USERNAME` to refer to your username in the following documentation, `MCU_NAME` to refer to the new microcontroller you are adding and `BOARD_NAME` to refer to the new board you are adding. Import an Mbed OS example, and add your fork of `mbed-os` using:
+[Mbed Cloud Client](https://cloud.mbed.com/docs/current) provides reference implementation for three Mbed Enabled targets: the K64F, NUCLEO_F429ZI and the UBLOX_EVK_ODIN_W2. You do not need to port Mbed OS or the Cloud Client if you are using one of these three boards.
+
+### Summary of steps
```
donatieng: Can we have one clear section per step?

docs/reference/contributing/target/target.md
```diff
@@ -34,9 +70,173 @@ Add the target description to `mbed-os\targets\targets.json` using keys that the
...
+/mbed-os/targets/TARGET_<Manf>/TARGET_<Device_Family>/TARGET_<specific_MCU>/TARGET_<specific_board>/
+```
+
+This sample implementation uses:
+
+```
+/mbed-os/targets/TARGET_MY_VENDOR/TARGET_MY_FAMILY/TARGET_MY_DEVICE_1/TARGET_MY_BOARD_1/
+```
+
+### Entry in `mbed_rtx.h`
```
donatieng: That looks good but shouldn't it go in `rtos.md`?

iriark01:
Closing; we're getting new porting content that will also cover this

PR #632, opened 2018-07-19
Update SPI docs with multiple slave scenario

SenRamakri:
Update SPI docs with multiple slave scenario and usage

### DIFF
--- a/docs/reference/api/drivers/SPI.md
+++ b/docs/reference/api/drivers/SPI.md
@@ -23,7 +23,7 @@ Mode |  Polarity |  Phase
 
 The SPI master generates a clock to synchronously drive a serial bit stream slave. The slave returns a bit stream, also synchronous to the clock. 
 If you want to communicate with multiple slave devices connected over same SPI peripheral, you must use multiple SPI objects, one for each slave but instantiate them with different SSEL(Slave select) and configuration corresponding to each slave.
-Note that SPI object will automatically configure the SPI peripheral with the right configuration when the application code invokes the interfaces on the SPI object.
+Note that SPI object will automatically configure the SPI peripheral with the current object's configuration when the application code invokes the interfaces on the SPI object.
 
 ### SPI class reference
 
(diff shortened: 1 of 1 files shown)


SenRamakri:
@AnotherButler, @kegilbert, @deepikabhavnani  - Please review.

docs/reference/api/drivers/SPI.md
```diff
@@ -23,6 +23,10 @@ Mode |  Polarity |  Phase
 
 The SPI master generates a clock to synchronously drive a serial bit stream slave. The slave returns a bit stream, also synchronous to the clock.
 
+To communicate with multiple slave devices connected over the same SPI peripheral, you must use multiple SPI objects, one for each slave, but instantiate them with different slave selects (SSEL) and configurations corresponding to each slave.
+
+Note that the SPI object automatically configures the SPI peripheral with the right configuration when the application code invokes the interfaces on the SPI object.
```
kegilbert: Sorry I'm still spinning up my brain from being out last week. I'm not quite clear on what you mean by 
>...application code invokes the interfaces on the SPI object.
Does this mean when I call the SPI object constructor the SPI hardware will be configured with the provided parameters based on the target automatically?

kegilbert:
Very good snippet to add, thanks! Only had a small clarification question in reference to the second statement posted inline as a comment above. @SenRamakri

docs/reference/api/drivers/SPI.md
SenRamakri: Yes, the SPI constructor calls spi_init() with the provided configuration.

kegilbert: Awesome, just making sure. Maybe something like the following would make it a bit more clear? Took me a reread to get what it was going after initially. 

> Note that the SPI object automatically configures the hardware peripheral with the right configuration when the application code invokes the SPI object constructor. 

Or I'm just derping and it's clear as is, I can go either way on this one.

AnotherButler:
@deepikabhavnani When you get a chance, could you please review this?

PR #633, opened 2018-07-19
Update toolchain support for ARMv8M

deepikabhavnani:
GCC_ARM has issues, only ARMC6 supported

CC @SenRamakri

### DIFF
--- a/docs/tools/mbed_targets.md
+++ b/docs/tools/mbed_targets.md
@@ -86,7 +86,7 @@ The name of the target's Arm core.
 
 Possible values: `"Cortex-M0"`, `"Cortex-M0+"`, `"Cortex-M1"`, `"Cortex-M3"`, `"Cortex-M4"`, `"Cortex-M4F"`, `"Cortex-M7"`, `"Cortex-M7F"`, `"Cortex-A9"`, `"Cortex-M23"`, `"Cortex-M23-NS"`, `"Cortex-M33"`, `"Cortex-M33-NS"`
 
-<span class="notes">**Note:** Mbed OS supports v8-M architecture (Cortex-M23 and Cortex-M33) devices only with the `ARMC6` toolchain (Version Version 6.10).</span>
+<span class="notes">**Note:** Mbed OS supports v8-M architecture (Cortex-M23 and Cortex-M33) devices only with the `ARMC6` toolchain (version 6.10).</span>
 
 #### `public`
 
--- a/docs/tools/mbed_targets.md
+++ b/docs/tools/mbed_targets.md
@@ -86,7 +86,7 @@ The name of the target's Arm core.
 
 Possible values: `"Cortex-M0"`, `"Cortex-M0+"`, `"Cortex-M1"`, `"Cortex-M3"`, `"Cortex-M4"`, `"Cortex-M4F"`, `"Cortex-M7"`, `"Cortex-M7F"`, `"Cortex-A9"`, `"Cortex-M23"`, `"Cortex-M23-NS"`, `"Cortex-M33"`, `"Cortex-M33-NS"`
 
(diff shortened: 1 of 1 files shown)


docs/tools/mbed_targets.md
```diff
@@ -86,7 +86,7 @@ The name of the target's Arm core.
 
 Possible values: `"Cortex-M0"`, `"Cortex-M0+"`, `"Cortex-M1"`, `"Cortex-M3"`, `"Cortex-M4"`, `"Cortex-M4F"`, `"Cortex-M7"`, `"Cortex-M7F"`, `"Cortex-A9"`, `"Cortex-M23"`, `"Cortex-M23-NS"`, `"Cortex-M33"`, `"Cortex-M33-NS"`
 
-<span class="notes">**Note:** Mbed OS supports v8-M architecture (Cortex-M23 and Cortex-M33) devices only with the `GCC_ARM` toolchain.</span>
+<span class="notes">**Note:** Mbed OS supports v8-M architecture (Cortex-M23 and Cortex-M33) devices only with the `ARMC6` toolchain (version 6.10).</span>
```
SenRamakri: Please remove (Cortex-M23 and Cortex-M33) as ARM also has also announced Cortex-M35P based on ARMv8M-Mainline and it will continue to grow. So its better just to say v8-M architecture. It would be nice if we could re-phrase this sentence. This sounds to me like "With ARMC6 we only support v8-Ms". Although its true, it will change in future. So, can you please re-phrase it?

deepikabhavnani: @SenRamakri - Hence we mention which cores we support.. Cortex-M35P is not supported

deepikabhavnani: Based on support in future we can update this.. But for now Mbed OS does support only M23 / M33 of v8M series with ARMC6 only.. Also ARMC6 is only for v8 architecture not for all Cortex-M devices

deepikabhavnani: Based on above understanding, @AnotherButler please suggest what should we rephrase to?

SenRamakri: Its ok to say we support M23/M33 ony, so the modifier "only" applies to cores here or the compiler? Yep, lets get Amanda's opinion.

AnotherButler: Maybe "Mbed OS supports v8-M architecture devices on the ARMC6 toolchain (version 6.10). To see all Mbed Enabled boards, please see our [boards page](https://os.mbed.com/platforms/)."?

SenRamakri: That sounds ok to me, but linking to boards page might be a problem as I don't know if we are really going to list Nuvoton board there and even if we list its only M23.

deepikabhavnani: We don't have any Mbed board at present.. Nuvoton M2351 is not mbed Enabled.

deepikabhavnani: Mbed OS supports v8-M architecture devices (Cortex-M23 and Cortex-M33)  on the ARMC6 toolchain (version 6.10). 

?

SenRamakri: I see, how about we say "Mbed OS supports v8-M architecture devices on the ARMC6 toolchain (version 6.10). Currently Mbed-OS supports Cortex-M23 and Cortex-M33 devices". Does that sound any better?

deepikabhavnani: Fine for me

SenRamakri: I'm ok with - "Mbed OS supports v8-M architecture devices (Cortex-M23 and Cortex-M33) on the ARMC6 toolchain (version 6.10)."

PR #640, opened 2018-07-26
Add instructions on how to compile multiple projects with one copy of Mbed OS

yennster:
This PR adds instructions on how a user can use the MBED_OS_DIR config within Mbed CLI to manage multiple Mbed projects that use the same version of Mbed OS (and therefore the same Mbed OS directory/copy).


yennster:
@AnotherButler 🎉

PR #643, opened 2018-07-27
Changes to USBMouse.md and USBMouseKeyboard.md

mrcoulter45:
- add USBMouse Joystick Example
- edit USBMouseKeyboard Example


docs/reference/api/usb/USBMouse.md
```diff
@@ -68,6 +68,75 @@ int main(void)
 }
 ```
 
+### USBMouse joystick example
+
+This example uses a Grove - Thumb Joystick to act as a mouse. In this example, you can use the joystick to move the mouse around the screen and click the joystick down to do a mouse left click.
+
+```C++
+//Grove - Thumb Joystick mouse
+//Constructor is blocking, so main does not run if USBPhy is not connected
```
c1728p9: You could probably use `the target's USB` in place of `USBPhy`

PR #644, opened 2018-07-27
Expand tutorials_intro.md

AnotherButler:
Expand tutorials overview page to explain the content of the sections. Major releases, in particular, had the potential to confuse readers.

### DIFF
--- a/docs/tutorials/tutorials_intro.md
+++ b/docs/tutorials/tutorials_intro.md
@@ -1,3 +1,19 @@
 ## Arm Mbed tutorials
 
-The Arm Mbed ecosystem is expansive and offers many opportunities. In contrast to other sections of the documentation, which provide background reference material, this section provides step-by-step instructions for specific tasks you may wish to perform. These include Blinky, our example application; applications that build on Blinky; and debugging tasks. There are also tutorials about using the Arm Mbed Online Compiler, optimizing memory and optimizing power.
+The Arm Mbed ecosystem is expansive and offers many opportunities. In contrast to other sections of the documentation, which provide background reference material, this section provides step-by-step instructions for specific tasks you may wish to perform. These include our quick start guide, example applications that build on it and debugging tasks. There are also tutorials about using the Arm Mbed Online Compiler, optimizing memory and optimizing power.
+
+The [Mbed OS quick start](mbed-os-quick-start.html) is our quick start guide, a tutorial that shows you how to set up and run an example application.
+
+[Major releases](major-releases.html) discusses the difference between Mbed OS 2 and Mbed OS 5 and answers frequently asked questions about using the two major versions of Mbed OS. This section also teaches you how to migrate projects to Mbed OS 5.
+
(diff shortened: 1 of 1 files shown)


AnotherButler:
@iriark01 Does this initial expansion require additional review?

docs/tutorials/tutorials_intro.md
```diff
@@ -1,3 +1,19 @@
...
+The Arm Mbed ecosystem is expansive and offers many opportunities. In contrast to other sections of the documentation, which provide background reference material, this section provides step-by-step instructions for specific tasks you may wish to perform. These include our quick start guide, example applications that build on it and debugging tasks. There are also tutorials about using the Arm Mbed Online Compiler, optimizing memory and optimizing power.
+
+The [Mbed OS quick start](mbed-os-quick-start.html) is our quick start guide, a tutorial that shows you how to set up and run an example application.
+
+[Major releases](major-releases.html) discusses the difference between Mbed OS 2 and Mbed OS 5 and answers frequently asked questions about using the two major versions of Mbed OS. This section also teaches you how to migrate projects to Mbed OS 5.
```
iriark01: Makes me think this doesn't belong in the tutorials. Maybe move it to the new porting area?

docs/tutorials/tutorials_intro.md
```diff
@@ -1,3 +1,19 @@
...
+
+[Using the Mbed Online Compiler](using-the-mbed-online-compiler.html) uses screenshots and detailed instructions to teach you how to use our online IDE. It guides you through the process of creating, developing and publishing a process and includes advanced steps, such as collaborating with multiple authors.
```
iriark01: "the process... and publishing a process"?

docs/tutorials/tutorials_intro.md
```diff
@@ -1,3 +1,19 @@
...
+
+[Serial communication](serial-communication.html) teaches you how to install a serial driver, communicate with your board over USB and program and debug with terminal applications.
+
+The [Using APIs](using-apis.html) shows you how to use our APIs to perform specific tasks. From creating an event loop to building a private LoRa network, these tutorials include RTOS APIs, drivers and connectivity APIs.
```
iriark01: Either remove "the" or add a word after "using APIs"

iriark01: My concern is that a lot of the wording in this intro is very specific to the current state of the docs and you'll need to remember to rewrite every time you add or remove something

docs/tutorials/tutorials_intro.md
AnotherButler: That should be program. Good catch.

docs/tutorials/tutorials_intro.md
AnotherButler: We had it in porting once upon a time, but there was a comment about how we were mixing porting concepts and migrating Mbed 2 to Mbed OS 5 is not the same as porting targets, so they don't belong in the same section.

AnotherButler: It probably doesn't belong in the intro, either.

PR #646, opened 2018-07-27
Update mbed_targets.md

Alex-EEE:
The file the CLI is looking for is "extra_targets.json," not custom_targets.json ( I think an older version of mbed CLI used custom_targets.json )  I just tried this at the command line and confirmed this.  Also, this PR talks about extra_targets https://github.com/ARMmbed/mbed-os/pull/4103


Alex-EEE:
Nevermind, I confused myself and now I don't see custom_targets or extra_targets working.  I'll file an issue against mbed-os

PR #661, opened 2018-08-13
Fix EventQueue API tutorial bugs

yennster:
Fixed EventQueue API tutorial bugs
- Updated events_ex_1 here: https://os.mbed.com/teams/mbed_example/code/events_ex_1/file/69c11c7877b6/main.cpp/


geky:
@yennster, it looks like the added printf (in the fall handler) will run in irq context?

yennster:
@geky Yes, that's because this tutorial is weird in that it shows you the incorrect way to use interrupt handlers first, and then explains how you fix it.

geky:
Hmmm, ok, I think we changed before because ironically it was wrong

PR #664, opened 2018-08-15
Link to testing tools fixed.

peknis01:


iriark01:
Weirdly seems correct in 5.9, even though the original complaint was about 5.9

PR #666, opened 2018-08-15
FlashIAP driver link fix.

peknis01:
In general, all links seem to be the full paths. Shouldn't they be just the short forms, starting with /docs...?


iriark01:
Fixed in 5.9

PR #676, opened 2018-08-22
Update QSPI porting guide

maciejbocianski:
Align docs with changes introduced in:
https://github.com/ARMmbed/mbed-os/pull/7817
https://github.com/ARMmbed/mbed-os/pull/7783

### DIFF
--- a/docs/reference/contributing/target/QuadSPI.md
+++ b/docs/reference/contributing/target/QuadSPI.md
@@ -4,8 +4,6 @@ Implementing QSPI enables Mbed OS to communicate with external memories much fas
 
 The most common use case is an external memory to use as additional data storage.
 
-<span class="warnings">**Warning:** We are changing the QSPI HAL API in an upcoming release of Mbed OS. This page documents code that exists on a feature branch of Mbed OS. You can find details on how it may affect you in the [Implementing the QSPI API](#implementing-quadspi) section.</span>
-
 ### Assumptions
 
 #### Defined behavior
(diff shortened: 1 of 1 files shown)


docs/reference/contributing/target/QuadSPI.md
```diff
@@ -53,4 +75,8 @@ To enable the QSPI HAL, define `QSPI` in the targets.json file inside `device_ha
 
 ### Testing
 
-To be implemented
+The Mbed OS HAL provides a set of conformance tests for the QSPI interface (**Note: QSPI HAL tests require onboard QSPI flash memory**). You can use these tests to validate the correctness of your implementation. To run the QSPI HAL tests, use the following command:
```
bulislaw: > **Note: QSPI HAL tests require onboard QSPI flash memory**

I would say it's still a valid use case to support QSPI even if the board doesn't have on board qspi flash. Could we only run the test requiring onboard flash if the QSPI_FLASH1_xxx are defined? @donatieng @jamesbeyond

maciejbocianski: @bulislaw Actually QSPI test is run only on targets defining `DEVICE_QSPI`.
Currently `DEVICE_QSPI` means device with onboard flash, so we don't recognize QSPI capable devices without onboard QSPI flash. I think that it would be good to have something like: `DEVICE_QSPI` and `DEVICE_QSPI_FLASH` to distinguish them.

bulislaw: Why do we only consider devices with onboard flash? Not every SPI device has SPI memory attached to it...

maciejbocianski: Not every. 
e.g. Nucleo-L496ZG  has no QSPI flash but it has QSPI signals routed out on external connector according schematics (waiting for STM confirmation on it in https://github.com/ARMmbed/mbed-os/pull/7817)

0xc0170: > I would say it's still a valid use case to support QSPI even if the board doesn't have on board qspi flash. Could we only run the test requiring onboard flash if the QSPI_FLASH1_xxx are defined? @donatieng @jamesbeyond

👍  Wording would be "requires QSPI FLASH pins to be defined" ? Targets with onboard flash would have it by default, others would need a config to add them.

docs/reference/contributing/target/QuadSPI.md
```diff
@@ -29,6 +29,28 @@ To make sure your platform is ready for the upcoming changes, you need to implem
```
AnotherButler: Query: Can we delete or change the warning in line 7?

kegilbert: The code was merged it looks like, should be good to remove that warning,

maciejbocianski: @AnotherButler warning was removed

AnotherButler:
Have all comments been addressed?

AnotherButler:
FYI @MarceloSalazar @ashok-rao

AnotherButler:
@donatieng @0xc0170 Is this OK to merge?

donatieng:
LGTM 😄!

PR #680, opened 2018-08-23
Add porting guide for bootloader

scartmell-arm:


AnotherButler:
I'm surprised to see only links to existing content. Are we expecting any new content?

scartmell-arm:
> I'm surprised to see only links to existing content. Are we expecting any new content?

There might be additional details added on how to specify bootloader binary location, but the necessary documentation for porting seems to already be contained in existing documentation.

MarceloSalazar:
@AnotherButler thanks for sharing this.

I think for now is ok to link to existing content - in the future would be nice to centralize bootloader related content in one place in os.mbed.com

It would be great to add links to the mbed-bootloader, which is what Partners/Customers need to use to create a bootloader for our device management platform.
https://github.com/ARMmbed/mbed-bootloader

docs/reference/contributing/target/bootloader.md
```diff
@@ -0,0 +1,18 @@
+## Bootloader
+
+To enable firmware update capability via bootloader on a new target instructions as well as an example of how to do it can be found in the mbed OS 5 handbook.
```
ashok-rao: Hmm.. this statement isn't very clear to me.. bootloader is only required to enable FW update on the device via Pelion DM.. it's not done "via" the BL - sounds more like there are other options too (which is not the case today)..so, something like "please find instructions and an example in the handbook to update the FW using the bootloader" sounds more like setting the scene.. @AnotherButler : can you please amend this to make it more an introduction / set the scene type.. thanks..

MelindaWeed:
I'll take a look.

docs/reference/contributing/target/bootloader.md
```diff
@@ -0,0 +1,17 @@
+## Bootloader
+
+Arm Mbed OS allows you to update your firmware. Our Pelion Device Management platform enables this capability. The following documentation includes instructions and examples on how to update firmware on a new target.
```
MelindaWeed: Nitpick, but I don't think Pelion Device Management is the platform. Pelion Device Management is one of the services available under the Pelion IoT Platform umbrella.

docs/reference/contributing/target/bootloader.md
```diff
@@ -0,0 +1,17 @@
...
+
+### Dependencies
+
+The target needs to support the Flash HAL API. You can find the guide to porting the Flash API in the [Mbed OS porting guide](flash.html).
+
+### Porting the bootloader
+
+You can learn to port the update client bootloader to new targets in [the porting guide](https://cloud.mbed.com/docs/current/porting/porting-the-update-client.html).
```
MelindaWeed: Update Client, but the jury's still out on this one for now.

AnotherButler:
@ashok-rao Does the new introductory part resolve your comments?

AnotherButler:
Note to self: Add to .json after merging.

docs/reference/contributing/target/bootloader.md
```diff
@@ -1,18 +1,17 @@
 ## Bootloader
 
-To enable firmware update capability via bootloader on a new target instructions as well as an example of how to do it can be found in the mbed OS 5 handbook.
+Arm Mbed OS allows you to update your firmware. Our Pelion IoT platform enables this capability. The following documentation includes instructions and examples on how to update firmware on a new target.
```
ashok-rao: 1. ..update your "device" firmware.
2. Remove "new" in the last line as the target isn't new while updating the FW.
3. Add after ..this capability: "One of the dependencies is a bootloader running on the device capable of updating the device firmware. The following docs include..." 

Should be good to go with this.. thanks @AnotherButler

AnotherButler: Good catches

PR #687, opened 2018-08-27
Remove outdated references to FEATURE_NANOSTACK, COMMON_PAL or others.

SeppoTakalo:
In Mbed OS 5.10 all network related "features" have been removed.
This is transparent change to applications, as those code blocks are
now always enabled in the build.
Only visible change is that applications do not need to explicitly
request FEATURE_NANOSTACK or others, and targets do not need to enable
FEATURE_LWIP to provide ethernet driver.

@AnotherButler Please review.
@MarceloSalazar Please review from technical point of view.

### DIFF
--- a/docs/reference/configuration/mesh.md
+++ b/docs/reference/configuration/mesh.md
@@ -23,7 +23,6 @@ An example of the configuration file:
 {
     "target_overrides": {
         "*": {
-            "target.features_add": ["NANOSTACK", "COMMON_PAL"],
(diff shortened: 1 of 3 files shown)


SeppoTakalo:
Actually, FEATURE_NANOSTACK was already dropped in 5.9.
LWIP was dropped from master, but not merged to 5.9.

Now in 5.10 all are enabled by default, so there is not need for those any more.

PR #688, opened 2018-08-27
Removing mention of the export 'projectfiles' directory

bridadan:
`mbed export` hasn't placed exported projects in the `projectfiles` directory for quite some time. They are now placed in the root directory.

FYI @theotherjimmy


AnotherButler:
ping @theotherjimmy

theotherjimmy:
It's correct.

PR #693, opened 2018-08-29
Document Source Discovery rules

theotherjimmy:
Changes already in Mbed OS

### DIFF
--- a/docs/tools/build_rules.md
+++ b/docs/tools/build_rules.md
@@ -37,18 +37,18 @@ The target labeled directories are used for selecting sources that are specific 
 
 ```json
 {
-    "FREESCALE": {
+    "MCUXPRESSO": {
         "inherits": ["Target"],
         "public": false
     },
     "TEENSY3_1": {
-        "inherits": ["FREESCALE"],
+        "inherits": ["MCUXPRESSO"],
         "extra_labels": ["K20XX", "K20DX256"]
     }
 }
 ```
 
-The above example will include source files in directories named `TARGET_FREESCALE` and `TARGET_TEENSY3_1`, from the inheritance heirachy and `TARGET_K20XX` and `TARGET_K20DX256` from the `target.extra_labels` value. It will not include source files within `TARGET_NORDIC`, `TARGET_K66F` or `TARGET_NUCLEO_F411` directories as they are not in the list of labels included in the label type `TARGET`
+The above example will include source files in directories named `TARGET_MCUXPRESSO` and `TARGET_TEENSY3_1`, from the inheritance heirachy and `TARGET_K20XX` and `TARGET_K20DX256` from the `target.extra_labels` value. It will not include source files within `TARGET_NORDIC`, `TARGET_K66F` or `TARGET_NUCLEO_F411` directories as they are not in the list of labels included in the label type `TARGET`
 
 #### Feature Directories
 
--- a/docs/tools/build_rules.md
+++ b/docs/tools/build_rules.md
@@ -78,7 +78,7 @@ The component labeled directories are used for software that implements some fun
 }
 ```
 
-In the above example, `mbed compile` will include files in directories named `COMPONENT_SPIF`, and not directories such as `COMPONENT_SD` or `COMPNONENT_FLASHIAP`.
+In the above example, `mbed compile` will include files in directories named `COMPONENT_SPIF`, and not directories such as `COMPONENT_SD` or `COMPONENT_FLASHIAP`.
 
 #### Toolchain Directories
 
--- a/docs/tools/build_rules.md
+++ b/docs/tools/build_rules.md
@@ -1,39 +1,38 @@
-## Mbed OS Build Rules
+## Mbed OS build rules
 
-The Mbed OS Build Tools scan your project for source files every time you compile. This document describes the rules that the build tools use to decide which source files to include in each build. The Mbed OS Build Tools use the [target configuration](mbed_targets.md) found in `targets.json`, `mbed_app.json` and `mbed_lib.json` as input to these rules.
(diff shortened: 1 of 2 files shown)


docs/tools/build_rules.md
```diff
@@ -0,0 +1,118 @@
+## Mbed OS Build Rules
+
+The Mbed OS Build Tools scan your project for source files every time you compile. This document describes the rules that the build tools use to decide which source files to include in each build. The Mbed OS Build Tools use the [target configuration](mbed_targets.md) found in `targets.json`, `mbed_app.json` and `mbed_lib.json` as input to these rules.
+
+The Build Tools include every source file found in the project unless it is in a label directory, it is in a test directory or it matches a pattern in an `.mbedignore file`.
```
TacoGrandeTX: Suggest identifying what constitutes a source file... especially concerned about *.asm/*.s for assembly

docs/tools/build_rules.md
```diff
@@ -0,0 +1,118 @@
...
+```json
+{
+    "NRF52_DK": {
+        "inherits": ["Target"],
+        "components": ["SPIF"]
+    }
+}
+```
+
+In the above example, `mbed compile` will include files in directories named `COMPONENT_SPIF`, and not directories such as `COMPONENT_SD` or `COMPNONENT_FLASHIAP`.
```
TacoGrandeTX: Typo on "COMPONENT_FLASHIAP"

docs/tools/build_rules.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Toolchain Directories
+
+The toolchain labeled directories are used for toolchain specific files such as assembly or linker files. The compilers use the following label sets:
+
```
TacoGrandeTX: I think this is more readable if alphabetically sorted by Toolchain

docs/tools/build_rules.md
theotherjimmy: That particular thing might change soon. I can document what we currently consider a source file.

docs/tools/build_rules.md
theotherjimmy: @TacoGrandeTX It helps if you respond to the first/second line in the table (instead of the lines above it) as it makes it possible to see what you're talking about, and it will collapse when I change the order.

theotherjimmy:
@MarceloSalazar Good point. Renamed to MCUXpresso

docs/tools/build_rules.md
theotherjimmy: Added by f77a96f

docs/tools/build_rules.md
theotherjimmy: fixed by 60dfad4

docs/tools/build_rules.md
```diff
@@ -0,0 +1,134 @@
...
+        "extra_labels": ["K20XX", "K20DX256"]
+    }
+}
+```
+
+The above example will include source files in directories named `TARGET_MCUXPRESSO` and `TARGET_TEENSY3_1`, from the inheritance heirachy and `TARGET_K20XX` and `TARGET_K20DX256` from the `target.extra_labels` value. It will not include source files within `TARGET_NORDIC`, `TARGET_K66F` or `TARGET_NUCLEO_F411` directories as they are not in the list of labels included in the label type `TARGET`
+
+#### Feature Directories
+
+The feature labeled directories are used for software that implements some funtionality that requires some amount of space even when it's not used. The set of directories include by the `FEATURE` label type is entirely controlled by the configuration value `target.features`. The following is a trimmed down version of an example `targets.json`:
```
bridadan: "The set of directories include by the `FEATURE`:" -> "The set of directories **included** by the `FEATURE`"

docs/tools/build_rules.md
```diff
@@ -0,0 +1,134 @@
...
+        "features": ["BLE"]
+    }
+}
+```
+
+In the above example, `mbed compile` will include files in directories named `FEATURE_BLE`, and not directories such as `FEATURE_STORAGE` or `FEATURE_CRYTOCELL310`.
+
+#### Component Directories
+
+The component labeled directories are used for software that implements some funtionality. They are within label directory primarily because we don't expect every program to use this software, and including this software in every build would needlessly increase build time.  The set of directories include by the `COMPONENT` label type is entirely controlled by the configuration value `target.components`. The following is a trimmed down version of an example `targets.json`:
```
bridadan: "The set of directories include by the `COMPONENT`" -> "The set of directories **included** by the `COMPONENT`"

AnotherButler:
@bridadan @TacoGrandeTX @MarceloSalazar Please do a final review and let me know whether all comments have been addressed.

docs/tools/build_rules.md
```diff
@@ -0,0 +1,133 @@
...
+        "features": ["BLE"]
+    }
+}
+```
+
+In the above example, `mbed compile` includes files in directories named `FEATURE_BLE`, and not directories such as `FEATURE_STORAGE` or `FEATURE_CRYTOCELL310`.
+
+#### Component directories
+
+The component labeled directories are used for software that implements funtionality. They are within label directory primarily because we don't expect every program to use this software, and including this software in every build would needlessly increase build time.  The configuration value `target.components` entirely controls the set of directories the `COMPONENT` label type includes. The following is a shortened version of an example `targets.json`:
```
bridadan: "They are within label directory primarily because" -> "They are within label **directories** primarily because"

TacoGrandeTX:
Yes - the new table showing file types is super!  It does have one typo:  "Toolchain's assebler"

AnotherButler:
@TacoGrandeTX Great catch. I've now fixed that.

PR #695, opened 2018-08-30
Minor QSPI contributing cleanup

kegilbert:
Remove reference to "future API" and separate PR requirements in the contributing doc now that QSPI is mainlined. Also cleaned up the intro slightly. 

@AnotherButler We seem to switch between `QSPI` and `QuadSPI` throughout the doc, should we stick to one or the other?

@maciejbocianski


maciejbocianski:
`QSPI` is a shortcut for `Quad SPI`

PR #699, opened 2018-08-31
Corrected overhead limitation

deepikabhavnani:
Overhead is added only in case of memory stats and not memory trace


AnotherButler:
ping @SenRamakri

PR #702, opened 2018-09-03
Create guidelines for the mbedtls_platform_context

RonEld:
1. Add guidelines for Applications to call the platform API
2. Add guidelines on how to port an alternative platform context.

### DIFF
--- a/docs/reference/api/security/TLS.md
+++ b/docs/reference/api/security/TLS.md
@@ -43,10 +43,11 @@ Then create a file named `mbed_app.json` at the root of your application with th
 
 ### Mbed TLS platform context
 
-Some hardware accelerators require initialization, regardless of specific cryptography engine. For this, `mbedtls_platform_setup()` and `mbedtls_platform_terminate()` was introduced.
+Some hardware accelerators require initialization, regardless of the specific cryptography engine. For this, we introduced `mbedtls_platform_setup()` and `mbedtls_platform_terminate()`.
 
-As shown in the [examples](#mbed-tls-examples), you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()`, to terminate the platform hardware driver. The platform context parameter for these function *should* be `NULL`.
+As the [examples](#mbed-tls-examples) show, you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()` to terminate the platform hardware driver. The platform context parameter for these function *should* be `NULL`.
 
+```
     int exit_code = MBEDTLS_EXIT_FAILURE;
 
     if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
@@ -58,6 +59,7 @@ As shown in the [examples](#mbed-tls-examples), you *must* call the `mbedtls_pla
 
     mbedtls_platform_teardown(NULL);
     return exit_code;
+```
 
 ### Other resources
 
--- a/docs/reference/api/security/TLS.md
+++ b/docs/reference/api/security/TLS.md
@@ -45,7 +45,7 @@ Then create a file named `mbed_app.json` at the root of your application with th
 
 Some hardware accelerators require initialization, regardless of the specific cryptography engine. For this, we introduced `mbedtls_platform_setup()` and `mbedtls_platform_terminate()`.
 
-As the [examples](#mbed-tls-examples) show, you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()` to terminate the platform hardware driver. The platform context parameter for these function *should* be `NULL`.
+As the [examples](#mbed-tls-examples) show, you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()` to terminate the platform hardware driver. We suggest you set the platform context parameter for these functions as `NULL`.
 
 ```
     int exit_code = MBEDTLS_EXIT_FAILURE;
--- a/docs/reference/api/security/TLS.md
+++ b/docs/reference/api/security/TLS.md
@@ -41,6 +41,24 @@ Then create a file named `mbed_app.json` at the root of your application with th
 }
 ```
 
+### Mbed TLS platform context
+
+Some hardware accelerators require initialization, regardless of specific cryptography engine. For this, `mbedtls_platform_setup()` and `mbedtls_platform_terminate()` was introduced.
+
+As shown in the [examples](#mbed-tls-examples), you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()`, to terminate the platform hardware driver. The platform context parameter for these function *should* be `NULL`.
+
+    int exit_code = MBEDTLS_EXIT_FAILURE;
+
+    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
+        mbedtls_printf("Platform initialization failed with error %d\r\n", exit_code);
+        return MBEDTLS_EXIT_FAILURE;
+    }
+
+    /* call Mbed TLS code here */
+
+    mbedtls_platform_teardown(NULL);
+    return exit_code;
+
 ### Other resources
 
 The [Mbed TLS website](https://tls.mbed.org) contains many other useful resources for developers, such as [developer documentation](https://tls.mbed.org/dev-corner), [knowledge base articles](https://tls.mbed.org/kb) and a [support forum](https://tls.mbed.org/discussions).
--- a/docs/reference/api/security/TLS.md
+++ b/docs/reference/api/security/TLS.md
@@ -45,20 +45,20 @@ Then create a file named `mbed_app.json` at the root of your application with th
 
 Some hardware accelerators require initialization, regardless of the specific cryptography engine. For this, we introduced `mbedtls_platform_setup()` and `mbedtls_platform_terminate()`.
 
-As the [examples](#mbed-tls-examples) show, you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()` to terminate the platform hardware driver. We suggest you set the platform context parameter for these functions as `NULL`.
+As the [examples](#mbed-tls-examples) show, you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()` to terminate the platform hardware driver. For readability reasons, we suggest you set the platform context parameter for these functions as `NULL`, as it is not being used in Mbed OS.
 
 ```
-    int exit_code = MBEDTLS_EXIT_FAILURE;
+    int ret = 0;
 
-    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
-        mbedtls_printf("Platform initialization failed with error %d\r\n", exit_code);
-        return MBEDTLS_EXIT_FAILURE;
+    if((ret = mbedtls_platform_setup(NULL)) != 0) {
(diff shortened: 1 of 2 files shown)


AnotherButler:
@RonEld Is this a requirement for the Mbed OS 5.10 release?

RonEld:
@AnotherButler I would say it is, as it describes the blocker bug fix https://github.com/ARMmbed/mbed-os/pull/7077 for OS 5.10

docs/reference/api/security/TLS.md
```diff
@@ -41,6 +41,26 @@ Then create a file named `mbed_app.json` at the root of your application with th
 }
 ```
 
+### Mbed TLS platform context
+
+Some hardware accelerators require initialization, regardless of the specific cryptography engine. For this, we introduced `mbedtls_platform_setup()` and `mbedtls_platform_terminate()`.
+
+As the [examples](#mbed-tls-examples) show, you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()` to terminate the platform hardware driver. The platform context parameter for these function *should* be `NULL`.
```
AnotherButler: Query: When we say "should", do we mean if everything's working correctly, the parameter is NULL? Or do we mean something else?

RonEld: I mean, that in Mbed OS, we actually ignore this parameter, so to avoid confusion, we recommend this parameter being NULL. However, having it a different value will not have any effect, except that it will make the code less underastood

AnotherButler:
@RonEld I've edited your PR. Please review my changes to make sure I didn't accidentally change the meaning of anything.

docs/reference/api/security/TLS.md
AnotherButler: Could we rephrase this as "We recommend you set the platform context parameter for these functions as `NULL`."?

RonEld:
@AnotherButler I have looked, and the meaning in hw-accelerated-crypto.md has changed

docs/reference/contributing/target/hw-accelerated-crypto.md
```diff
@@ -135,13 +135,7 @@ The default implementation of the modules are usually in the file `feature/mbedt
...
 ### Mbed TLS platform context
 
-Some hardware accelerators require initialization, regardless of specific cryptography engine. For this, `mbedtls_platform_setup()` and `mbedtls_platform_terminate()` was introduced.
-
-When your hardware accelerator driver requires initialization, you should do the following operations:
-
-1. Define `MBEDTLS_PLATFORM_SETUP_TEARDOWN_ALT` in `mbedtls_device.h`
-1. Implement `crypto_platform_setup(crypto_platform_ctx *ctx)` and `crypto_platform_terminate(crypto_platform_ctx *ctx)` that will initialize and terminate your hardware accelerator driver.
-1. Define `crypto_platform_ctx` in `crypto_platform.h` in a way that suits your implementation.
+Some hardware accelerators require initialization, regardless of specific cryptography engine. For instructions on how to do this, please see our [security page](/docs/development/apis/tls.html#mbed-tls-platform-context).
```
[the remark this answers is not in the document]
AnotherButler: Good catch. I missed that.

RonEld:
@AnotherButler Now that I think about it, do you think we should first ask a review from one of the dev team?

AnotherButler:
@RonEld I've added that section back and added some edits. Please re-review.

RonEld:
@k-stachowiak could you please review this, as you updated the relevant document in the Mbed TLS docs repo?
@AnotherButler I don't have additional comments. Thanks!

docs/reference/api/security/TLS.md
```diff
@@ -41,6 +41,26 @@ Then create a file named `mbed_app.json` at the root of your application with th
 }
 ```
 
+### Mbed TLS platform context
+
+Some hardware accelerators require initialization, regardless of the specific cryptography engine. For this, we introduced `mbedtls_platform_setup()` and `mbedtls_platform_terminate()`.
+
+As the [examples](#mbed-tls-examples) show, you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()` to terminate the platform hardware driver. We suggest you set the platform context parameter for these functions as `NULL`.
```
k-stachowiak: I think, that the advice about passing NULL to the functions may be a little confusing. I am not sure if it is possible to explain it briefly enough so that it suits this document, but I'd definitely try doing it. Otherwise I would consider omitting this advice and leave it for the detailed documentation.

docs/reference/api/security/TLS.md
```diff
@@ -41,6 +41,26 @@ Then create a file named `mbed_app.json` at the root of your application with th
...
+
+```
+    int exit_code = MBEDTLS_EXIT_FAILURE;
+
+    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
+        mbedtls_printf("Platform initialization failed with error %d\r\n", exit_code);
```
k-stachowiak: Since this is Mbed OS documentation, I wonder if `mbedtls_printf` is the most suitable. Maybe this line can be dropped, as it is not relevant to what is discussed here?

docs/reference/api/security/TLS.md
```diff
@@ -41,6 +41,26 @@ Then create a file named `mbed_app.json` at the root of your application with th
...
 ```
 
+### Mbed TLS platform context
+
+Some hardware accelerators require initialization, regardless of the specific cryptography engine. For this, we introduced `mbedtls_platform_setup()` and `mbedtls_platform_terminate()`.
+
+As the [examples](#mbed-tls-examples) show, you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()` to terminate the platform hardware driver. We suggest you set the platform context parameter for these functions as `NULL`.
+
+```
+    int exit_code = MBEDTLS_EXIT_FAILURE;
```
k-stachowiak: The exit code variable has a mixed meaning in the example: the return code for the program and for the Mbed TLS API. It has been demonstrated that Mbed TLS return codes exceed allowed exit codes for some platforms (e.g. POSIX), and therefore it may be advisable not to return them from Mbed TLS programs. I think, that this example can be expressed without the `exit_code` variable. Instead you could only declare `ret` to capture the Mbed TLS error code.

docs/reference/api/security/TLS.md
```diff
@@ -41,6 +41,26 @@ Then create a file named `mbed_app.json` at the root of your application with th
...
+
+    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
+        mbedtls_printf("Platform initialization failed with error %d\r\n", exit_code);
+        return MBEDTLS_EXIT_FAILURE;
+    }
+
+    /* call Mbed TLS code here */
+
+    mbedtls_platform_teardown(NULL);
+    return exit_code;
```
k-stachowiak: From just the code presented in the example (e.g. not taking the `/* call Mbed TLS code here */` part, one may conclude that this program returns the default `exit_code`'s value upon success, which has been set to `MBEDTLS_EXIT_FAILURE` earlier.

docs/reference/api/security/TLS.md
k-stachowiak: Since this is an Mbed OS related example, I am not sure, using `MBEDTLS_EXIT_FAILURE` and such is suitable.

docs/reference/api/security/TLS.md
RonEld: That detailed documentation. This should be the detailed documentation, but again, not too detailed, as we don't want to confuse.
What is confusing?

docs/reference/api/security/TLS.md
RonEld: Note that our Mbed OS examples use `mbedtls_printf`

docs/reference/api/security/TLS.md
RonEld: ok

RonEld: Note that this was taken from our examples, do you think we should change them?

docs/reference/api/security/TLS.md
RonEld: ok

docs/reference/api/security/TLS.md
RonEld: ok. Note that this is still an Mbed TLS documentation, on the Mbed OS.

RonEld:
@k-stachowiak Thank you for your comments!
I believe I have addressed them.

docs/reference/api/security/TLS.md
k-stachowiak: If I first saw such documentation I would have some questions in my mind, such as:

* If this function is so important (one *must* call it), why is its argument irrelevant?
* If I am to pass in NULL, why does the function take a parameter at all?
* What are the cases, in which we pass an actual object?
* If I want to know the details, where do I look? (OK this may be obviously the API reference, but we may put a link here, so that it is conveniently accessible)

To make myself clear - I think the way you put it is enough to use the API correctly, so I think it is correct itself. I am just trying to foresee issues, that new user may have.

docs/reference/api/security/TLS.md
k-stachowiak: I don't think the examples need to be changed regarding this. The `exit_code` is set correctly there, but it is done in places that you had to cut away for this example.

docs/reference/api/security/TLS.md
```diff
@@ -45,20 +45,20 @@ Then create a file named `mbed_app.json` at the root of your application with th
...
 ```
-    int exit_code = MBEDTLS_EXIT_FAILURE;
+    int ret = 0;
 
-    if((exit_code = mbedtls_platform_setup(NULL)) != 0) {
-        mbedtls_printf("Platform initialization failed with error %d\r\n", exit_code);
-        return MBEDTLS_EXIT_FAILURE;
+    if((ret = mbedtls_platform_setup(NULL)) != 0) {
+        mbedtls_printf("Platform initialization failed with error %d\r\n", ret);
+        return ret;
```
k-stachowiak: Returning `1` here would be safer, as, in general, we should not return Mbed TLS error codes from main (as they may exceed the valid range handled by an OS). I have not heard of it ever being an actual issue, apart from conforming to standards (such as POSIX) or not. You may consider changing it, but I don't find it necessary.

docs/reference/api/security/TLS.md
AnotherButler: Briefly here: https://os-doc-builder.test.mbed.com/docs/development/apis/tls.html

docs/reference/api/security/TLS.md
```diff
@@ -41,6 +41,26 @@ Then create a file named `mbed_app.json` at the root of your application with th
 }
 ```
 
+### Mbed TLS platform context
+
+Some hardware accelerators require initialization, regardless of the specific cryptography engine. For this, we introduced `mbedtls_platform_setup()` and `mbedtls_platform_terminate()`.
+
+As the [examples](#mbed-tls-examples) show, you *must* call the `mbedtls_platform_setup()` function before you call any Mbed TLS API. After using the Mbed TLS API, you *must* call `mbedtls_platform_teardown()` to terminate the platform hardware driver. For readability reasons, we suggest you set the platform context parameter for these functions as `NULL`, as it is not being used in Mbed OS.
```
sbutcher-arm: I don't think it makes sense to say:
> we suggest you set the platform context parameter for these functions as `NULL`

This parameter must ALWAYS be NULL, and must be ignored by Mbed OS - because that's what the BLE code has to do.

It would be better to say:

> Mbed OS makes no use of the platform context parameter, and within Mbed OS functions should always be passed a `NULL` parameter for the context.

sbutcher-arm:
@RonEld - I think the documentation is wrong and misleading concerning the parameter passed in `mbedtls_platform_setup()`, however the PR has been approved, so I guess we need another PR to correct it.

Can you respond to my review comment?

docs/reference/api/security/TLS.md
RonEld: Say that you MUST set the parameter to NULL implies that if it isn't NULL, then the function returns an error. That is not the case. The function simply ignores the parameter. This was don for backwards compatibility.
It is a suggestion to set to NULL, because it is ignored, and it is better readable that way, but it isn't mandatory

RonEld:
@sbutcher-arm I responded to your comment.
As mentioned, since the parameter is ignored, it's not a MUST to be NULL, only good practice

Issue #719, opened 2018-09-10
incorrect line number for example


SamiLappalainenArm:
1) go to url: https://os-doc-builder.test.mbed.com/docs/development/tools/adding-and-configuring-targets.html
2) download example .pdsc file via link "http://www.keil.com/pack/Keil.Kinetis_K20_DFP.pdsc is the PDSC that contains TEENSY_31 device (MK20DX256xxx7). "
3) open file with text (file content is .xml)
4) search string "MK20DX256xxx7"

configuration starts from row 204 but documentation states information starts at row 156.

AnotherButler:
Resolved by PR #729 

Nice catch @SamiLappalainenArm 👍

PR #729, opened 2018-09-11
Fix line number in mbed_targets.md

AnotherButler:
Fix incorrect line number.


AnotherButler:
This fixes issue #719

PR #737, opened 2018-09-12
FastModels - Add information about DS-5 and add to PATH

yennster:
This PR adds information about using the DS-5 installation to load Mbed OS images to the FVP platforms provided in DS-5. Also information on adding the DS-5 bin folder to the system PATH

### DIFF
--- a/docs/tools/fastmodels/fastmodels.md
+++ b/docs/tools/fastmodels/fastmodels.md
@@ -2,11 +2,13 @@
 
 Arm Fast Models are software models of Arm cores, subsystems, peripherals and so on. You can run Mbed OS on the software model instead of the hardware.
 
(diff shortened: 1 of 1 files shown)


jamesbeyond:
Hi @yennster @AnotherButler, Thanks for the updates.

**Here is what I think:**
Put the information of DS-5 in the first section would be confusing to some people, especially the people don't know the relations of DS-5 and Fast Models. 
DS-5 and Fast Models actually 2 separate products. We don't need DS-5 at all to run FastModels.
Install DS-5 is happened to be the most conveninet way for Arm internal people, but not necessary for the external people.

I don't want to give people the impression that they need purchases both Fast Model and DS-5 to use this feature.

In line 33:
>To run Mbed OS examples with Fast Models, you need to install the Fast Models product and set up the license.

This actually implicitly means that people need to put the model binary in the PATH

**Here are my suggestions**
1. Move the paragraph down and merge with Line 33
2. "DS-5 also provides fast models targets" can be  mentioned, but no need to put up the link for purchasing DS-5.
3. Putting the bin folder into the PATH can be mentioned as a reminder. they can ether put the DS-5 bin or Fast Model bin.

AnotherButler:
@jamesbeyond Do my changes resolve your concerns?

jamesbeyond:
Thanks @AnotherButler , that looks good!

PR #739, opened 2018-09-12
Update linker script templates to include stack

c1728p9:
Update the linker script templates to include the ability to adjust the boot stack stack size.

This depends on the code change here:
https://github.com/ARMmbed/mbed-os/pull/8039

### DIFF
--- a/docs/porting/target/bootstrap.md
+++ b/docs/porting/target/bootstrap.md
@@ -17,12 +17,17 @@ After adding the core files, the next step is to add linker scripts for Mbed OS.
 If you are updating your own linker script, you must:
 
 - Reserve space for the RAM vector table.
-- Define the start of the heap:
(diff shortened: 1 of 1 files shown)


c1728p9:
Hi @deepikabhavnani, I set the default stack size to 4K to match the value in targets.json.  The RTOS mbed_lib.json overrides this to 1K when present.

This linker script template is for both 2 and 5. Are you wanting this to be stated more explicitly? Can you elaborate on what you would like updated in bootstrap.md?

AnotherButler:
@deepikabhavnani Is this code OK to merge, or does @c1728p9 need to make further changes?

deepikabhavnani:
Needs changes

c1728p9:
Made changes and rebased to latest.

PR #742, opened 2018-09-18
Fixed test-config name

JuhPuur:
Need to change the test-config name following change in mbed OS PR https://github.com/ARMmbed/mbed-os/pull/7996


OPpuolitaival:
@JuhPuur  can be merged now?

JuhPuur:
Yes the code dependency was merged. @AnotherButler?

PR #744, opened 2018-09-20
Document network status changes

SeppoTakalo:
This clarifies responsibilities between application and interface on
network errors and reconnection logic.

@lauri-piikivi @yogpan01 @AriParkkila Please review

@AnotherButler Please hold the review until we have discussed the technical details. I need to open the PR for discussion as this crosses guite many team boundaries. Also, please don't move images until developers are finished with the review. I'll ping you later.


yogpan01:
@SeppoTakalo Looks like a good starting point. I already have a question, Is NetworkInterface connect() one time setup and user doesn't need to call it again if the state goes to "Connecting" state.
Are NetworkInterfaces self recovering or not? We need to add these kind of information in this document.
Also, explain to user , what they should be doing on their end when they receive these states on NetworkInterface.

SeppoTakalo:
I tried to clarify this in the document:

```
When network is lost, route is lost or any other cause limits the connectivity, interface may change 
its state back to `Connecting` or `Local connectivity`. Interface tries reconnecting until application 
chooses to call `NetworkInterface::disconnect()`. Depending on the network, this reconnection 
might have internal back off periods.
```


But is it not clear?

What I try to say that when you call `NetworkInterface::connect()` it will go into indefinite loop where it tries to connect as long as the device runs.

And we have bugs that many interfaces do not do that, therefore before fixing, I'll document the correct behaviour.

Only reason for calling `disconnect()` would be if you want to save battery power. Or something like that.

So to your question: Interfaces should be self-recovering.
But not all of them are, until we fix them.

SeppoTakalo:
@lauri-piikivi @yogpan01 Updated a little bit.

Lauri, can you review and comment?

yogpan01:
@SeppoTakalo Document looks fine. One thing I would like you to add is explanation about state like  
```
 NSAPI_STATUS_LOCAL_UP,            /*!< local IP address set */
 NSAPI_STATUS_GLOBAL_UP,           /*!< global IP address set */

```

We need to explain what user or application can do about NSAPI_STATUS_LOCAL_UP and NSAPI_STATUS_GLOBAL_UP.

We have use case where Cloud server would run inside the same network as Client ( on-prem solution) so in that case is it that Network interface will never receive GLOBAL_UP because there is no outside connectivity to configuration. We need to then follow local connectivity instead of Global one.

Can you clarify that in document?

SeppoTakalo:
In the network interface document:

image

In reality, Ethernet and WiFi devices issue LOCAL_UP and GLOBAL_UP at the same time.
They only receive one IP address from DHCP, and basically it is assumed that route is global.

yogpan01:
@SeppoTakalo Thanks for clarifying, Is this table already part of some document or should you add it to this document for clarity.

SeppoTakalo:
It is there, just load the second document `networkinterface.md` and click view.

Direct link:
https://github.com/ARMmbed/mbed-os-5-docs/blob/e766c27218c4afff4f6f753d8a251c12eb002e91/docs/api/networkinterfaces/networkinterface.md

SeppoTakalo:
First document is only about the API, how to attach the event handler.
Second document is the actual information what should be done in those event, and how interfaces behave.

On the handbook the second document is actually one layer up.

yogpan01 (approved):

lauri-piikivi (commented):
I think it looks good, and has my approval. Still a question, @SeppoTakalo @yogpan01 could we list the actions application should take with the cloud client? We now just say close sockets, but there is, hopefully, the action to pause the cloud client
Maybe just add link to the client pages regarding this behaviour?

docs/api/networkinterfaces/networkinterface.md
```diff
@@ -1,8 +1,8 @@
 ## Network interface overview
 
-A socket requires a NetworkInterface instance when opened to indicate which NetworkInterface the socket should be created on. The NetworkInterface provides a network stack that implements the underlying socket operations.
+A socket requires a NetworkInterface instance to indicate which NetworkInterface the socket should be created on. The NetworkInterface provides a network stack that implements the underlying socket operations.
```
AnotherButler: Query: Are we talking about the class every time we use NetworkInterface and the concept every time we use network interface? Please confirm.

AnotherButler:
@MelindaWeed I've done an initial pass on this. Could you please do another when you get the chance?

docs/api/networkinterfaces/networkinterface.md
SeppoTakalo: Yes.
When NetworkInterface is used, it means instance of the class. So maybe it should be `NetworkInterface` because it refers to API.

When "network interface" is instead used, we refer to actual physical interface.

SeppoTakalo:
@melwee01 I suppose this is pending for your review? Can you do it please.

AnotherButler:
It looks like a code snippet is failing. Could an engineer please check this?

SeppoTakalo:
Code snippets fixed.

AnotherButler:
@SeppoTakalo Do these changes apply to 5.10, as well as 5.11?

SeppoTakalo:
Yes. These changes apply to both versions.

PR #745, opened 2018-09-20
Add FileHandle/poll documentation

kjbracey:
Was asked to do `poll` documentation, but can't really do it without `FileHandle` documentation.

First attempt at a combined document. Examples will need extracting.


SeppoTakalo:
ping @geky and @bulislaw 

Please review.

AnotherButler:
ping @geky

deepikabhavnani (approved):
Looks good to me

AnotherButler:
Please create separate pages for FileHandle and Poll because they're separate APIs.

docs/reference/api/platform/FileHandle.md
```diff
@@ -0,0 +1,274 @@
+## FileHandle
+
+For general information on [files](file.html) and [filing systems](filesystem.html), see the respective documentation. This chapter covers the abstract API, with an emphasis on devices, as they're less straightforward.
```
geky: Small suggestion, feel free to tweak/reject:
```suggestion
For general information on [files](file.html) and [filing systems](filesystem.html), classes specifically for persistant storage, see the respective documentation. This chapter covers the abstract API, with an emphasis on devices, as they're less straightforward.
```

docs/reference/api/platform/FileHandle.md
```diff
@@ -0,0 +1,274 @@
...
+The `FileHandle` abstraction permits stream-handling code to be device-independent. For example the console input and output streams used for C's `stdin` and `stdout` can be retargeted to something other than the default serial port via the `FileHandle` API, and `ATCmdParser` and the PPP connection to lwIP work on abstract `FileHandle` pointers.
+
+Exactly which operations a `FileHandle` supports will depend on the underlying device, and will in turn restrict what applications it is suitable for. For example, a database application might require random-access and `seek`, but this may not be available on a limited filesystem, and certainly not on a stream device. Only a `File` on a full `FileSystem` such as `FATFileSystem` would generally implement the entire API. Specialised devices may have particular limitations or behavior, which they should document, and which would limit their general utility.
+
+### Relationship of FileHandle to other APIs
+
+Mbed `FileHandle`s can be used directly, but they are often manipulated via POSIX or C/C++ APIs. The layering is that stdio calls taking `FILE *stream` call the POSIX APIs taking `int fd`, which call methods on `FileHandle` objects.
+
```
geky: It may help to add a picture here. Maybe something like this?

``` cpp
C stdio API:                  FILE *fopen(const char *path, const char *mode);
                                      |
                                      v
POSIX API:                    int open(const char *path, int flags);
                                      |
                                      v
Mbed OS C++ FileHandle API:   int FileHandle::open(const char *path, int flags);
                                      |
                                      v
Mbed OS C++ File:             int File::open(FileSystem *fs, const char *path, int flags);
                                      |
                                      v
Mbed OS C++ FileSystem:       int FileSystem::file_open(mbed::fs_file_t *f, const char *path, int flags);
                                      |
                                      v
Mbed OS C++ FATFileSystem:    int FATFileSystem::file_open(mbed::fs_file_t *f, const char *path, int flags);
                                      |
                                      v
ChanFS FAT File System:       FRESULT f_open (FIL* fp, const TCHAR* path, BYTE mode);
```

Ok I may have gotten a bit carried away. Feel free to cut that off where it's relevant.

docs/reference/api/platform/FileHandle.md
```diff
@@ -0,0 +1,274 @@
...
+* The sigio may be issued from interrupt context. You cannot portably issue `read` or `write` calls directly from this callback, so you should queue an `Event` or wake a thread to perform the `read` or `write`.
+* The sigio callback is only guaranteed when a `FileHandle` _becomes_ readable or writable. So if you do not fully drain the input, or fully fill the output, no sigio may be generated. This is also important on start-up - don't wait for sigio before attempting to read/write for the first time, but only use it as a "try again" signal after seeing an `EAGAIN` error.
+* Spurious sigios are permitted - you can't assume data will be available after a sigio.
+* Given all the above, use of sigio normally implies use of non-blocking mode, or possibly `poll`. 
+
+Ordinary files do not generate sigio callbacks, because they are always readable and writable, so never become readable or writable.
+
+### Stream-derived FileHandles
+
+Note that `FileHandle` implementations derived from `Stream`, such as `Serial`, have various limitations:
```
geky: This comes out of nowhere

docs/reference/api/platform/FileHandle.md
```diff
@@ -0,0 +1,274 @@
...
+As per POSIX, ordinary files always indicate readable and writable via poll.
+
+
+### Event-driven I/O
+
+If using non-blocking I/O, you will probably want to be know when to next attempt a `read` or `write` if they indicate no data is available. `FileHandle::sigio` lets you attach a `Callback` which will be called whenever the `FileHandle` becomes readable or writable.
+
+Important notes on sigio:
+
+* The sigio may be issued from interrupt context. You cannot portably issue `read` or `write` calls directly from this callback, so you should queue an `Event` or wake a thread to perform the `read` or `write`.
```
geky: Can we link to event queue documentation?

docs/reference/api/platform/FileHandle.md
```diff
@@ -0,0 +1,274 @@
...
+
+The `FileHandle` abstraction permits stream-handling code to be device-independent. For example the console input and output streams used for C's `stdin` and `stdout` can be retargeted to something other than the default serial port via the `FileHandle` API, and `ATCmdParser` and the PPP connection to lwIP work on abstract `FileHandle` pointers.
+
+Exactly which operations a `FileHandle` supports will depend on the underlying device, and will in turn restrict what applications it is suitable for. For example, a database application might require random-access and `seek`, but this may not be available on a limited filesystem, and certainly not on a stream device. Only a `File` on a full `FileSystem` such as `FATFileSystem` would generally implement the entire API. Specialised devices may have particular limitations or behavior, which they should document, and which would limit their general utility.
```
geky: Shoud we note that if not supported, implementations should return ENOSYS?

docs/reference/api/platform/FileHandle.md
```diff
@@ -0,0 +1,274 @@
...
+            copy_some(fds[0].fh, fds[1].fh);
+        }
+    }
+}
+```
+
+### Related content
+
+- [File](file.html).
+- [FileSystem](filesystem.html).
```
geky: Maybe add UARTSerial here?

Eventually it would be nice to make a "implementation" list section similar to the blockdevices. Though we're probably not there yet with a single non-filesystem implementation :laughing:

geky (approved):
This looks great! Sorry it took so long before I could review it. Thanks for the writing.

Left some suggestions, though it looks good to me.

docs/reference/api/platform/FileHandle.md
geky: Whoops, that was a note to myself, please ignore

geky: This does kinda come out of no where. Should we describe the Stream class?
```suggestion
`Stream` is a legacy class that provides an abstract interface for streams similar to the `FileHandle` class. The difference is that the `Stream` API is built around the `getc` and `putc` set of functions, whereas `FileHandle` is build around `read` and `write`. This makes implementations simpler but limits what is possible with the API. Because of this, the FileHandle API is suggested API for new designs.

Note that `FileHandle` implementations derived from `Stream`, such as `Serial`, have various limitations:
```

kjbracey: True. Will add an introductory paragraph along those lines.

docs/reference/api/platform/FileHandle.md
kjbracey: Will note that "not supported" errors should be indicated - sometimes there's a more specific code, eg `ESPIPE` for `invalid seek`.

docs/reference/api/platform/FileHandle.md
kjbracey: Probably helps. Think we fall out of scope after `File` - a concrete implementation of `FileHandle`.

And `open` is a special case. Maybe stick to a core call.

docs/reference/api/platform/FileHandle.md
geky: Good point

AnotherButler:
What's the status of this?

kjbracey-arm:
Reworked incorporating @geky's feedback, and splitting into separate FileHandle and poll pages.

Notes:

* Not certain that the links to other documents are necessarily correct for position in tree.
* Is diagram acceptable size/colour/whatever?
* Examples need extracting.

AnotherButler:
Thanks for the hard work 👍 Please review my edits to make sure I didn't accidentally change the meaning of anything.

kjbracey-arm:
Gah, just pressed "Cancel review" to close the "review comment" box and deleted all my pending review comments.

docs/api/platform/Poll.md
```diff
@@ -2,49 +2,51 @@
...
+- `int mbed::poll(pollfh fhs[], unsigned nfhs, int timeout)`, the variant taking an array of `FileHandle *`.
 
-See the POSIX specification of `poll` for full details of the API. Mbed OS provides the `POLLIN`, `POLLOUT`, `POLLERR` and `POLLHUP` event flags, but implementation is dependent on device.
+See the POSIX specification of `poll` for full details of the API. Mbed OS provides the `POLLIN`, `POLLOUT`, `POLLERR` and `POLLHUP` event flags, but implementation depends on the device.
 
-Note that the `poll` is not affected by the blocking/non-blocking setting of the individual file handles. The wait behaviour is always determined by the `timeout` argument: 0 for immediate (non-blocking) return, -1 for wait forever, or positive for a limited wait.
+Note that the `poll` is not affected by the blocking or nonblocking setting of the individual file handles. The `timeout` argument always determines the wait behavior: 0 for immediate (nonblocking) return, -1 for wait forever or positive for a limited wait.
 
-As per POSIX, ordinary files always indicate readable and writable via poll.
+Ordinary POSIX files are readable and writable using poll.
```
kjbracey: I meant "As per the POSIX specification, `poll` always indicates that ordinary files (as opposed to devices) are readable and writable".

docs/api/platform/FileHandle.md
```diff
@@ -1,22 +1,22 @@
...
+The `FileHandle` abstraction represents an already-opened file or device, so it has no `open` method of its own - the opening may take the form of a call to another class that returns a `FileHandle`, such as `FileSystem::open`, or it may be implicit in the construction of an object such as `UARTSerial`.
 
-The `FileHandle` abstraction permits stream-handling code to be device-independent. For example the console input and output streams used for C's `stdin` and `stdout` can be retargeted to something other than the default serial port via the `FileHandle` API, and `ATCmdParser` and the PPP connection to lwIP work on abstract `FileHandle` pointers.
+The `FileHandle` abstraction permits stream-handling code to be device-independent. For example, you can use the the `FileHandle` API to retarget the console input and output streams used for C's `stdin` and `stdout` to something other than the default serial port. You can also retarget `ATCmdParser` and the PPP connection to lwIP work on abstract `FileHandle` pointers.
```
kjbracey: Last sentence doesn't want the "You can also retarget".

After than, not sure if it flows right. This is giving 3 examples of device-independence

* Console retargetting stuff
* ATCmdParser
* PPP connection to lwIP

Maybe rework to:

> The `FileHandle` abstraction permits stream-handling code to be device-independent, rather than tied to a specific device like a serial port. Examples of such code in Mbed OS are:
> * the console input and output streams (`stdin` and `stdout`) 
> * the `ATCmdParser` helper
> * the PPP connection to lwIP

docs/api/platform/FileHandle.md
```diff
@@ -1,22 +1,22 @@
...
 
 ### Relationship of FileHandle to other APIs
 
-Mbed `FileHandle`s can be used directly, but they are often manipulated via POSIX or C/C++ APIs. The layering is that stdio calls taking `FILE *stream` call the POSIX APIs taking `int fd`, which call methods on `FileHandle` objects.
+You can use `FileHandle` directly, but POSX or C/C++ APIs often manipulate it. Stdio calls taking `FILE *stream` call the POSIX APIs taking `int fd`, which call methods on `FileHandle` objects.
```
kjbracey: POSX->POSIX

How about "You can use a `FileHandle` directly, or you can use standard POSIX or C/C++ APIs to manipulate it".

docs/api/platform/FileHandle.md
```diff
@@ -25,40 +25,39 @@ API                            | C/C++                   | POSIX
...
+- `FILE *fdopen(int fd, const char *mode)` bind a POSIX file descriptor to a stdio FILE.
+- `FILE *fdopen(FileHandle *fh, const char *mode)` bind a FileHandle to a stdio FILE.
 
-There is currently no call to map from POSIX file descriptor to `FileHandle`; this may be added in future.
+The standard POSIX function `int fileno(FILE *stream)` may be available to map from `FILE` to file descriptor, depending on the toolchain and C library in use - it is not usable in fully portable Mbed OS code.
 
-The standard POSIX function `int fileno(FILE *stream)` may be available to map from `FILE` to file descriptor, depending on the toolchain and C library in use - it is not useable in fully-portable Mbed OS code.
-
-Given those limitations on mapping, if code needs to access the lower levels, a lower-level open call should be used, so the  lower-level handle is known, then that is bound to the higher level(s).
+Given those limitations on mapping, if code needs to access the lower levels, use a lower-level open call, so the lower-level handle is known. Then, that is bound to the higher level.
```
kjbracey: This paragraph is dangling a bit with the "currently no call to map from POSIX fd to FileHandle" paragraph gone.

Change the intro to "As it is not possible to map from higher levels to lower levels,".

Final sentence should be imperative to match previous. "Then bind that to the higher level."

docs/api/platform/FileHandle.md
```diff
@@ -67,73 +66,75 @@ This can be overridden by the target providing `mbed::mbed_target_override_conso
...
-* Given all the above, use of sigio normally implies use of non-blocking mode, or possibly `poll`. 
+- The sigio may be issued from interrupt context. You cannot portably issue `read` or `write` calls directly from this callback, so you should queue an [`Event`](event.html) or wake a thread to perform the `read` or `write`.
+- The sigio callback is only guaranteed when a `FileHandle` _becomes_ readable or writable. If you do not fully drain the input or fully fill the output, no sigio may be generated. This is also important on start-up - don't wait for sigio before attempting to read or write for the first time, but only use it as a "try again" signal after seeing an `EAGAIN` error.
+- Spurious sigios are permitted - you can't assume data will be available after a sigio.
+- Given all the above, use of sigio normally implies use of nonblocking mode or possibly `poll`. 
 
-Ordinary files do not generate sigio callbacks, because they are always readable and writable, so never become readable or writable.
+Ordinary files do not generate sigio callbacks because they are always readable and writable.
```
kjbracey, line 111: I thought the last comma clause was necessary to spell out the logic, but maybe it was going too far. Up to you.

* The sigio callback is only guaranteed when a `FileHandle` _becomes_ readable or writable
* Ordinary files are always readable and writable
* Therefore they never _become_ readable or writable
* Therefore they never generate a sigio callback.

kjbracey (commented):
Thanks for edits - good stuff. Just a couple of points that have ended up not quite right.

PR #747, opened 2018-09-26
Clarify usage of custom bootloader

MarceloSalazar:
During development of applications and usage of the custom bootloader feature, I've noticed the header of the bootloader was not being been included at compile time.

```
  Region bootloader: size 0x8000, offset 0x8000000
  Region header: size 0x70, offset 0x8010000   <------- was missing
  Region application: size 0x1efc00, offset 0x8010400
```

And the bootloader was not working ok.

Adding `"target.features_add": ["BOOTLOADER"]` seems to solve the problem.

cc @bogdanm @theotherjimmy @AnotherButler

### DIFF
--- a/docs/tutorials/bootloader.md
+++ b/docs/tutorials/bootloader.md
@@ -61,6 +61,7 @@ To create an application using a bootloader, you must first have created the boo
 "target_overrides": {
     ...
     "<TARGET_NAME>": {
+        "target.features_add": ["BOOTLOADER"],
         "target.bootloader_img": "bootloader/my_bootloader.bin"
     },
     ...
@@ -165,6 +166,7 @@ You can place the prebuilt binary bootloader in `mbed-os/feature/FEATURE_BOOTLOA
    ```
    "target_overrides": {
            "YOUR_TARGET": {
+               "target.features_add": ["BOOTLOADER"],
                "target.app_offset": "0x10400",
                "target.header_offset": "0x10000",
                "target.header_format": [
@@ -184,6 +186,7 @@ Alternatively, you can edit `mbed_app.json`, and override the target bootloader 
    ```
        "target_overrides": {
            "YOUR_TARGET": {
+               "target.features_add": ["BOOTLOADER"],
                "target.bootloader_img": "PATH_TO_BOOTLOADER/your_bootloader.bin"
                "target.app_offset": "0x10400",
                "target.header_offset": "0x10000",


theotherjimmy:
@MarceloSalazar The header region is not required by all bootloaders, so that's not an error.

theotherjimmy:
Furthermore, your change,
```
"target.features_add": ["BOOTLOADER"]
```
is not the solution to the problem. What happens when you have a different header format from the Mbed OS bootloader: This will fail to boot.

MarceloSalazar:
I may be doing something wrong, but FYI this is what I see:

- Config 1:  the bootloader fails to jump to the app: `[ERR ] Failed to jump to application!`

```
            "target.bootloader_img": "bootloader/mbed-bootloader-F429ZI.bin",
```

This is the output of the compile task:

```
    Using ROM regions bootloader, application in this build.
      Region bootloader: size 0x8000, offset 0x8000000
      Region application: size 0x1f8000, offset 0x8008000
```

- Config 2: the bootloader jumps to the app and works as expected

```
            "target.features_add"  : ["BOOTLOADER"],
            "target.bootloader_img": "bootloader/mbed-bootloader-F429ZI.bin",
```

This is the output of the compile task:

```
    Using ROM regions bootloader, header, application in this build.
      Region bootloader: size 0x8000, offset 0x8000000
      Region header: size 0x70, offset 0x8010000
      Region application: size 0x1efc00, offset 0x8010400
```

I'd expect the bootloader to contain all information needed about the header.

Maybe I need to add these [2 lines](https://github.com/ARMmbed/mbed-os/blob/master/features/FEATURE_BOOTLOADER/targets/TARGET_STM/TARGET_STM32F4/TARGET_STM32F429xI/TARGET_NUCLEO_F429ZI/mbed_lib.json#L5-L6) in my `mbed_app.json`

theotherjimmy:
@MarceloSalazar You're using a bootloader that's identical to the one configured by the library configuration enabled by `"target.features_add": ["BOOTLOADER"]`. That's why it works. 

I could very easily write a bootloader that would work without a header (this is much easier than writing one that works with a header). With such a bootloader (mbed-os-example-bootloader was one) enabling the features bootloader is incorrect and will cause boot to fail! Therefore enabling the features bootloader has no place in documentation about how to use a custom bootloader.

--------------------

Further, you have offered no explanation offered for why this is required. The only justification that you have given is that it worked for you. That is not sufficient evidence to justify that it will work for everyone.

theotherjimmy:
Those 2 lines will probably help, but you would still be missing the `"target.header_format"` configuration value.

theotherjimmy:
What you have encountered is that the feature BOOTLOADER does not enable an extremely simple, bare bones bootloader. Instead, it uses a bootloader that is designed to fit many use cases without modification. The bootloader also includes the appropriate configuration so that it can work in one step (enabling). In using this bootloader to test the "custom bootloader conifguration" (which is just managed bootloader mode, and is used by the feature BOOTLOADER), you have embarked on a journey to re-create the very configuration used by the feature BOOTLOADER.

Maybe you should use a simpler bootloader.

MarceloSalazar:
@theotherjimmy thanks for the detailed reasoning!

You're right, I didn't realize the tools were pulling the default configuration, hence why the app was working 'ok'.

FYI I'm building a custom bootloader based on this repo/config:
https://github.com/ARMmbed/mbed-bootloader/blob/master/mbed_app.json#L119-L128
I'm only changing the nvstore configuration because the default one doesn't suite the expected behavior (need to have this at the very last sector to facilitate binary programming).

I got confused on [this](https://github.com/ARMmbed/mbed-os-5-docs/blob/development/docs/tutorials/bootloader.md#creating-the-main-program) particular section, as `"target.bootloader_img"` was the only config shown and didn't see comments about `header_format`.

> Maybe you should use a simpler bootloader.

The bootloader that **most** people are interested in using right now is the [mbed-bootloader](https://github.com/ARMmbed/mbed-bootloader) to work with Pelion and do FW Updates...

I'm experimenting with different configuration parameters for `"target.app_offset"`, `"target.header_offset"` and ` "target.header_format"` but don't seem to have luck on getting something to work. I might be missing something really basic here.

I feel we have a gap on how to use the `mbed-bootloader` with Mbed OS 5.10 applications.

How did you manage to compile the binaries for Mbed OS that are available here?
https://github.com/ARMmbed/mbed-os/tree/master/features/FEATURE_BOOTLOADER/targets

@evedon @bogdanm is this something you can help with?
Do we need a `header_format` config?

According to [this](https://github.com/ARMmbed/mbed-os/blob/master/features/FEATURE_BOOTLOADER/targets/TARGET_STM/TARGET_STM32F4/TARGET_STM32F429xI/TARGET_NUCLEO_F429ZI/mbed_lib.json) file, it seems we don't.

However, according to the output of the build, this line `Region header: .....` is still missing, and the bootloader can't find the app at run time. Thanks!

theotherjimmy:
@MarceloSalazar You're looking at a tutorial that walks you through each step from "I have no bootloader" through "boot passthrough" and onward. Maybe you should look at the specification of what each configuration does https://os.mbed.com/docs/v5.10/tools/configuring-tools.html

> as "target.bootloader_img" was the only config shown and didn't see comments about header_format.

That's because it's the step where you take the bootloader created by the section above, and add it to an application project. The section above creates a bootloader that simply passes off control to the application it's booting. Such a bootloader does not need a header, so for the sake of simplicity and correctness, no header is configured.

>The bootloader that most people are interested in using right now is the mbed-bootloader to work with Pelion and do FW Updates...

Which is exactly why `target.features_add: BOOTLOADER` exists, and is not a justification for adding unrelated configuration into a tutorial about building a new bootloader using Mbed OS and it's tools. There is a section dedicated to used the mbed-bootloader, and it is directly below.

> I'm experimenting with different configuration parameters for "target.app_offset", "target.header_offset" and "target.header_format" but don't seem to have luck on getting something to work. I might be missing something really basic here.

You need `"target.header_format"` to generate a firmware header, and for `"target.header_offset"` to mean something.

>I feel we have a gap on how to use the mbed-bootloader with Mbed OS 5.10 applications.

The fact that https://os.mbed.com/docs/v5.10/tools/configuring-tools.html is not the same as the page you edited may be related

I'm not seeing this "gap".

>According to this file, it seems we don't.

That file is not the only mbed_lib.json included. See https://github.com/ARMmbed/mbed-os/blob/master/features/FEATURE_BOOTLOADER/mbed_lib.json

theotherjimmy:
> I got confused on this particular section, as "target.bootloader_img" was the only config shown and didn't see comments about header_format.

It's also worth noting that "header_format" is the entire content of the following section.

MarceloSalazar:
@theotherjimmy thanks again for the details and references. I had missed the configuring-tools doc.

I think the docs in Mbed OS docs contain lots of detailed information (which is great) but are agnostic of the bootloader.

I was looking for docs specifically for mbed-bootloader and Mbed OS 5.10 apps.
I've written a short porting guide to guide people to use it:
https://github.com/ARMmbed/simple-mbed-cloud-client-template-restricted/blob/def_network/README.md#enabling-firmware-updates

I'm closing this PR, as it's not correct and not the right place to add this info. Thanks!

theotherjimmy:
@MarceloSalazar Since you missed it, where were you looking? That way we can at least put a link there.

MarceloSalazar:
I just missed the references to the config doc, my bad.

There are a couple of issues, maybe for @AnotherButler to review?

1. The link to 'bootloader configuration' in this section is broken:
https://github.com/ARMmbed/mbed-os-5-docs/blob/development/docs/tutorials/bootloader.md#unmanaged-bootloader

2. I feel the 'bootloader configuration' doc belongs to the Reference->[Configuration](https://os.mbed.com/docs/v5.10/reference/configuration.html) instead of Tools->[Configuring Tools](https://os.mbed.com/docs/v5.10/tools/configuring-tools.html).
Would you agree?

theotherjimmy:
Agreed on both points @MarceloSalazar. @AnotherButler How do you feel about moving the bootloader config doc?

MarceloSalazar:
@AnotherButler can you please have a look at this and comment?

theotherjimmy:
@MarceloSalazar 1) Works in the `os.mbed.com` version of the page.
2) https://github.com/ARMmbed/mbed-os-5-docs/pull/773

PR #760, opened 2018-10-03
Fix flow-control tutorial link

kegilbert:
@AnotherButler Can you confirm the new link is correct? Should point here:
https://github.com/ARMmbed/mbed-os-5-docs/blob/development/docs/tutorials/using_apis/flow_control.md

### DIFF
--- a/docs/api/rtos/Thread.md
+++ b/docs/api/rtos/Thread.md
@@ -83,4 +83,5 @@ The Callback API provides a convenient way to pass arguments to spawned threads.
 
 ### Related content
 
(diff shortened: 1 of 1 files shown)


AnotherButler:
Actually, I think it's correct as is.

AnotherButler:
Our links don't point to the markdown names. They point to the unique heading ID (or h2 if there is no unique heading ID).

kegilbert:
I get an error from this page when selecting the flow control link at the bottom:
https://os.mbed.com/docs/v5.10/apis/thread.html

AnotherButler:
Aha. It's missing a slash in front. "/docs..."

PR #761, opened 2018-10-04
Add intro to example in SPI.md

AnotherButler:
Add context to example for clarity.

### DIFF
--- a/docs/api/drivers/SPI.md
+++ b/docs/api/drivers/SPI.md
@@ -29,4 +29,6 @@ The SPI master generates a clock to synchronously drive a serial bit stream slav
 
 ### SPI hello, world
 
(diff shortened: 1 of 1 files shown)


docs/api/drivers/SPI.md
```diff
@@ -29,4 +29,6 @@ The SPI master generates a clock to synchronously drive a serial bit stream slav
 
 ### SPI hello, world
 
+The following example uses the WHOAMI register. The WHOAMI register is slave dependent and will differ based on slave setup. In other words, it's just an example that you can write to a slave's register. In this example, Mbed OS acts as the SPI master, and DigitalOut acts as the slave chip select (cs).
```
ashok-rao: "..slave setup" -> There is no setup required. The WHOAMI register is just an ID register of the slave device. In this example, the SPI master is the MCU on which Mbed OS is running and the slave is an external SPI device connected to the D11, D12, D13 pins.. a general purpose digitalout pin - D0 is being used to drive the slave's CS line either high or low to enable / disable (i.e. turn on/turn off) the slave (to start / stop sending the requested data from the master)

AnotherButler:
@ashok-rao Does that change clarify?

PR #766, opened 2018-10-10
add SPI migration guide

ithinuel:
Add SPI HAL API migration guide

### DIFF
--- a/docs/porting/target/spi.md
+++ b/docs/porting/target/spi.md
@@ -102,7 +102,7 @@ A target must also define these elements:
 - `#define SPI_COUNT (xxxxxU)`.
 - The number of SPI peripherals available on the device. A good place for that macro is `PeripheralNames.h` next to the `SPIName` enumeration.
 
-<span class="notes">**Note:**You can find more details about the design choices in the [SPI design document](https://github.com/ARMmbed/mbed-os/blob/feature-hal-spec-spi/docs/design-documents/hal/0000-spi-overhaul.md).</span>
+<span class="notes">**Note:** You can find more details about the design choices in the [SPI design document](https://github.com/ARMmbed/mbed-os/blob/feature-hal-spec-spi/docs/design-documents/hal/0000-spi-overhaul.md).</span>
 
 ### Dependencies
 
--- a/docs/porting/target/spi.md
+++ b/docs/porting/target/spi.md
@@ -1,0 +1,133 @@
+<h2 id="spi-port">Serial Peripheral Interface (SPI)</h2>
+
+The **Serial Peripheral Interface** allows you to send or receive a data stream over a synchronous serial interface made of 3 to 4 lines.
+
+- MISO: Master In Slave Out.
+- MOSI: Master Out Slave In.
+- MCLK: Clock.
+- SS: Slave Select.
+
+A typical use case of this interface is with SDCard, memory blocks and DACs or ADCs.
+
+This highly configurable interface has a wide range of elements you can adjust:
+
+- Frame length.
+- Clocks polarity and phase.
+
+<span class="warnings">**Warning:** We are introducing the SPI API in an upcoming release of Mbed OS. This page documents code that exists on a feature branch of Mbed OS. You can find details on how it may affect you in the [implementing the SPI API](#implementing-the-spi-api) section.
+
+### Assumptions
+
+#### Defined Behaviours
+
+- `spi_get_module()` returns the `SPIName` unique identifier to the peripheral associated to this SPI channel.
+- `spi_get_capabilities()` fills the given `spi_capabilities_t` instance
+- `spi_get_capabilities()` should consider the `ssel` pin when evaluation the `support_slave_mode` capability.  
+  If the given `ssel` pin cannot be managed by hardware in slave mode, `support_slave_mode` should be false.
+- At least a symbol width of 8bit must be supported.
+- The supported frequency range must include the range [0.2..2] MHz.
+- The shortest part of the duty cycle must not be shorter than 50% of the expected period.
+- `spi_init()` initializes the pins leaving the configuration registers unchanged.
+- `spi_init()` if `is_slave` is false:
+    - if `ssel` is `NC` the hal implementation ignores this pin.
+    - if `ssel` is not `NC` then the hal implementation owns the pin and its management.
+- When managed by the hal implementation, `ssel` is always considered active low.
+- When the hardware supports the half-duplex (3-wire) mode, if `miso` (exclusive) or `mosi` is missing in any function that expects pins, the bus is assumed to be half-duplex.
+- `spi_free()` resets the pins to their default state.
+- `spi_free()` disables the peripheral clock.
+- `spi_format()` sets :
+  - the number of bits per symbol
+  - the mode :
+    0. Clock idle state is *low*, data are sampled when the clock becomes *active* (polarity = 0, phase = 0)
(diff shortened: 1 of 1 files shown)


docs/porting/target/spi.md
```diff
@@ -0,0 +1,133 @@
+<h2 id="spi-port">Serial Peripheral Interface (SPI)</h2>
```
ashok-rao: @ithinuel : This is a HAL API doc & not a user facing API doc, correct? If so, could you please add a line in the introduction that this is only for the HAL..thanks.

docs/porting/target/spi.md
```diff
@@ -0,0 +1,133 @@
...
+- `spi_init()` if `is_slave` is false:
+    - if `ssel` is `NC` the hal implementation ignores this pin.
+    - if `ssel` is not `NC` then the hal implementation owns the pin and its management.
+- When managed by the hal implementation, `ssel` is always considered active low.
+- When the hardware supports the half-duplex (3-wire) mode, if `miso` (exclusive) or `mosi` is missing in any function that expects pins, the bus is assumed to be half-duplex.
+- `spi_free()` resets the pins to their default state.
+- `spi_free()` disables the peripheral clock.
+- `spi_format()` sets :
+  - the number of bits per symbol
+  - the mode :
```
ashok-rao: @AnotherButler : Seems to be a md formatting issue here.. mode 0 isn't being treated as a bullet point..

docs/porting/target/spi.md
```diff
@@ -0,0 +1,133 @@
...
+
+<span class="warnings">**Warning:** We are introducing the SPI API in an upcoming release of Mbed OS. This page documents code that exists on a feature branch of Mbed OS. You can find details on how it may affect you in the [implementing the SPI API](#implementing-the-spi-api) section.
+
+### Assumptions
+
+#### Defined Behaviours
+
+- `spi_get_module()` returns the `SPIName` unique identifier to the peripheral associated to this SPI channel.
+- `spi_get_capabilities()` fills the given `spi_capabilities_t` instance
+- `spi_get_capabilities()` should consider the `ssel` pin when evaluation the `support_slave_mode` capability.
```
ashok-rao: SSEL pin may have different names for each target (SSEL, NCS etc).. how will this be handled?

docs/porting/target/spi.md
```diff
@@ -0,0 +1,133 @@
...
+  - the mode :
+    0. Clock idle state is *low*, data are sampled when the clock becomes *active* (polarity = 0, phase = 0)
+    1. Clock idle state is *low*, data are sampled when the clock becomes *inactive* (polarity = 0, phase = 1)
+    2. Clock idle state is *high*, data are sampled when the clock becomes *active* (polarity = 1, phase = 0)
+    3. Clock idle state is *high*, data are sampled when the clock becomes *inactive* (polarity = 1, phase = 1)
+  - the bit ordering (lsb/msb first).
+- `spi_format()` updates the configuration of the peripheral except the baud rate generator.
+- `spi_frequency()` sets the frequency to use during the transfer.
+- `spi_frequency()` returns the actual frequency that will be used.
+- `spi_frequency()` updates the baud rate generator leaving other configurations unchanged.
```
ashok-rao: Why format() should leave baud rate generator un-touched? And the frequency() should only configure the baud rate generator? Can we have format() setting everything and frequency() change "only" the baud rate gen along with the other 2 functions above?

docs/porting/target/spi.md
```diff
@@ -0,0 +1,133 @@
...
+- Passing an invalid pointer as `handler` to `spi_transfer_async`.
+- Calling `spi_transfer_async_abort()` while no async transfer is being processed (no transfer or a synchronous transfer).
+- In half-duplex mode, any mechanism (if any is present) to detect or prevent collision is implementation defined.
+
+#### Other requirements
+
+A target must also define these elements:
+
+- `#define SPI_COUNT (xxxxxU)`.
+  The number of SPI peripheral available on the device.
```
ashok-rao: Might be helpful to provide the filename where this should be added..

docs/porting/target/spi.md
ithinuel: The actual name of the value does not matter much as what's considered by the `spi_get_capabilities` is the argument that's passed to it.

docs/porting/target/spi.md
ithinuel: I have declared in `PeripheralNames.h` in the reference implementation but is it THE best place ?
I looked sensible to me as the driver includes `hal/spi_api.h` that in turn includes `PeripheralNames.h`.

What do you think about it ?

docs/porting/target/spi.md
ithinuel: This is in the target porting guide. All chapter that live here relate to the hal layer.
The other guides don't have any warning about that.
Should this be added to all other guides too ?

docs/porting/target/spi.md
ithinuel: fixed in next commit.

docs/porting/target/spi.md
ithinuel: Because the responsibility of `spi_format` is only to configure the format of the frames. Changing the baudrate generator in `spi_format` could be considered as a side effect as it overlaps with the dedicated method `spi_frequency`.

The choice of having two functions was made during the [RFC process](https://github.com/ARMmbed/mbed-os/pull/7671) because the format is very unlikely to change during the lifetime of a spi instance while the frequency is likely to be raised after the first few exchanges with the slave peripheral.

docs/porting/target/spi.md
ashok-rao: Thanks @ithinuel . In that case, the warning is not required. Thank you.

docs/porting/target/spi.md
ashok-rao: @ithinuel : Yes, I was referring to ```PeripheralNames.h``` . Can you maybe add a note in this line to add the number of peripherals there?

ashok-rao (approved):
Thanks @ithinuel .. LGTM!

MarceloSalazar (approved):

donatieng (approved):

AnotherButler:
@c1728p9 Is this ready for my review?

c1728p9 (approved):

c1728p9:
@AnotherButler yeah, this is ready

AnotherButler:
Query: It looks like the Doxygen links don't work - Does that mean this is the first time we're using this feature branch? Do I need to create it?

AnotherButler:
ping @c1728p9

c1728p9:
@AnotherButler I think it is. Yes, please create it

docs/porting/target/spi.md
```diff
@@ -0,0 +1,130 @@
...
+
+The Mbed OS HAL provides a set of conformance tests for SPI. You can use these tests to validate the correctness of your implementation. To run the SPI HAL tests, use the following command:
+
+```
+mbed test -t <toolchain> -m <target> -n "tests-mbed_hal-spi*"
+```
+
+You can read more about the test cases:
+
+[![View code](https://www.mbed.com/embed/?type=library)](http://os.mbed.com/docs/development/feature-hal-spec-spi-doxy/group__hal__spi__tests.html)
```
AnotherButler, line 130: I've generated the branch, and the SPI class link earlier in the doc works. However, this one doesn't. Could you please look into it?

PR #767, opened 2018-10-11
improving can documentation

aashishc1988:
Improving CAN documentation. Pointing it to new sha1 for the improved example.


AnotherButler:
That link doesn't work for me. Do we need to wait for example changes to be merged?

aashishc1988:
I think so. First the example would have to be merged. :(

PR #769, opened 2018-10-11
Mutex doc

aashishc1988:
Correct mutex documentation according to latest mbed os version

### DIFF
--- a/docs/api/rtos/Mutex.md
+++ b/docs/api/rtos/Mutex.md
@@ -4,7 +4,7 @@
 
 A Mutex is used to synchronize the execution of threads, for example to protect the access to a shared resource.
 
-<span class="notes"> **Note:** The `Mutex` methods cannot be called from interrupt service routines (ISR). In the current version of Mbed OS, if you attempt to use a mutex from within an ISR, it will treat that as a fatal system error, and you will see an error like this: </span>
+The `Mutex` methods cannot be called from interrupt service routines (ISR). In the current version of Mbed OS, if you attempt to use a mutex from within an ISR, it will treat that as a fatal system error, and you will see an error like this:
 
 ```C
 ++ MbedOS Error Info ++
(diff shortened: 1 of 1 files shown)


aashishc1988:
I dont know why it combined my CAN changes (for which I already filed PR) with this PR. Looking into it

AnotherButler:
Thanks for the PR 👍

docs/api/rtos/Mutex.md
```diff
@@ -1,10 +1,24 @@
...
+### Mutex class hierarchy
+
+<span class="images">![](https://os-doc-builder.test.mbed.com/docs/development/mbed-os-api-doxy/classrtos_1_1_mutex.png)</span>
 
 A Mutex is used to synchronize the execution of threads, for example to protect the access to a shared resource.
 
-<span class="notes"> **Note:** The `Mutex` methods cannot be called from interrupt service routines (ISR). In the current version of Mbed OS, if you attempt to use a mutex from within an ISR, nothing happens; attempts to lock a mutex succeed immediately, regardless of whether the lock is actually free. In other words, if you acquire a mutex lock in an ISR, you can break the thread safety mechanisms and introduce race-conditions into an otherwise safe piece of code. Future versions of Mbed OS will provide warnings and ultimately prevent this from happening. </span>
+<span class="notes"> **Note:** The `Mutex` methods cannot be called from interrupt service routines (ISR). In the current version of Mbed OS, if you attempt to use a mutex from within an ISR, you see an error like this: </span>
```
deepikabhavnani: Can we change
`if you attempt to use a mutex from within an ISR, you see an error like this:`

to

`if you attempt to use a mutex from within an ISR, it will treat it as fatal system error and you see an error like this`

deepikabhavnani: Also we can add if synchronization is required in ISR, consider using semaphores

aashishc1988: Both good points. @AnotherButler can we hold off the merge of this PR while I work on these changes.?

AnotherButler: "if you attempt to use a mutex from within an ISR, the ISR treats the mutex as a fatal system error, and you see an error like this:"?

PR #771, opened 2018-10-11
Updating docs with new struct members

SenRamakri:
Update memory statics struct with overhead_size info.


bridadan:
The docs website does not currently have a Doxygen-style function or struct reference like on most of the pages. Would it be possible to add these?

deepikabhavnani:
Additional references like this are present only in tutorials and not API docs

PR #780, opened 2018-10-12
Minor doc change

aashishc1988:
Minor Doc change


AnotherButler:
Let's keep the current template in place.

PR #784, opened 2018-10-12
Update PortIn.md

TacoGrandeTX:
- Explain that this defines which pins are to be used as inputs and the port name is device specific
- Add references to PortNames.h, PinNames.h and PortOut
- Explain the default mask value


AnotherButler:
ping @SenRamakri

AnotherButler:
ping @SenRamakri

PR #788, opened 2018-10-13
Add validation for code snippets

geky:
Added a little script in travis to compile/validate any c/c++/json code snippets.

There were a few hickups with target specific blobs, so I also parse out an optional target to use for compilation (defaults to K64F). Or, you can simply put either NOCI or TODO to opt out of validation.

cc @iriark01, @AnotherButler, @SenRamakri, @kegilbert 
replaces https://github.com/ARMmbed/mbed-os-5-docs/pull/246

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,106 @@
+# Language and cache settings
+language: python
+python: 2.7
+cache:
+  pip: true
+  directories:
+    - $HOME/.cache/apt
+
+# Common install step
+install:
+  # Make sure pipefail
+  - set -o pipefail
+  # Set up apt to cache
+  - mkdir -p $HOME/.cache/apt/partial
+  - sudo rm -rf /var/cache/apt/archives
+  - sudo ln -s $HOME/.cache/apt /var/cache/apt/archives
+  # Set up ppa to make sure arm-none-eabi-gcc is the correct version
+  - sudo add-apt-repository -y ppa:team-gcc-arm-embedded/ppa
+  # Fix for "The following signatures were invalid: KEYEXPIRED 1515625755" failed". See https://github.com/travis-ci/travis-ci/issues/9037
+  - sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 0C49F3730359A14518585931BC711F9BA15703C6
+  # Loop until update succeeds (timeouts can occur)
+  - travis_retry $(! sudo apt-get update 2>&1 | grep Failed)
+  # Finally install gcc-arm-embedded
+  - sudo apt-get install gcc-arm-embedded
+  # Get mbed-os
+  - git clone https://github.com/armmbed/mbed-os.git
+  # Install python dependencies
+  - pip install -r mbed-os/requirements.txt
+
+# CI matrix
+jobs:
+  include:
+    # native testing
+    - stage: test
+      env:
+        - NAME=code-snippets
+      script:
+        # Run script to validate code snippets
+        - ./check_tools/find_bad_code_snippets.sh
+        # Grep to find remaining TODOs
+        - |
+            TODO_COUNT=0
+            for f in $(find -name mbed-os -prune -o -name '*.md' -print)
+            do
+                for l in $(sed -n '/```.*TODO/I=' $f)
+                do
+                    echo "TODO in $f line $l"
+                    TODO_COUNT=$(expr $TODO_COUNT + 1)
+                done
+            done
+            echo "Total number of TODOs: $TODO_COUNT"
+        # Update status with number of found TODOs if we passed testing
+        - |
+            if [ "$TRAVIS_TEST_RESULT" -eq 0 ]
+            then
+                CURR=$TODO_COUNT
+                PREV=$(curl -u "$MBED_BOT" https://api.github.com/repos/$TRAVIS_REPO_SLUG/status/master \
+                    | jq -re "select(.sha != \"$TRAVIS_COMMIT\")
+                        | .statuses[] | select(.context == \"$STAGE/$NAME\").description
+                        | capture(\"(?<count>[0-9]+) TODOs\").count" \
+                    || echo 0)
+                STATUS="Passed, ${CURR} TODOs"
+                if [ "$PREV" -ne 0 ]
+                then
+                    STATUS="$STATUS ($(python -c "print '%+d' % ($CURR-$PREV)") TODOs)"
+                fi
+            fi
+
+# Manage statuses
+before_install:
+  - |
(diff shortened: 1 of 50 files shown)


iriark01:
I have no idea what happened here, but I'm sure it's awesome.

AnotherButler:
@geky Is this ready for rebasing and merging, or are we waiting on access from @iriark01 ?

iriark01:
I think I did that last night. Comic sans, etc.

check_tools/find_bad_code_snippets.sh
```diff
@@ -0,0 +1,58 @@
...
+    fi
+}
+     
+for f in $(find -name mbed-os -prune -o -name '*.md' -print)
+do
+    # compile c++ snippets
+    for l in $(sed -n '/``` *\(cpp\|c++\)\($\| \+\)/I=' $f)
+    do
+        echo "Validating cpp $f:$l ..."
+        echo "#include \"mbed.h\"" > main.cpp
```
cmonr: Haha, I thought we were suppose to _not_ be doing this.

check_tools/find_bad_code_snippets.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+set -eo pipefail
+
+mkdir -p BUILD
+echo '*' > BUILD/.mbedignore
+
+# simple fake main for snippets without a main function
+cat > fake_main.c <<MAIN
+#include "mbed_toolchain.h"
```
cmonr: Is the fake_main only used to pull in the proper main.{c,cpp} file?

geky:
Haha, and we're already out of date. @iriark01 or someone with admin access, would you be able to add @mbed-bot to the users with push access? The github token in use doesn't actually allow push access, but push access is needed for status access. (sorry about all the hoops to make this work)

.travis.yml
```diff
@@ -0,0 +1,106 @@
...
+    # native testing
+    - stage: test
+      env:
+        - NAME=code-snippets
+      script:
+        # Run script to validate code snippets
+        - ./check_tools/find_bad_code_snippets.sh
+        # Grep to find remaining TODOs
+        - |
+            TODO_COUNT=0
```
0xc0170: What this TODO means here? Is it "NOT_COMPILE" like ? I would expect not having any todo on the documentation we publish but rather have the code that should compile if I copy-paste it or code that is to illustrate - stated there and defined as "not ready for compilation, needs edits"

check_tools/find_bad_code_snippets.sh
geky: Well the examples assume mbed is present don't they?

check_tools/find_bad_code_snippets.sh
geky: The script creates a main.cpp or main.c as necessary. fake_main just provides a weak main that is a noop which gets removed at linking time if the example already contains a main.

This is a way to get examples with only class declarations to compile. Otherwise the linker would complain about a missing main function. There may be a different way to do this with compiler flags, but I was lazy.

.travis.yml
geky: Ah I didn't explain it that well did I. @SenRamakri asked to present this next demo day, so I was going to go into more explanation there.

Basically there are 3 types of code snippets now:
- ` ``` cpp `
  By default code snippets should compile. You can opt-out with the other tags, but if you're just being lazy @iriark01 or @AnotherButler may get mad.

- ` ``` cpp NOCI `
  Code snippets that are not expected to compile. This may be snippets with only a single statement or an example of broken code (also makes @iriark01+@AnotherButler mad). If the snippets more than a handful of statements it should be wrapped in a function so that it can be compiled.

- ` ``` cpp TODO `
  Code snippets that should compile but don't. This was added because we currently have a lot of examples that should compile but don't. I tried to fix them where I could, but when I referenced the mbed docs the examples were broken there too!

  From talking with @AnotherButler, I believe the current plan is to take code snippets with TODO and create tasks for each of them for the owner of each technology to clean up. A lot of them should be simple (I think the USB examples just need the correct headers?).

You can find the TODO list in Travis log, and the bot should be responding with the number of TODOs so that the doc writers can see any changes (similar to the astyle reporting). But that's currently waiting on adding @mbed-bot to the users with push access. I was going to copy and paste that TODO list here once it's ready to go.

iriark01:
@geky I think your bot is alive

geky:
@iriark01 it is alive but has no mouth, a truly torturous existence. The push access lets it update statuses (it doesn't actually have permission to push at all). It will end up looking like it does on Mbed OS, where it has the same permissions:
image

iriark01:
Did I do it wrong?

geky:
@iriark01 ah! my bad! I misunderstood. Yep it's alive! (and broken, need to fix)

.travis.yml
geky: > we wont run any validation

That's what it's saying isn't it? What other validation are we running. IMO WONT_FIX isn't right since it shouldn't be broken.

---

Yep, to be clear the goal is to eventually remove all TODOs.

Maintaining a separate branch for TODO cleanup is not going to be tractable. This affects the entire handbook and the handbook moves too fast. The PR already has conflicts despite being up for only a day.

This is the same strategy @0xc0170 is applying with the astyle fixes in mbed-os, and it's the best approach I can think of without disrupted everyone who is currently working on handbook changes.

SenRamakri: @geky - We don't have this script to be gating now although we should start running(let it fail for now) so everyone knows we have failing snippets. Actually, if we let the script pass with all TODOs now it becomes slower to fix them.

geky: This has come up before, but the problem is that all PRs will have red Xs for thing the PR owner can't fix. This leads to users simply ignoring the CI results.

check_tools/find_bad_code_snippets.sh
kegilbert: I think for the purpose of just seeing if they compile its fine and much better than adding a bunch of manual include paths to individual snippets.

.travis.yml
kegilbert: I personally prefer the TODO method. Like @geky said if we push this in without that feature and let the multiple dozens of broken code just fail people will likely tend to disregard the CI results completely. Having a counter for remaining TODOs and having action items in Jira/Github to fix each TODO with a corresponding owner makes more sense to me.

kegilbert:
Re-re-rebase

cmonr:
Needs a Re-re-re-rebase

cmonr:
Or fewer open PRs ;)

AnotherButler:
Sorry - I'm finished now.

PR #790, opened 2018-10-15
Link to build rules documentation

theotherjimmy:
### Description

As we have had several people stumble over finding this documentation,
I'm linking to it from everywhere we mentoin `--source`, as the build 
rules are the primary affect of the aforemntioned switch. 

### Pull request type

    [x] Fix
    [ ] Refactor
    [ ] Target update
    [ ] Functionality change
    [ ] Breaking change


iriark01:
Chucking this in, although I think the content isn't there yet because I haven't merged 789. Working in order is for people who can count.

PR #792, opened 2018-10-15
Kernel documentation updated with new additions to Kernel namespace

SenRamakri:
Kernel documentation updated with new additions to Kernel namespace:

A new example has been created as well: https://github.com/ARMmbed/mbed-os-example-kernel-hooks


SenRamakri:
@kjbracey-arm - Good catch, Ill fix this as well.

PR #800, opened 2018-10-22
Update updating_firmware.md

mariannecrowder:
- L13: this section is not limited to developer workflow - connect certificate could be development or production. 
- L16: "You can also do this with..." confused me - do what? Importing the client is not called out explicitly in the Quick Connect guide - we do it as part of importing the example application. 
- L23: Certificates and manifest should all be created offline for production use cases. The manifest is not included in the image - it's a separate artefact. 
L48: "Deployment" vs "production"?


iriark01:
This is confusing me, so I'm going to merge the edit and then review the comments on my local copy

PR #809, opened 2018-10-31
For MacOS install remove 'sudo' and add "--user"

j3hill:
It is not recommended to use 'sudo' when installing MBED CLI components on a MacOS system. It is better to install using the "--user" flag.


theotherjimmy:
@j3hill I recall having to modify the user's `PATH` environment variable. Do you plan to document that as part of this PR or as another PR?

j3hill:
@theotherjimmy That is true. The '--user' flag places the new MBED CLI Python install in the /Users/{user}/Library/Python/2.7/bin folder. I added that to the path in the /.bash_profile using Nano. Is Nano the best way to do that, and should I just add a few lines of instruction on that? If that is the best way to do this, I can add this to this PR.

PR #810, opened 2018-10-31
Reorganize Using APIs page

kegilbert:
Move from comma separate list of tutorials to categories. Will be expanding this list with the new tutorials. Open to suggestions on whitespace/component layout. Moved connectivity up as we get requests to highlight the networking components of Mbed OS more as an IoT platform.

@AnotherButler


kegilbert:
Test site is fine, this page is going to be reworked further anyway before 5.11 goes live.

PR #812, opened 2018-11-02
Move mbedignore description into build rules

theotherjimmy:
These rules apply to more than just Mbed CLI, they also apply to 
Mbed Studio and the online compiler. As such, they should be part
of the generic project structure section


AnotherButler:
I'm going to close this one because it's trying to make about 5,000 commits and I think is from when GH was messing up earlier.

PR #814, opened 2018-11-05
CLA removal

0xc0170:
Related to https://github.com/ARMmbed/mbed-os/pull/8641 (defines inbound/outbound license). PR 8641 should go in first, and this one right after.


0xc0170:
Dependency was updated, this is ready for integration

PR #817, opened 2018-11-06
Add Reference API page for Trusted Storage

danny4478:


danny4478:
CC @alzix

AnotherButler:
So is this a user API or a porting API?

docs/api/security/trusted_storage.md
```diff
@@ -0,0 +1,27 @@
+<h2 id="blockdevice-port">Trusted Storage Reference API</h2>
```
alzix: this line probably copied from elsewhere and is not rellevant

docs/api/security/trusted_storage.md
```diff
@@ -0,0 +1,27 @@
...
+
+## PSA Internal Trusted Storage
+
+The PSA Internal Trusted Storage is a PSA Root of Trust Service which provides secure storage of small data items, for example, cryptographic keys.
```
alzix:
```suggestion
The PSA Internal Trusted Storage is a PSA Root of Trust Service which provides key value secure storage in internal flash for small data items. The main use cases for PSA Internal Trusted Storage APIs are:
- provide storage solution for persistent keys in PSA Crypto Service.
- storing rollback protection tokens for external application secure storage.
- storing persistent context in PSA Root of Trust services (i.e. epoch, device unique key. e.t.c.)
```

docs/api/security/trusted_storage.md
```diff
@@ -0,0 +1,27 @@
...
+It stores data by a uid, an identifier for the data.
```
alzix:
```suggestion
 PSA Internal Trusted Storage stores data by a uid, a unique identifier for the data.
```

docs/api/security/trusted_storage.md
dreemkiller: The term "secure" here is ambiguous and unnecessary. 
"which provides key value secure storage in internal flash for small data items"->"which provides key value storage in internal flash for small data items. It provides confidentiality and integrity by preventing non-PSA Root of Trust software from reading from or writing to the storage directly."

alzix: the later will be true on PSA platforms only - i think it should be mentioned explicitely as we plan to provide these APIs on non-PSA platforms (like k64f) for compatibility reasons

dreemkiller: good point.

alzix:
@AnotherButler,

> So is this a user API or a porting API?

It is user facing APIs, although we discourage direct usage of this APIs.

docs/api/security/trusted_storage.md
danny4478: It is the page headline. It says this is the Reference API page.
The section explains about the ITS and has its own headline.
I will make a small change here

docs/api/security/trusted_storage.md
danny4478: Guys, I am applying Alex's suggestion (almost) as is.
Please review one more time to see if acceptable.

docs/api/security/trusted_storage.md
danny4478: Done

dreemkiller:
If we are discouraging direct usage of this API, at the least we should note that in this documentation. Alternatively, we could not document this and refer users to the PSA API specification. Not sure what's best here.
Beyond that, it looks good to me.

danny4478:
> If we are discouraging direct usage of this API, at the least we should note that in this documentation. Alternatively, we could not document this and refer users to the PSA API specification. Not sure what's best here.
> Beyond that, it looks good to me.

Do we indeed discourage direct usage of this API?

dreemkiller:
I think we should. Others may disagree. @flit @geky your opinions are appreciated here.

docs/api/security/trusted_storage.md
```diff
@@ -0,0 +1,32 @@
+<h2 id="blockdevice-port">PSA Internal Trusted Storage Reference API</h2>
+
+## PSA Internal Trusted Storage
+
+The PSA Internal Trusted Storage is a PSA Root of Trust Service which provides key (uid) value secure storage in internal flash for small data items.
+The main use cases for PSA Internal Trusted Storage APIs are:
+- Provide storage solution for persistent keys in PSA Crypto Service.
+- Store rollback protection tokens for external application secure storage.
```
geky: Should we call this "replay protection" from now on?

geky:
This sounds like a more general question on if we're documenting the PSA APIs in the Mbed OS handbook. Is PSA documentation hosted elsewhere? Forking the documentation here would make it easy for the documentation to go out of date wouldn't it?

For similar reasons we don't document the CMSIS RTOS APIs, since the Mbed OS specific docs cover the C++ RTOS classes instead.

danny4478:
@geky Regarding other PSA documentation, we currently have 2 other PSA related PRs: One is PSA overview page under the security directory in handbook; the other is a PSA Porting Guide.

For now I will add a note that ITS is designed for the crypto service use and that ARM recommends that applications use other storage stacks.

docs/api/security/trusted_storage.md
alzix: @geky, i tend to disagree. It is a first time i encounter "replay protection" terminology, while "rollback protection" is widely adopted term in security domain.

google both term by yourself and see the difference.

@dreemkiller - you are the security expert here - please vote.

geky: This came from a discussion with @dreemkiller and @flit. From what I understand we're NOT providing strict rollback protection as it's known in security. So we are prefering "replay protection" to avoid bad assumptions.

But I'm not an expert though, just didn't know if it was overlooked. So don't let me block anything :+1:

docs/api/security/trusted_storage.md
```diff
@@ -0,0 +1,33 @@
...
+
+The API allows users to:
+* Create a new or modify an existing uid/data pair
+* Retrieve the value associated with a provided uid
+* Retrieve the metadata about the provided uid
+* Remove the provided uid and its associated data from the storage
+
+The Internal Trusted Storage supports a "write once" feature which enforces that the data associated with a key will not be able to be modified or deleted.
+
+**Note:** PSA Internal Trusted Storage is designed for the use of PSA Crypto Service. ARM recommends that applications use other storage stacks.
```
geky:
```suggestion
**Note:** PSA Internal Trusted Storage is designed for the use of PSA Crypto Service. ARM recommends that applications use other storage stacks. You can find the Mbed OS storage stack [here](https://os.mbed.com/docs/v5.10/apis/storage.html)
```

geky:
@danny4478, those are both documents describing how PSA + Mbed OS interact. I guess my question is would a user go to Mbed OS for this document or would they go look at the PSA documentation?

dreemkiller:
"Replay Protection" is the more accurate term.
"Rollback Protection" is generally used for software versioning, where rollback to a previous software version is prohibited.

dreemkiller:
@ndevillard

AnotherButler:
@ndevillard We've made the changes from our call. Please review at your convenience. Also, do you have the PSA link to put here?

AnotherButler:
This PR is waiting for engineering and product approval and a link.

danny4478:
@AnotherButler @ndevillard 
The original intention was to have a section of PSA ITS overview, and then add 2 links: 
1. A link to the Doxygen file
2. A link to the ITS example code
Seeing these changes by @AnotherButler , my questions are:
- What is the content of the new link that Amanda is referring to?
- Do we still plan to add these 2 links? This is a reference API page so I guess Doxygen is a must

danny4478:
@ndevillard I understand there are discussions going on regarding what should be the content of this page and the link.
Any decisions yet?
cc @AnotherButler cc @alzix

docs/api/security/trusted_storage.md
```diff
@@ -0,0 +1,5 @@
+<h2 id="psa-storage">PSA Internal Trusted Storage Reference API</h2>
+
+## PSA internal trusted storage
+
+Mbed OS uses PSA internal trusted storage APIs to support PSA specifications. For more information about PSA internal trusted storage API, please refer to [the documentation](https://developer.arm.com/products/architecture/security-architectures/platform-security-architecture ).
```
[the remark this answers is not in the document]
AnotherButler: @danny4478 Good catch. We're expecting an updated link from @ndevillard by EOW.

PR #823, opened 2018-11-08
workflow: define closure time

0xc0170:
It has been known that more than 2 weeks for PR being idle means it can be closed.
We did not enforce it that often previously.

To be clear, closure does not mean "won't fix" but rather take your time and update
PR by reopening it.

I added this to the PR state (as states are right after that define time boxes for each state, in general more than 2 weeks with no comment/commit - first notification, then few days later closure).

@ARMmbed/mbed-os-maintainers Please review


docs/reference/contributing/guidelines/workflow.md
```diff
@@ -175,6 +175,8 @@ The Mbed OS maintainers add labels to a pull request that represent the pull req
 
 Each state is time boxed. In most cases, this is sufficient time to move to another state. The pull request can be closed if no update is provided within the time frame.
 
+If a pull request is idle for more than 2 weeks, it will closed. It can be reopened with an update anytime by an author or the maintainer.
```
cmonr:
```suggestion
If a pull request is idle for more than 2 weeks, it will be closed. It can be reopened with an update anytime by an author or the maintainer.
```

PR #824, opened 2018-11-09
Quick Start Rework

kegilbert:
Initial rework of the quick start to reference the new "blinky" starting example:
https://github.com/ARMmbed/mbed-os-quick-start-blinky

Putting this up now to get some initial feedback and checkout how it renders in Circle.

@AnotherButler 
@iriark01 

Open questions for docs team:
- We do have a link to "more examples" that goes to the teams page.
   - Do we want to keep pointing to that location long term?
   - Should we make the examples link more prominent by making it bigger, moving it higher up a bit, and/or adding a section for it in the navigation bar on the left?
- The new "blinky" quick start example lives only in Github at the moment. Is there a way to keep the "Import me!" button like I have without it expanding the main.cpp file? Without a valid link there it seems to not work but I can only confirm on the Github markdown for now.
- Related to our previous discussions, we need to finalize a method of importing readme content from the example repos. This discussion is pending, I'll setup a meeting with all of the associated parties on a separate forum so we can iron out the details in more...detail.
- Is moving the environment selection links to header formats to make them stand out ok? Will that cause issues with the navigation bar? Feel free to remove if so.

TODO:
[  ] Engineering review of new example (reviews in process)
[  ] Add link to reworked tutorials page in docs-site
[  ] Link to tagged release of Github example once the tag is created to tie it to 5.10
[  ] Double check links
[  ] Pull in README content from example repo (method of doing this pending)


iriark01:
I don't understand the question about the Import Button. I can promise you it *doesn't* work on GitHub. 

Please create a tag and link to 5.11, not 5.10. This won't be released until the 5.11 full release.

kegilbert:
I don't have permissions to add reviewers/tags it looks like, can you add it please @AnotherButler

kegilbert:
Updated the next page headers to images, will see how they look once the images in #828 are uploaded to the AWS database.

AnotherButler:
Merging to test site - will edit after.

PR #826, opened 2018-11-13
workflow: remove explicit CLA references

0xc0170:
We use inbound-outbound license. This follows the latest update to the
documentation that removed CLA page.

My latest update missed these, I realized after reviewing public docs that these are still there

cc @ARMmbed/mbed-os-maintainers


0xc0170:
@AnotherButler if you find any other reference to CLA, please just remove. I should have caught all with this PR.

PR #852, opened 2018-11-27
Add config options range limit notes

kegilbert:
Additionally add small blurb on how not setting the value field/setting it to null affects the parameter (this might've been described somewhere else, let me know if this is already here and we can remove that portion).


AnotherButler:
ping @theotherjimmy

AnotherButler:
@kegilbert Any idea why this is failing CI?

kegilbert:
@AnotherButler Looks like build timeout issues, I don't have permissions to kick off a rebuild can you try that?

AnotherButler:
I don't have permissions, either. @rik ?

rik:
This is not a timeout issue. It is trying to fetch content from `os-doc-builder.test.mbed.com` which is only reachable inside the ARM network. We've fixed that in the development branch a week ago (by fetching from `os.mbed.com`) so if you rebase this on the latest development branch, it should build well.

kegilbert:
Jimmy approved this, anything else we're waiting on?

AnotherButler:
Should this go live now, or are we waiting on any code releases first?

kegilbert:
This code is currently live, we should be good to go.

Came in with https://github.com/ARMmbed/mbed-os/pull/8673 for the 5.11 release.

PR #864, opened 2018-12-06
Add ThisThread API Page

kegilbert:
@SenRamakri Does this look fine? Any other additions you'd like for the `ThisThread` docs?

### DIFF
--- a/docs/api/api.md
+++ b/docs/api/api.md
@@ -103,6 +103,7 @@ The [Mbed OS RTOS](rtos.html) capabilities include managing objects such as thre
 <tbody>
 <tr>
 <td><a href="thread.html">Thread</a></td>
+<td><a href="thisthread.html">ThisThread</a></td>
 <td><a href="queue.html">Queue</a></td>
 <td><a href="mail.html">Mail</a></td>
 <td><a href="event.html">Event</a></td>
--- a/docs/api/rtos/ThisThread.md
+++ b/docs/api/rtos/ThisThread.md
@@ -1,0 +1,13 @@
+## ThisThread
+
+Use the ThisThread class to control the current thread.
+
+### ThisThread class reference
+
+[![View code](https://www.mbed.com/embed/?type=library)](https://os.mbed.com/docs/development/mbed-os-api-doxy/namespacertos_1_1_this_thread.html)
+
+### ThisThread example
+
+Spawn a thread to blink for 5 seconds before setting a flag to trigger the thread to terminate.
+
+[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed/code/rtos_timer/)](https://os.mbed.com/teams/mbed/code/rtos_timer/file/tip/m<TODO>?.cpp)
--- a/docs/api/rtos/ThisThread.md
+++ b/docs/api/rtos/ThisThread.md
@@ -10,4 +10,4 @@ Use the ThisThread class to control the current thread.
 
 Spawn a thread to blink for 5 seconds before setting a flag to trigger the thread to terminate.
 
-[![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed/code/rtos_timer/)](https://os.mbed.com/teams/mbed/code/rtos_timer/file/tip/m<TODO>?.cpp)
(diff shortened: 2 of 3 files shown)


docs/api/rtos/ThisThread.md
```diff
@@ -0,0 +1,13 @@
+## ThisThread
+
+Use the ThisThread class to control the current thread.
```
AnotherButler: In what circumstances would we use this API over another? (In other words, what are the use cases?)

SenRamakri (approved):
Looks good to me, approved.

docs/api/rtos/ThisThread.md
SenRamakri: I think this class is used only when you are controlling the current thread. If you controlling another thread, you have to use other mechanisms. I'll leave that to @kegilbert to see if we need to explicitly capture that or not.

AnotherButler: When would you want to control the current thread? Are there times you wouldn't?

kegilbert: Thread allows you to create, join, and kill Threads. 
ThisThread allows you to control the currently running thread (Sleep, wait for flags, etc).

kjbracey: `ThisThread` was designed to avoid a specific confusion with static `Thread` methods that quietly acted on the current thread, rather than any thread named. It makes that explicit. All such confusing `Thread` methods have been deprecated.

Confusion was caused because the static methods in `Thread` meant this wasn't clear:

      threadA->signal_set(1);     // signals 1 to `thread`
      Thread::signal_wait(1);     // makes this thread wait for signal 1
      threadA->signal_wait(1);    // XXX makes *this* thread wait for signal 1 - `threadA` value is ignored!

      Thread::gettid();           // reads current thread's ID
      threadA->gettid();          // XXX reads *current* thread's ID, not ThreadA's!

Those confusing static methods are now deprecated, and code would now look like

    threadA->flags_set(1);            // ("signal" changed to "flags" to match CMSIS-RTOS 2)
    ThisThread::flags_wait_any(1); 
    //threadA->flags_wait_any(1);     // this would correctly fail to compile

    threadA->get_id();
    ThisThread::get_id();             // both legal, both do what you expect.

Not quite sure what to do with that information in the documentation. That's the justification for deprecating the old API. (Together with the fact that C11 and C++11 both follow the same new model).

But for new users, just use the methods that are available depending on what you want to do - they're in the correct classes, and will do the right thing if they exist.

There are a bunch of operations which only make sense on this thread like `flags_wait_any`, so only exist in `ThisThread`; many operations would normally only be used on a different thread, so exist only in `Thread` (eg `flags_set`); a couple of operations make sense for either, so exist in both (eg `get_id`).

I guess one thing to note is that a thread may actually have no corresponding Mbed `Thread` object, because you can create a thread directly with CMSIS-RTOS APIs, or you might be `main`'s thread. You couldn't manipulate those with `Thread` methods, but `ThisThread` functions still work from inside them.

AnotherButler:
Is this still waiting on code/something else?

kegilbert:
Shoot thanks for the reminder, this is pending an example still. I'll get on that.

AnotherButler:
Please resolve merge conflicts and Circle failure.

kegilbert:
Apologies for the delay, was using this is a test case for creating new examples in the MonoRepo a bit. Example can be found here:
https://github.com/ARMmbed/mbed-os-example-thisthread

@AnotherButler Did I merge docs.json correctly? Had to fiddle with it while rebasing.

docs/api/api.md
```diff
@@ -103,6 +103,7 @@ The [Mbed OS RTOS](rtos.html) capabilities include managing objects such as thre
 <tbody>
 <tr>
 <td><a href="thread.html">Thread</a></td>
+<td><a href="thisthread.html">ThisThread</a></td>
```
AnotherButler, line 106: This goes down a section and pushes the other APIs down. If you'd like, I can make this change in a separate PR.

AnotherButler:
@kegilbert I think docs.json looks good. Please review my changes, and see my query.

docs/api/api.md
kegilbert: Separate PR is fine with me.

kegilbert:
LGTM, thanks!

PR #867, opened 2018-12-07
If creating new JSON file, must add the * wildcard 1st

ConradBraam:
When doing OOB for 5.11 release, made the mistake of creating a JSON file completely from memory where the special wildcard "*" key appeared after my target overrride.
	"target_overrides": {
		"DISCO_L475VG_IOT01A":{
			"target.network-default-interface-type": "WIFI",	
},
		"*": {
			"target.network-default-interface-type": "ETHERNET"
		},
}
This was never going to do the intended, because the toolchain scans from the top for matches and the wildcard '*' thus overwrites. JSON dictionaries are not supposed to be ordering sensitive, but in the toolchain our dictionary is order-sensitive. So it must appear 1st!

### DIFF
--- a/docs/reference/configuration/configuration.md
+++ b/docs/reference/configuration/configuration.md
@@ -88,7 +88,7 @@ An application may have one `mbed_app.json` in the root of the application and m
 
 The configuration system allows a user to override any defined configuration parameter with a JSON object named `"target_overrides"`.
 
-The keys in the `"target_overrides"` section are the names of a target that the overrides apply to, or the special wildcard `*` that applies to all targets. The values within the `"target_overrides"` section are objects that map configuration parameters, as printed by `mbed compile --config`, to new values. See the example `"target_overrides"` section below.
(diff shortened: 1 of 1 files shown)


theotherjimmy:
@ConradBraam This sounds like a bug. Have you opened an issue with mbed-os?

ConradBraam:
I was actually running a out-of-box, so I wanted to use one environment to build 3 vastly different targets easily. It's less likely that anyone using just one target/job would get stuck here because it needs someone not copy-pasting config files and familiar with JSON being a dictionary not a top-down list. So although the docs already imply that the file is parsed top-down, I'm just re-enforcing that now. But I agree @theotherjimmy , raised a small bug to tweak the parsing in toolchain to scan for the wildcard key first. Jira: IOTTOO-3558

AnotherButler:
What's the status on this? Are we waiting on the tools issue to close?

ConradBraam:
Not dependent on tools change, so this PR should be closed. The tools issue was a background improvement that needed raising, so we should unlink them.

AnotherButler:
Thanks for clarifying; closing now

PR #868, opened 2018-12-10
Document how to turn MPU off at compile time

c1728p9:
Add information to the MPU management page about using the config "platform.disable-mpu" to disable the MPU to save space. Also update the platform configs to include this new value.

This PR has a code dependency on:
https://github.com/ARMmbed/mbed-os/pull/9040

### DIFF
--- a/docs/reference/configuration/Platform.md
+++ b/docs/reference/configuration/Platform.md
@@ -8,10 +8,23 @@ The following is the complete list of platform configuration parameters, as gene
 Configuration parameters
 ------------------------
 
+Name: platform.all-stats-enabled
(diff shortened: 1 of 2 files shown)


docs/reference/configuration/Platform.md
```diff
@@ -94,4 +116,9 @@ Name: platform.thread-stats-enabled
     Description: Set to 1 to enable thread stats. When enabled the function mbed_stats_thread_get_each returns non-zero data. See mbed_stats.h for more information
     Defined by: library:platform
     No value set
+Name: platform.use-mpu
```
JanneKiiskila: This is still a bit too cryptic (or the Mbed OS config system is). So, how do I REALLY disable it? I think I need to  modify the mbed_app.json, but how? (Noobie question, playing the devils advocate here).

PR #886, opened 2018-12-28
Changed NV to volatile in HeapBlockDevice

andrewc-arm:
Media that does not persist across power cycles is volatile. It's not nonvolatile.


AnotherButler:
ping @offirko

yossi2le (approved):
Thanks for catching that one :)

PR #889, opened 2019-01-03
Update example in MeshInterface.md

AnotherButler:
Point example to mesh_led_control_example.cpp instead of main.cpp.

### DIFF
--- a/docs/api/networkinterfaces/MeshInterface.md
+++ b/docs/api/networkinterfaces/MeshInterface.md
@@ -43,7 +43,7 @@ See the example application [mbed-os-example-mesh-minimal](https://github.com/AR
 
 The application below demonstrates a simple light control application, where devices can control the LED status of all devices in the network. You can build the application for the unsecure 6LoWPAN-ND or Thread network.
 
-[![View code](https://www.mbed.com/embed/?url=https://github.com/ARMmbed/mbed-os-example-mesh-minimal)](https://github.com/ARMmbed/mbed-os-example-mesh-minimal/blob/mbed-os-5.10.0/main.cpp)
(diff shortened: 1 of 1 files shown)


docs/api/networkinterfaces/MeshInterface.md
```diff
@@ -43,7 +43,7 @@ See the example application [mbed-os-example-mesh-minimal](https://github.com/AR
 
 The application below demonstrates a simple light control application, where devices can control the LED status of all devices in the network. You can build the application for the unsecure 6LoWPAN-ND or Thread network.
 
-[![View code](https://www.mbed.com/embed/?url=https://github.com/ARMmbed/mbed-os-example-mesh-minimal)](https://github.com/ARMmbed/mbed-os-example-mesh-minimal/blob/mbed-os-5.10.0/main.cpp)
+[![View code](https://www.mbed.com/embed/?url=https://github.com/ARMmbed/mbed-os-example-mesh-minimal)](https://github.com/ARMmbed/mbed-os-example-mesh-minimal/blob/mbed-os-5.10.0/mesh_led_control_example.cpp)
```
artokin: `mesh_led_control_example.cpp` is used from` main.cpp`. I think we need both files for this example.

AnotherButler: Each transcluded example shows one file. Users can get to the other files by clicking the blue text at the top of the rendered example. If that's not enough, we can add an example, so there are two.

artokin: OK, no need to add another example. I'm fine with this change.

artokin (approved):

PR #895, opened 2019-01-08
Use new docs.json format

rosiemiller2:
Create a product object with a name that can be used in the page title tags.

Also take the opportunity to run a formatter.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -1,1178 +1,1182 @@
-[{
-        "title": "Introduction",
-        "description": "Introducing Mbed OS: the overview, architecture, features and licensing",
-        "slug": "introduction",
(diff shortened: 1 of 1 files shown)


docs.json
```diff
@@ -1,1178 +1,1182 @@
...
-        ]
-    },
-    {
-        "title": "Reference",
-        "description": "The reference manual for Mbed OS: APIs, configuration and implementation details",
-        "slug": "reference",
-        "type": "markdown",
-        "chapters": [{
+{
+    "product_name": "Mbed OS",
```
rik:
```suggestion
    "product_name": "Mbed OS 5",
```

PR #916, opened 2019-01-14
5.6 update tools links

bridadan:
Similar to #915 but for the 5.6 branch.


iriark01:
I hate it to call it blind faith, but... basically just merging.

PR #917, opened 2019-01-14
5.7 update tools links

bridadan:
Similar to #915 but for the 5.7 branch.


iriark01:
I'm really just checking that you based against the wrong branch; so long as you don't have 700 files in your PR, we're probably good.

PR #925, opened 2019-01-17
Remove reference to Mbed Device Server

artokin:
Replace Mbed Device Server with Pelion Device Management.


AnotherButler (approved):
Thanks for the PR 👍

PR #931, opened 2019-01-24
Enhance config system doc

aashishc1988:
Enhance config system documentation

### DIFF
--- a/docs/reference/configuration/configuration.md
+++ b/docs/reference/configuration/configuration.md
@@ -15,12 +15,13 @@ Here is how a default JSON file might look like:
         "cellular.random_max_start_delay": "100"
     },
     "K64F": {
-        "cellular.use-apn-lookup": false
+        "cellular.use-apn-lookup": false,
+        "platform.stdio-baud-rate": 9600
     }
 }
 ```
 
-The Arm Mbed OS configuration system gathers and interprets the configuration defined in the target in its [target configuration](../reference/adding-and-configuring-targets.html), all `mbed_lib.json` files and the `mbed_app.json` file. The configuration system creates a single header file, `mbed_config.h`, that contains all of the defined configuration parameters converted into C preprocessor macros. For instance, based on the JSON snippet above, you would see the macro, `MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY` in `mbed_config.h`. `mbed compile` places `mbed_config.h` in the build directory, and `mbed export` places it in the application root. `mbed compile` runs the Mbed configuration system before invoking the compiler, and `mbed export` runs the configuration system before creating project files.
+The Arm Mbed OS configuration system gathers and interprets the configuration defined in the target in its [target configuration](../reference/adding-and-configuring-targets.html), all `mbed_lib.json` files and the `mbed_app.json` file. The configuration system creates a single header file, `mbed_config.h`, that contains all of the defined configuration parameters converted into C preprocessor macros. For instance, based on the JSON snippet above, you would see the macro, `MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY` in `mbed_config.h`. This macro will be created for all targets and would be set to `100`, whereas, macro `MBED_CONF_PLATFORM_STDIO_BAUD_RATE` will be set to `9600` only for K64F. `mbed compile` places `mbed_config.h` in the build directory, and `mbed export` places it in the application root. `mbed compile` runs the Mbed configuration system before invoking the compiler, and `mbed export` runs the configuration system before creating project files.
 
 <span class="notes">**Note:** Throughout this document, "library" means any reusable piece of code within its own directory.</span>
 
@@ -87,7 +88,7 @@ When compiling or exporting, the configuration system generates C preprocessor m
 <file truncated for brevity>
 ```
 
-The name of the macro for a configuration parameter is either a prefixed name or explicitly specified by `macro_name`. The configuration system constructs a prefixed name from the prefix `MBED_CONF_`, followed by the name of the library or `APP`, followed by the name of the parameter. The configuration system then capitalizes the prefixed name and converts it to a valid C macro name. For example, the configuration system converts the `random_max_start_delay` configuration parameter in the library `cellular` to `MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY`. As for using `macro_name` field, it is strongly encouraged to avoid its use unless the intent is to support an already defined macro that is heavily used in Mbed OS, or user applications.
+The name of the macro for a configuration parameter is either a prefixed name or explicitly specified by `macro_name`. The configuration system constructs a prefixed name from the prefix `MBED_CONF_`, followed by the name of the library or `APP`, followed by the name of the parameter. The configuration system then capitalizes the prefixed name and converts it to a valid C macro name. For example, the configuration system converts the `random_max_start_delay` configuration parameter in the library `cellular` to `MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY`. As for using `macro_name` field, it is strongly encouraged to avoid its use unless the intent is to support an already defined macro that is heavily used in Mbed OS, or user applications. Reason we do not encourage the use of `macro_name` is because macros created by JSON, are prefixed with `MBED_CONF_` which makes them easy to identify and link to a particular config file, unlike `macro_names` which can be mistaken to be local for a particular file or feature. 
 
 The Mbed OS build tools instruct the compiler to process the file `mbed_config.h` as if it were the first include of any C or C++ source file, so you do not have to include `mbed_config.h` manually.
 
@@ -95,7 +96,7 @@ Do not edit `mbed_config.h` manually. It may be overwritten the next time you co
 
 ### Configuration parameters in `mbed_app.json`, `mbed_lib.json`
 
-An application may have one `mbed_app.json` in the root of the application and many `mbed_lib.json` files throughout the application. When present, `mbed_app.json` may override configuration parameters defined in libraries and the target, and define new configuration parameters.  
+An application may have one `mbed_app.json` in the root of the application and many `mbed_lib.json` files throughout the application. When present, `mbed_app.json` may override configuration parameters defined in libraries and the target, and define new configuration parameters.
 
 #### Overriding configuration parameters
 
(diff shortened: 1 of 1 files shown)


aashishc1988:
@SenRamakri @kegilbert

docs/reference/configuration/configuration.md
```diff
@@ -7,7 +7,20 @@ The Arm Mbed OS configuration system, a part of the Arm Mbed OS build tools, cus
 - The receive buffer size of a serial communication library.
 - The flash and RAM memory size of an Mbed target.
 
-The Arm Mbed OS configuration system gathers and interprets the configuration defined in the target in its [target configuration](../reference/adding-and-configuring-targets.html), all `mbed_lib.json` files and the `mbed_app.json` file. The configuration system creates a single header file, `mbed_config.h`, that contains all of the defined configuration parameters converted into C preprocessor macros. `mbed compile` places `mbed_config.h` in the build directory, and `mbed export` places it in the application root. `mbed compile` runs the Mbed configuration system before invoking the compiler, and `mbed export` runs the configuration system before creating project files.
+Here is how a default JSON file might look like:
```
SenRamakri: Is it better to say "sample JSON file" instead of "default JSON file"?

docs/reference/configuration/configuration.md
```diff
@@ -74,15 +87,15 @@ When compiling or exporting, the configuration system generates C preprocessor m
...
 
 The Mbed OS build tools instruct the compiler to process the file `mbed_config.h` as if it were the first include of any C or C++ source file, so you do not have to include `mbed_config.h` manually.
 
 Do not edit `mbed_config.h` manually. It may be overwritten the next time you compile or export your application, and you will lose all your changes.
 
 ### Configuration parameters in `mbed_app.json`, `mbed_lib.json`
 
-An application may have one `mbed_app.json` in the root of the application and many `mbed_lib.json` files throughout the application. When present, `mbed_app.json` may override configuration parameters defined in libraries and the target and define new configuration parameters.
+An application may have one `mbed_app.json` in the root of the application and many `mbed_lib.json` files throughout the application. When present, `mbed_app.json` may override configuration parameters defined in libraries and the target, and define new configuration parameters.
```
kegilbert: Minor whitespace fix
```suggestion
An application may have one `mbed_app.json` in the root of the application and many `mbed_lib.json` files throughout the application. When present, `mbed_app.json` may override configuration parameters defined in libraries and the target, and define new configuration parameters.
```

docs/reference/configuration/configuration.md
SenRamakri: Also I think a different example showing usage of "name" and some other common fields might be more helpful. May be we can refer one of the mbed_lib.jsons

docs/reference/configuration/configuration.md
```diff
@@ -7,7 +7,20 @@ The Arm Mbed OS configuration system, a part of the Arm Mbed OS build tools, cus
 - The receive buffer size of a serial communication library.
 - The flash and RAM memory size of an Mbed target.
 
-The Arm Mbed OS configuration system gathers and interprets the configuration defined in the target in its [target configuration](../reference/adding-and-configuring-targets.html), all `mbed_lib.json` files and the `mbed_app.json` file. The configuration system creates a single header file, `mbed_config.h`, that contains all of the defined configuration parameters converted into C preprocessor macros. `mbed compile` places `mbed_config.h` in the build directory, and `mbed export` places it in the application root. `mbed compile` runs the Mbed configuration system before invoking the compiler, and `mbed export` runs the configuration system before creating project files.
+Here is how a default JSON file might look like:
+
```
kegilbert, line 13: I like the example. The text below sort of goes into what this will produce but might also be nice to have a small statement before it saying what the actual end effect will be (...default JSON file that configures X for all targets and Y for the K64F might look like:)

docs/reference/configuration/configuration.md
```diff
@@ -74,15 +87,15 @@ When compiling or exporting, the configuration system generates C preprocessor m
...
+The name of the macro for a configuration parameter is either a prefixed name or explicitly specified by `macro_name`. The configuration system constructs a prefixed name from the prefix `MBED_CONF_`, followed by the name of the library or `APP`, followed by the name of the parameter. The configuration system then capitalizes the prefixed name and converts it to a valid C macro name. For example, the configuration system converts the `random_max_start_delay` configuration parameter in the library `cellular` to `MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY`. As for using `macro_name` field, it is strongly encouraged to avoid its use unless the intent is to support an already defined macro that is heavily used in Mbed OS, or user applications.
```
SenRamakri: It would be good to add a sentence mentioning that "macro_name" usage is not recommended because its easy to figure out that MBED_CONF_ flags are driven by config files.

docs/reference/configuration/configuration.md
```diff
@@ -256,7 +269,7 @@ It is an error to both add and subtract the same value from a cumulative attribu
 
 ### `mbed_app.json` Specification
 
-`mbed_app.json` may be present at the root of your application or specified as the argument of the `--app-config` parameter to `mbed compile` and `mbed export`. The configuration system interprets only one `mbed_app.json` during `mbed compile` or `mbed export`, unlike library configuration. Like `mbed_lib.json`, `mbed_app.json` is a JSON formatted document that contains a root JSON Object. The keys within this object are sections. The allowed sections and their meanings are below:
+`mbed_app.json` may be present at the root of your application or specified as the argument of the `--app-config` parameter to `mbed compile` and `mbed export`. When you create a new mbed project using `mbed new`, `mbed_app.json` is created by default in the root of the application. The configuration system interprets only one `mbed_app.json` during `mbed compile` or `mbed export`, unlike library configuration. Like `mbed_lib.json`, `mbed_app.json` is a JSON formatted document that contains a root JSON Object. The keys within this object are sections. The allowed sections and their meanings are below:
```
kegilbert: Minor syntax nit (...unlike library configuration->configurations):
```suggestion
`mbed_app.json` may be present at the root of your application or specified as the argument of the `--app-config` parameter to `mbed compile` and `mbed export`. When you create a new mbed project using `mbed new`, `mbed_app.json` is created by default in the root of the application. The configuration system interprets only one `mbed_app.json` during `mbed compile` or `mbed export`, unlike library configurations. Like `mbed_lib.json`, `mbed_app.json` is a JSON formatted document that contains a root JSON Object. The keys within this object are sections. The allowed sections and their meanings are below:
```

kegilbert (commented):
Left a minor comment, looks awesome thanks @aashishc1988 !

docs/reference/configuration/configuration.md
kegilbert: Apologies if he added this after your comment, but he mentions:
>As for using `macro_name` field, it is strongly encouraged to avoid its use unless the intent is to support an already defined macro that is heavily used in Mbed OS, or user applications.

SenRamakri: @kegilbert - Yes he mentions that, what I meant was it might be useful to say why we are discouraging people from using 'macro_name' instead of just saying we discourage that usage. Not a big deal, just a thought.

docs/reference/configuration/configuration.md
aashishc1988: I can add few more fields, like baud rate to the example.

docs/reference/configuration/configuration.md
aashishc1988: I can add little more detail to it.

docs/reference/configuration/configuration.md
aashishc1988: I like it. Let me modify accordingly.

aashishc1988:
@kegilbert @SenRamakri @AnotherButler 
Made recommended changes. Please review.

docs/reference/configuration/configuration.md
```diff
@@ -15,12 +15,13 @@ Here is how a default JSON file might look like:
         "cellular.random_max_start_delay": "100"
     },
     "K64F": {
-        "cellular.use-apn-lookup": false
+        "cellular.use-apn-lookup": false,
+        "platform.stdio-baud-rate": 9600
```
kegilbert, line 21: Since 9600 is the default baud rate for this target, may be worth changing to something like `115200` (no strong feelings one way or the other though).

docs/reference/configuration/configuration.md
```diff
@@ -7,7 +7,21 @@ The Arm Mbed OS configuration system, a part of the Arm Mbed OS build tools, cus
 - The receive buffer size of a serial communication library.
 - The flash and RAM memory size of an Mbed target.
 
-The Arm Mbed OS configuration system gathers and interprets the configuration defined in the target in its [target configuration](../reference/adding-and-configuring-targets.html), all `mbed_lib.json` files and the `mbed_app.json` file. The configuration system creates a single header file, `mbed_config.h`, that contains all of the defined configuration parameters converted into C preprocessor macros. `mbed compile` places `mbed_config.h` in the build directory, and `mbed export` places it in the application root. `mbed compile` runs the Mbed configuration system before invoking the compiler, and `mbed export` runs the configuration system before creating project files.
+Here is how a default JSON file might look like:
```
SenRamakri: @aashishc1988 @kegilbert - Here we say "default" JSON file. Is that ok? I thought this is more of a sample json file.
Default implies this would be the default content.

docs/reference/configuration/configuration.md
aashishc1988: Hmm. I see your point. However changing it to something that is not valid for that target makes me bit nervous.

docs/reference/configuration/configuration.md
aashishc1988: Actually when i think more about it, we are talking about an example, so sample would be more apt here.

docs/reference/configuration/configuration.md
kegilbert: Good point. Since this is configuring for the K64F specifically I'd think it would be fine but could be confusing if someone is using a different target and just copy/pasted this.

aashishc1988: For a parameter like baud rate there is only really one right answer, but I see your point. In that case, can you suggest an alternative parameter that is not so susceptible to change in targets. Alternatively, I can add a note which says that this parameter is for K64F and might be different for another target.

aashishc1988: Actually looking at the description we have for the snippet, it should be clear that the baud-rate value here should only be set for K64F.

aashishc1988:
@SenRamakri @kegilbert 
Ready for another review.

kegilbert (approved):

SenRamakri (approved):

docs/reference/configuration/configuration.md
```diff
@@ -7,7 +7,23 @@ The Arm Mbed OS configuration system, a part of the Arm Mbed OS build tools, cus
...
+        "cellular.random_max_start_delay": "100"
+    },
+    "K64F": {
+        "cellular.use-apn-lookup": false,
+        "platform.stdio-baud-rate": 9600
+    }
+}
+```
+
+If you use the example JSON snippet above, you can see the macro `MBED_CONF_CELLULAR_RANDOM_MAX_START_DELAY` in `mbed_config.h`. This macro is created for all targets and is set to `100`, whereas macro `MBED_CONF_PLATFORM_STDIO_BAUD_RATE` is set to `9600` only for the K64F. `mbed compile` places `mbed_config.h` in the build directory, and `mbed export` places it in the application root. `mbed compile` runs the Mbed configuration system before invoking the compiler, and `mbed export` runs the configuration system before creating project files.
```
AnotherButler, line 26: I don't understand the bit about the K64F. We don't mention it before this line. Could you please clarify?

aashishc1988: So if we take a look at the snippet, there are parameters that are defined under asterik (it is not letting me use that character here), and others that are defined for specific target (K64F in this case). The intent is to make sure that users understand the scope of a macro value based on the targets its defined for. Would it make the documentation clearer if I mention here that asterik is a wildcard that applies to all targets?

AnotherButler: I think it would, but only if our developers aren't expected to know that.

AnotherButler: If they are, then please ignore.

aashishc1988: That stuff is very basic, and it is mentioned later in this documentation.

AnotherButler (approved):

PR #941, opened 2019-02-06
Use path instead of url

rik:
And remove optional `"type": "markdown"` inside books


rik:
@AnotherButler Feel free to merge once the whole thing is green on CircleCI.

PR #947, opened 2019-02-11
workflow: add item for splitting bigger pull requests

0xc0170:
Use dependencies, to avoid long review time. Updating 3rd party driver should
not be part of refactor or target update for instance.

Waiting for some feedback how to improve PR time - we experienced some big PR recently and would like to avoid having these landing if it could be prevented

@ARMmbed/mbed-os-maintainers

### DIFF
--- a/docs/reference/contributing/guidelines/workflow.md
+++ b/docs/reference/contributing/guidelines/workflow.md
@@ -63,6 +63,7 @@ Pull requests on GitHub have to meet the following requirements to keep the code
 - Avoid merging commmits. (Always rebase when possible.)
 - Comment in the pull request on every change (rebase or new commits). This helps reviewers to be up to date with changes
 - Pull requests should fix a bug, add a feature or refactor.
+- Smaller pull requests are easier to review, faster to integrate. Use dependencies – split your work by pull request type or functional changes. Adding 3rd party driver ? Send it in a separate pull request and add it as a dependency to your pull request.
 
 If commits do not follow the above guidelines, we may request you modify the commit history (often to add more details to address _what_ and _why_ rather than _how_).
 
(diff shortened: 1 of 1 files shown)


0xc0170:
cc @Patater

Patater:
From the email thread, this is the proposal: split SDK dumps that aren't reviewed to the same level into their own folder (e.g. `thirdparty`) in Mbed OS, from other partner contributions that should be reviewed to the same level as other Mbed OS contributions.

> One idea on the table to make these sorts of code dumps easier to review, and the current "double standard" more apparent, is to have the SDK dump done in a separate PR from the Mbed OS specific integration code.
> For example, we could do the following.
> - Add the latest Freescale Kinetis SDK (KSDK) in a PR to a special "contrib" or "thirdparty" folder, emphasizing that the code is not reviewed as carefully by the Mbed OS team and the quality of which is mainly guaranteed by the partner (Freescale/NXP in this case).
> - Add the code to integrate the SDK with Mbed OS in a separate PR, which should be smaller and easier for us to review properly. This code is more new (likely not reviewed or as well tested by the partner as their primary SDK) and particular to Mbed OS, and benefits the most from an in-depth Mbed OS expert review.

docs/reference/contributing/guidelines/workflow.md
```diff
@@ -63,6 +63,7 @@ Pull requests on GitHub have to meet the following requirements to keep the code
 - Avoid merging commmits. (Always rebase when possible.)
 - Comment in the pull request on every change (rebase or new commits). This helps reviewers to be up to date with changes
 - Pull requests should fix a bug, add a feature or refactor.
+- Smaller pull requests are easier to review, faster to integrate. Use dependencies – split your work by pull request type or functional changes. Adding 3rd party driver ? Send it in a separate pull request and add it as a dependency to your pull request.
```
Patater: Does "3rd party driver" mean the same thing as "code copy pasted from the partner SDK"?

0xc0170: yes, any code coming outside of Mbed OS tree (so we won't style it for instance).

AnotherButler (approved):

AnotherButler:
@Patater Does this look OK?

cmonr (approved):
LGTM

AnotherButler:
If anyone has any further comments, please make them in a new PR.

PR #948, opened 2019-02-11
Config Output Auto-update Script

kegilbert:
Add a manually run script that will replace the config system verbose output in the docs with the output from the config list command. Bit of a simple implementation without many automated checks at the moment, I'd still recommend double checking the output pretty closely until after some more review.

The LWIP page had to be consolidated for this to work but everything else should be unchanged. 

Tested mainly with Python3, Python2 worked but ran into an occasional UTF-8 encoding error (I can poke around that more if need be).


kegilbert:
@cmonr, @geky is busy and can't review this atm. Left Geky on so he can double check the LWIP file changes if you'd be free to double check the scripty-boi. Thanks!

check_tools/config-update.py
```diff
@@ -0,0 +1,63 @@
...
+                file = file[:start+4] + out[:out.index("Macros") - 1] + file[end:]
+
+        except Exception as e:
+            print("Error")
+            print(e)
+            print("____________________")
+            exc_type, exc_obj, exc_tb = sys.exc_info()
+            fname = os.path.split(exc_tb.tb_frame.f_code.co_filename)[1]
+            print(exc_type, fname, exc_tb.tb_lineno)
+            pass
```
geky, line 39: I asked offline why this has exception handling here as it looks like it doesn't do anything. I was wrong, this one passes (is there another way to do this?). You may want a comment to explain this for idiots like me :)

geky (approved):

check_tools/config-update.py
kegilbert: Updooted, I'm also OK with removing this section as its more of a vestigial hold out from development (although can still be helpful, let me catch minor exceptions that didn't cause issues without exiting and provided a full list of all files that caused issues).

AnotherButler (approved):
Thanks for the PR. This looks great 👍

AnotherButler:
@cmonr Is this ready to merge?

AnotherButler:
@cmonr It was this one.

check_tools/config-update.py
```diff
@@ -0,0 +1,65 @@
...
+        end    = ranges[i][1]
+
+        print(snippet_indices)
+
+        try:
+            blocks[i] = file[start : end + 3]
+            if ('Name: ' in blocks[i]):
+                lib = blocks[i].split('Name: ')[1].split('.')[0]
+                print("=================   %s   =================" % lib)
+                out = subprocess.check_output(["mbed", "compile", "--config", "-v", "--prefix", lib]).decode()
```
cmonr: You probably want:

```suggestion
                out = subprocess.check_output(["mbed", "compile", "--config", "-v", "--prefix", lib]).encode('utf-8')
```

For Py2/Py3 compat.

check_tools/config-update.py
```diff
@@ -0,0 +1,65 @@
...
+        path = sys.argv[1]
+
+    if (path == '-h' or path == '--help'):
+        print("By default the script runs out of the docs tools directory and iterates through reference/configuration.\n"
+              "You may pass in a directory path that will run on all files contained within, or a single file path optionally.")
+        exit(0)
+
+    if (os.path.isfile(path)):
+        main(path)
+    else:
```
cmonr: Due to the arg parsing in https://github.com/ARMmbed/mbed-os-5-docs/pull/948/files#diff-e52e11a85f5305dc93291efd2f777122R54, this should verify that path is a directory or throw an error.

check_tools/config-update.py
kegilbert: The subprocess output in Python3 is a Bytes object that I cannot encode. I had added  the decode initially to allow it to run in both Python2/3 (although I did run the following error in nvstore in Python2):

```
'ascii' codec can't decode byte 0xe2 in position 996: ordinal not in range(128)
```

cmonr: https://github.com/ARMmbed/mbed-os/pulls?q=is%3Apr+is%3Aclosed+author%3Acmonr

A couple of my recent PRs were Python 2/3 compat related if you want to take a look.

Imo, it might be better if this works with Py3 if you had to choose one. 

Also, I _might_ be confusing `encode` and `decode`.

cmonr: It would be good to know _what_ causes the occasional convert error.

cmonr:
> Add a manually run script

If this is manually run, how is this expected to be used? When, where, by who, and how often?

AnotherButler:
The longterm goal is to get this into CI, so it's completely automated. In the short term, the docs team can run it every feature release if given written, detailed instructions.

kegilbert:
Correct I'd love to see this in CI, but until then it _should_ be as simple to run as:

```
# Run from the check_tools directory or pass in the docs/reference/configuration path manually
python config-update.py 
#OR  
python check_tools/config-update.py docs/reference/configuration

# View the results with git diff locally first or just push up the changes as a new PR and view
# from there. Feel free to use whatever git methodology you'd like, just showing an example here
git diff
git checkout -b <BRANCH>
git add docs/reference/configuration/*.md
git commit 
git push <REMOTE> <BRANCH>
# Create PR
```

AnotherButler:
@cmonr Have all of your concerns been addressed?

kegilbert:
@cmonr Fixed the string formatting, runs on Python 2.7/3.5

cmonr (approved):

cmonr:
I personally feel like the script could use more documentation, but I'm not going to block the PR on it.

https://blog.codinghorror.com/coding-for-violent-psychopaths/

@kegilbert We'll see how well this comment ages 😁

kegilbert:
I'll put up another a bit with some better docs

PR #954, opened 2019-02-18
license: add extension to LICENSE file

0xc0170:
The license file without the extension is shown as "raw" text file. The links do not work as they should. As this file
should point to files with non Apache licenses (the most used case) it should be viewed as markdown.

This should address the issue found here https://github.com/ARMmbed/mbed-os/pull/9502#issuecomment-464666713

Reviewing https://help.github.com/articles/licensing-a-repository/ (they also mention either no extension or .md extension). We rather use .md extension so license can be viewed on Github and be navigated (see the main LICENSE file in the root, overview of all license files in the tree).

@ARMmbed/mbed-os-maintainers

### DIFF
--- a/docs/reference/contributing/guidelines/license.md
+++ b/docs/reference/contributing/guidelines/license.md
@@ -30,7 +30,7 @@ The full text of the license can be found on the [Apache website](http://www.apa
 
 In order to clearly reflect the Apache 2.0 license, please create two text files:
 
-- A *LICENSE* file with the following text:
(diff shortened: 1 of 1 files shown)


0xc0170:
The file view without extension https://github.com/ARMmbed/mbed-os/blob/master/LICENSE

AnotherButler (approved):
Thanks for the PR 👍

cmonr (approved):

PR #965, opened 2019-02-26
Quickstart baremetal notes

kegilbert:
Add Bare Metal build profile instructions to the quick start. Additionally remove hardcoded expected serial output. The serial output is not updated each release and is liable to change. As such I've removed until we have a method to automatically update the output each patch release.


SenRamakri (approved):

AnotherButler:
Does this one also require [ARMmbed/mbed-os#9800](https://github.com/ARMmbed/mbed-os/pull/9800)?

AnotherButler (approved):
Thanks for the PR 👍

kegilbert:
@AnotherButler Good point, yes it does. The `bare-metal` could be changed to:
```JSON
{
    "requires": ["platform", "drivers"]
}
```
for the same effect, but would probably be better to wait for the other PR to merge.

bulislaw (approved):

PR #968, opened 2019-02-26
docs: I2CSlave API: Add feature check to example code

naveenkaje:
Add pre-processor feature check to the example code so that
it's easier for the user to know if the feature is enabled
on the platform under test


AnotherButler:
@SenRamakri Please review this or tag someone else if you're too busy or not the right person.

AnotherButler (approved):
Thanks for the PR 👍

SenRamakri (approved):

PR #970, opened 2019-02-27
Correct command line for listing supported exporters

theotherjimmy:
The command line included is valid since 5.5.0.


AnotherButler (approved):
Thanks for the quick PR 👍

PR #984, opened 2019-02-28
Removed PR template from docs

cmonr:
We have this: https://github.com/ARMmbed/mbed-os/blob/master/.github/pull_request_template.md

Do we a duplicate in the docs that then also needs to be maintained?

(For reference, this was brought up by @mrcoulter45 when creating https://github.com/ARMmbed/mbed-os/pull/9895#issue-257229865)

### DIFF
--- a/docs/reference/contributing/guidelines/workflow.md
+++ b/docs/reference/contributing/guidelines/workflow.md
@@ -141,33 +141,6 @@ A contribution containing a breaking change is the most difficult PR to get merg
 
 Release: major
 
-### Pull request template
-
-Below is a good example of a pull request:
-
-```
-Fix deep sleep locking bug
-
-# Description
-
-Fix problems that could leave deep sleep locked unintentionally, along with adding tests to verify this behavior is fixed.
-
-Tested locally with two targets and all toolchains.
-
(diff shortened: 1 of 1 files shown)


adbridge:
IF this repo wants to use any of our scripts properly then it will need to keep the template....

0xc0170:
The intention was to show an example of good pull request . However it's hard to keep up to date if we improve (=change) things. We have this covered in how to create PR, should be sufficient.

Fine to be removed.

AnotherButler:
I like not duplicating content that could get out of sync, but not being able to use scripts sounds concerning.

AnotherButler:
@0xc0170 @adbridge Could y'all please confirm whether this will affect scripts?

adbridge:
> 
> 
> IF this repo wants to use any of our scripts properly then it will need to keep the template....

The key is this, you don't have to have the template as long as people raise the PRs using the format that is in the template. That format is essential for the scripts. So it is up to you whether you lead people by giving the template automatically, or continually check that people have raised their PRs using the right format. Personally I still think your template is the best option, people rarely read guidelines properly....

AnotherButler (approved):

PR #988, opened 2019-03-04
Relative bootloader offset clarification

bridadan:
It wasn't clear before that these offset values are relative to the start of ROM. I have called this detail out.


AnotherButler (approved):
Thanks for the PR 👍

theotherjimmy (approved):

PR #997, opened 2019-03-07
Remove example from psa_attestation.md

AnotherButler:
Remove example because we do not link to examples in `TESTS`.


AnotherButler:
I discussed this example with @SenRamakri  and @kegilbert , and we cannot link to `TESTS` code for example because it has the Greentea constructs that are not required.

Please create an example for this API that does not pull in unnecessary content. Please don't hesitate to ask us if you want help or more information about where and how to create this.

GuyWi:
I have updated the document with an example provided by @avolinski. 
@kegilbert, could you please review and approve?
Thank you!

kegilbert (approved):
LGTM from a basic example standpoint, I'm not well versed in PSA so can't comment on that aspect. Made a minor refactor PR : https://github.com/ARMmbed/mbed-os-example-attestation/pull/1

PR #998, opened 2019-03-07
Remove bare metal references in porting.md

AnotherButler:
Remove references to bare metal Blinky because it has been renamed direct register Blinky.


kegilbert (approved):
Thanks!

PR #999, opened 2019-03-07
Added documentation for using ARM Compiler 5 with 5.12 release

SenRamakri:
This is documentation addition to capture how a developer can force ARM Compiler 5 with targets
already moved to ARM Compiler 6 with 5.12 release.

### DIFF
--- a/docs/tools/tools_intro.md
+++ b/docs/tools/tools_intro.md
@@ -53,7 +53,7 @@ be able to compile with ARM Compiler 5 or you may see undefined behaviors. In ot
 
 ##### By creating a mbed_app.json to override `supported_toolchains`
 
-In this method, you can create or update your `mbed_app.json` with the following content. Note that you can still keep other entries such as `GCC_ARM` or `IAR` while overriding `supported_toolchains` as below.
+In this method, you can create or update your `mbed_app.json` with the following content. Override `supported_toolchains` as below.
 
 ```
 {
--- a/docs/tools/tools_intro.md
+++ b/docs/tools/tools_intro.md
@@ -44,3 +44,39 @@ For more information, please see the [Online Compiler page](developing-mbed-onli
 
 You can export your project from any of our tools to third party tools. For instructions, as well as tool-specific information, see [the Exporting to third party toolchains page](exporting.html).
 
+#### Forcing compilation with ARM Compiler 5 for targets already supporting ARM Compiler 6
+
+It's possible that some developers may need to update to Mbed 5.12 release but still requires compiling with ARM Compiler 5 until they are in possession of ARM Compiler 6.
+In those cases, you may still be able to use ARM Compiler 5 depending on the target. If your target uses any ARM Compiler 6 specific binaries or code, then it may not
+be able to compile with ARM Compiler 5 or you may see undefined behaviors. In other cases, if you want to try force ARM Compiler 5 you can do so with the following options:
+
+##### By creating a mbed_app.json to override `supported_toolchains`
+
+In this method, you can create or update your `mbed_app.json` with the following content. Note that you can still keep other entries such as `GCC_ARM` or `IAR` while overriding `supported_toolchains` as below.
+
+```
+{
+  "target_overrides": {
+      "*": {
+          "target.supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"]
+      }
+  }
+}
+```
+
+##### By local modifications to `targets.json`
+
+In this method, you have to modify the `supported_toolchains` entry for your target in targets.json to remove all `ARM`, `ARMC6` entries and replace it with `ARMC5`. Note that you can still keep other entries such as `GCC_ARM` or `IAR`.
+
+See below for example:
+```
+"MY_TARGET_NAME": {
+        "supported_form_factors": [...],
+        "core": "Cortex-M4",
+        "supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"],
+        ...
+}
+```
+
+<span class="note"> **Note:** The above methods to override ARM toolchain version is made available only to enable developers migrating from ARM Compiler 5 to ARM Compiler 6. Future releases of Mbed OS may remove
+this option and thus developers are strongly encouraged to move to ARM Compiler 6.</span>
--- a/docs/tools/tools_intro.md
+++ b/docs/tools/tools_intro.md
@@ -45,31 +45,24 @@ For more information, please see the [Online Compiler page](developing-mbed-onli
 
 You can export your project from any of our tools to third party tools. For instructions, as well as tool-specific information, see [the Exporting to third party toolchains page](exporting.html).
 
-#### Forcing compilation with ARM Compiler 5 for targets already supporting ARM Compiler 6
+<div style="background-color:#F3F3F3; text-align:left; vertical-align: middle; padding:15px 30px;"> **Note:** We encourage you to switch to Arm Compiler 6 soon because we will deprecate Arm Compiler 5 support in the future. However, if you need to update to Mbed OS 5.12 but still require compiling with Arm Compiler 5 until you are in possession of Arm Compiler 6, we provide methods to override the Arm toolchain version. If you do this, your target may not be able to compile with Arm Compiler 5, or you may see undefined behaviors.
 
-It's possible that some developers may need to update to Mbed 5.12 release but still requires compiling with ARM Compiler 5 until they are in possession of ARM Compiler 6.
-In those cases, you may still be able to use ARM Compiler 5 depending on the target. If your target uses any ARM Compiler 6 specific binaries or code, then it may not
-be able to compile with ARM Compiler 5 or you may see undefined behaviors. In other cases, if you want to try force ARM Compiler 5 you can do so with the following options:
+To force Arm Compiler 5, you can use the following options:
 
-##### By creating a mbed_app.json to override `supported_toolchains`
-
-In this method, you can create or update your `mbed_app.json` with the following content. Override `supported_toolchains` as below.
+- Create or update your `mbed_app.json` with:
 
 ```
 {
   "target_overrides": {
       "*": {
-          "target.supported_toolchains": ["ARMC5"]
+          "target.supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"]
       }
   }
 }
 ```
 
-##### By local modifications to `targets.json`
+- Modify the `supported_toolchains` entry in targets.json to replace all `ARM`, `ARMC6` entries with `ARMC5`: 
 
-In this method, you have to modify the `supported_toolchains` entry for your target in targets.json to remove all `ARM`, `ARMC6` entries and replace it with `ARMC5`. Note that you can still keep other entries such as `GCC_ARM` or `IAR`.
-
-See below for example:
 ```
 "MY_TARGET_NAME": {
         "supported_form_factors": [...],
@@ -78,5 +71,4 @@ See below for example:
         ...
 }
 ```
-
-<span class="note"> **Note:** The above methods to override ARM toolchain version is made available only to enable developers migrating from ARM Compiler 5 to ARM Compiler 6. We encourage developers to make plans to switch to ARM Compiler 6 soon, as we plan to deprecate ARM Compiler 5 support in the future and this migration would ensure that your software is compatible with it. </span>
+</div> 
--- a/docs/tools/tools_intro.md
+++ b/docs/tools/tools_intro.md
@@ -59,7 +59,7 @@ In this method, you can create or update your `mbed_app.json` with the following
 {
   "target_overrides": {
       "*": {
-          "target.supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"]
+          "target.supported_toolchains": ["ARMC5"]
       }
   }
 }
--- a/docs/tools/tools_intro.md
+++ b/docs/tools/tools_intro.md
@@ -39,13 +39,17 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
 
 For more information, please see the [Online Compiler page](developing-mbed-online-compiler.html).
 
+<span class="note"> **Note:** Arm Compiler 6 is the default ARM toolchain for Mbed OS developmet. Most Mbed OS platforms are already compatible with Arm Compiler 6. Some existing targets still supporting Arm Compiler 5 will also be migrated to ARM Compiler 6 in the future. Please be aware that you must use Arm Compiler 6 for future development, and we will validate all code contributions to Mbed OS with Arm Compiler 6. </span>
+
 ##### Third party development tools
 
 You can export your project from any of our tools to third party tools. For instructions, as well as tool-specific information, see [the Exporting to third party toolchains page](exporting.html).
 
-<div style="background-color:#F3F3F3; text-align:left; vertical-align: middle; padding:15px 30px;"> **Note:** Arm Compiler 6 is the default ARM toolchain for Mbed OS development. Most Mbed OS platforms are already compatible with Arm Compiler 6. Some existing targets still supporting Arm Compiler 5 will also be migrated to ARM Compiler 6 in the future. Please be aware that you must use Arm Compiler 6 for future development, and we will validate all code contributions to Mbed OS with Arm Compiler 6. We encourage you to switch to Arm Compiler 6 soon because we will deprecate Arm Compiler 5 support in the future. However, if you need to update to Mbed OS 5.12 but still require compiling with Arm Compiler 5 until you are in possession of Arm Compiler 6, we provide methods to override the Arm toolchain version. If you do this, your target may not be able to compile with Arm Compiler 5, or you may see undefined behaviors.
+<div style="background-color:#F3F3F3; text-align:left; vertical-align: middle; padding:15px 30px;"> **Note:** We encourage you to switch to Arm Compiler 6 soon because we will deprecate Arm Compiler 5 support in the future. However, if you need to update to Mbed OS 5.12 but still require compiling with Arm Compiler 5 until you are in possession of Arm Compiler 6, we provide methods to override the Arm toolchain version. If you do this, your target may not be able to compile with Arm Compiler 5, or you may see undefined behaviors.
 
-To force Arm Compiler 5, you can create or update your `mbed_app.json` with:
+To force Arm Compiler 5, you can use the following options:
+
+- Create or update your `mbed_app.json` with:
 
 ```
 {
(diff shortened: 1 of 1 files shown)


SenRamakri:
@AnotherButler - Please review this new modifications to tools_intro. Im also not sure if this is in the right document, if not please let me know and I'll move it. Also before merging I would like to get broader acceptance on this content. So, Ill go ahead add more reviewers here.

SenRamakri:
@ChiefBureaucraticOfficer - Please see if the new content addressed the concern with developers updating to 5.12 and not able to use ARMC5.

bridadan (commented):
> In those cases, you may still be able to use ARM Compiler 5 depending on the target. If your target uses any ARM Compiler 6 specific binaries or code, then it may not be able to compile with ARM Compiler 5 or you may see undefined behaviors.

If a target has a prebuilt binary, we don't currently have a way of identifying the toolchain it was compiled with right?

Are any of the prebuilt managed bootloaders inside of mbed os still built with ARMC5 only?

SenRamakri:
@bridadan - You are right, we dont have a way to tell the toolchain used to build a binary.
And regarding bootloader binaries it could have been GCC_ARM or ARMC5. But I'm assuming that should not be an issue as they are independently built and are not linked together, if that's what you are concerned about.

theotherjimmy:
@SenRamakri Why are we documenting this?

SenRamakri:
> @SenRamakri Why are we documenting this?

Good question @theotherjimmy, there has been concerns raised around the fact that developers currently using ARMC5 cannot compile 5.12 with ARMc5 without knowing these options. So, we are trying to document this more explicitly if it helps. But please raise concerns if you think this can only cause more confusion. Maybe its easier for people to move to 5.12 when they have ARMC6 development environment instead of using these options.

bridadan (changes requested):
> And regarding bootloader binaries it could have been GCC_ARM or ARMC5

Oops yup good point, it was late in the day and my brain wasn't firing on all cylinders 😄.

I'd really suggest we don't document this unless we are prepared to test 5.12 and later releases with ARMC5 for all targets in CI. Just because it works in 5.12.0 doesn't mean it'll work 5.12.1, 5.12.2, etc.

I really think the appropriate place for this is in the release notes and the release blog.

SenRamakri:
@ChiefBureaucraticOfficer - Should we rather capture this in blog/release notes as @bridadan pointed out above?

ghost:
> @ChiefBureaucraticOfficer - Should we rather capture this in blog/release notes as @bridadan pointed out above?

Let me confer with the product guys.  It should be clear how to use the system with C5 if needed, but the question is the correct placement of that messaging.

theotherjimmy:
> Maybe its easier for people to move to 5.12 when they have ARMC6 development environment instead of using these options.

This is my preference. I don't think we want to offer them a way to enable _untested_ _unverified_ compilers. Otherwise, why do we even have `target.supported_toolchains` at all?

screamerbg:
As I posted here to another Github thread.

@bridadan I just checked with the Mbed product management and according to them, we are not "switching" to ARM Compiler 6 - we are making "Arm Compiler 6" the default ARM Compiler. We are also keeping support for ARM Compiler 5 as one of the supported compilers until the support for it is officially announced as deprecated by the *Product Management*. In fact, we have a number of Partners and Customers that will continue to use ARM Compiler 5 with Mbed OS 5.12 as a main vehicle for their Products (check with @ChiefBureaucraticOfficer), including a customer of U-blox.

This documentation is key for these customers.

screamerbg (approved):
Looks great! Thank you @SenRamakri. @MarceloSalazar, @ashok-rao, @maclobdell Can you review for your customer cases and OOB?

docs/tools/tools_intro.md
```diff
@@ -44,3 +44,39 @@ For more information, please see the [Online Compiler page](developing-mbed-onli
...
+```
+"MY_TARGET_NAME": {
+        "supported_form_factors": [...],
+        "core": "Cortex-M4",
+        "supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"],
+        ...
+}
+```
+
+<span class="note"> **Note:** The above methods to override ARM toolchain version is made available only to enable developers migrating from ARM Compiler 5 to ARM Compiler 6. Future releases of Mbed OS may remove
```
screamerbg: Can this be rephrased to reflect the Product Management direction, e.g. "We encourage you to plan to make use of ARM Compiler 6 soon. We do plan to deprecate ARM Compiler 5 support in the future and this migration would ensure that your software is compatible with it."

ghost: Agree with softening of tone, but still pressing the message.

docs/tools/tools_intro.md
```diff
@@ -44,3 +44,39 @@ For more information, please see the [Online Compiler page](developing-mbed-onli
 
 You can export your project from any of our tools to third party tools. For instructions, as well as tool-specific information, see [the Exporting to third party toolchains page](exporting.html).
 
+#### Forcing compilation with ARM Compiler 5 for targets already supporting ARM Compiler 6
+
+It's possible that some developers may need to update to Mbed 5.12 release but still requires compiling with ARM Compiler 5 until they are in possession of ARM Compiler 6.
```
ashok-rao: Could we add a list of currently known targets that cannot be used with ARMC6 yet (IIRC.. PSoC6, KW24D, any others) ?

ashok-rao (commented):
1 minor comment, otherwise LGTM!

MarceloSalazar:
I've just found we're not mentioning Arm Compiler 5 in the list of supported compilers for 5.12.
Please tweak the docs as suggested here:

---

Mbed OS 5 can be built with various toolchains. The currently supported versions are:
- [Arm Compiler 6.11](https://developer.arm.com/products/software-development-tools/compilers/arm-compiler/downloads/version-6) (default)
- [Arm Compiler 5.06 update 6](https://developer.arm.com/products/software-development-tools/compilers/arm-compiler/downloads/version-5) (to be deprecated in the future)
...

ghost (approved):
My approval is not required, so long as Mohit, @screamerbg and @SenRamakri are aligned.

docs/tools/tools_intro.md
SenRamakri: @ashok-rao - I think we should not capture that list explicitly here as we plan to move the rest of the targets to ARMC6 soon, updating the list as each target gets moved into ARMC6 might be error-prone or we may miss it and docs also need re-publishing. Instead, the mbed_targets.md captures the fact that if the supported_toolchains contains ARMC5 the target uses ARMC5. In addition you can also list the toolchain support for targets using the command "mbed config -S" which is already documented.

SenRamakri:
I have made the review comment fixes. There was a question around whether this documentation belongs here as if its going to create more confusion than helping with ARMC6 migration. But based on broader feedback it appears we need this documented and I'm ok with providing this content here, particularly because other ways of documenting this like blogs, release notes can make it harder to find. But note that this will probably be removed in one of the future releases once we have all targets moved to ARMC6/when we deprecate ARMC5 support.

docs/tools/tools_intro.md
```diff
@@ -44,3 +45,38 @@ For more information, please see the [Online Compiler page](developing-mbed-onli
 
 You can export your project from any of our tools to third party tools. For instructions, as well as tool-specific information, see [the Exporting to third party toolchains page](exporting.html).
 
+#### Forcing compilation with ARM Compiler 5 for targets already supporting ARM Compiler 6
```
AnotherButler: Can we make this a note and shorten it?

Maybe something like:

<span class="note"> **Note:** We encourage you to switch to Arm Compiler 6 soon because we will deprecate Arm Compiler 5 support in the future. However, if you need to update to Mbed OS 5.12 but still require compiling with Arm Compiler 5 until you are in possession of Arm Compiler 6, we provide methods to override the Arm toolchain version. If you do this, your target may not be able to compile with Arm Compiler 5, or you may see undefined behaviors.</span>

To force Arm Compiler 5, you can use the following options:

- Create or update your `mbed_app.json` with:

```
{
  "target_overrides": {
      "*": {
          "target.supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"]
      }
  }
}
```

- Modify the `supported_toolchains` entry in targets.json to replace all `ARM`, `ARMC6` entries with `ARMC5`: 

```
"MY_TARGET_NAME": {
        "supported_form_factors": [...],
        "core": "Cortex-M4",
        "supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"],
        ...
}
```

docs/tools/tools_intro.md
```diff
@@ -21,7 +21,8 @@ We created the Mbed command-line tool (Mbed CLI), a Python-based tool, specifica
 
 Mbed OS 5 can be built with various toolchains. The currently supported versions are:
 
-- [Arm compiler 6.11](https://developer.arm.com/products/software-development-tools/compilers/arm-compiler/downloads/version-6).
+- [Arm compiler 6.11 (default ARM toolchain)](https://developer.arm.com/products/software-development-tools/compilers/arm-compiler/downloads/version-6).
+- [Arm compiler 5.06 update 6 (to be deprecated in the future)](https://developer.arm.com/products/software-development-tools/compilers/arm-compiler-5/downloads).
```
theotherjimmy: This strikes me as odd: how is "to be deprecated in the future" different from "deprecated"? My understanding of deprecated is that it means: discourage the use of in favor of a newer or better alternative.

docs/tools/tools_intro.md
```diff
@@ -44,3 +45,38 @@ For more information, please see the [Online Compiler page](developing-mbed-onli
...
+
+##### By creating a mbed_app.json to override `supported_toolchains`
+
+In this method, you can create or update your `mbed_app.json` with the following content. Note that you can still keep other entries such as `GCC_ARM` or `IAR` while overriding `supported_toolchains` as below.
+
+```
+{
+  "target_overrides": {
+      "*": {
+          "target.supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"]
```
theotherjimmy, line 58: I don't think that we should encourage users to turn of GCC_ARM and IAR support for targets, especially targets that may not support them. This section is about ARMC5, let's keep it that way.

```suggestion
          "target.supported_toolchains": ["ARMC5"]
```

docs/tools/tools_intro.md
```diff
@@ -44,3 +45,38 @@ For more information, please see the [Online Compiler page](developing-mbed-onli
...
 
+#### Forcing compilation with ARM Compiler 5 for targets already supporting ARM Compiler 6
+
+It's possible that some developers may need to update to Mbed 5.12 release but still requires compiling with ARM Compiler 5 until they are in possession of ARM Compiler 6.
+In those cases, you may still be able to use ARM Compiler 5 depending on the target. If your target uses any ARM Compiler 6 specific binaries or code, then it may not
+be able to compile with ARM Compiler 5 or you may see undefined behaviors. In other cases, if you want to try force ARM Compiler 5 you can do so with the following options:
+
+##### By creating a mbed_app.json to override `supported_toolchains`
+
+In this method, you can create or update your `mbed_app.json` with the following content. Note that you can still keep other entries such as `GCC_ARM` or `IAR` while overriding `supported_toolchains` as below.
```
theotherjimmy: This is about ARMC5, so can we avoid mentoining GCC_ARM and IAR please?
```suggestion
In this method, you can create or update your `mbed_app.json` with the following content. Override `supported_toolchains` as below.
```

docs/tools/tools_intro.md
```diff
@@ -44,3 +45,38 @@ For more information, please see the [Online Compiler page](developing-mbed-onli
...
+{
+  "target_overrides": {
+      "*": {
+          "target.supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"]
+      }
+  }
+}
+```
+
+##### By local modifications to `targets.json`
```
theotherjimmy: We don't recommend that applications modify Mbed OS elsewhere in the documentation. Perhaps this needs to be in a porting targets section instead.

theotherjimmy (changes requested):
L.et's leave GCC_ARM and IAR out of this.

docs/tools/tools_intro.md
theotherjimmy: See my comment below about `targets.json` modifications. I think that bit may need to move to target porting.

SenRamakri: @ChiefBureaucraticOfficer @screamerbg - Please review if the shortened note of the content provided by @AnotherButler above works. 

@theotherjimmy - I don't think this belongs in porting section just because we refer targets.json. Also it makes hard to find if its in porting section as this would be the page most people would land on to find compiler info.

docs/tools/tools_intro.md
SenRamakri: @theotherjimmy - My understanding is that its not deprecated yet as we still have targets still compiling with ARMC5, but will be deprecated soon. This provides the warning that developers should actively look into moving to ARMC6 toolchain. Once we have all targets moved to ARMC6 we will probably say its deprecated and remove it completely in the release afterwards. 
@MarceloSalazar - Do you have more comments on this?

MarceloSalazar: @SenRamakri 
The important point here is that Arm Compiler 5 is still supported in 5.12.
As you've incorporated this already, I don't have more comments. Thanks!

MarceloSalazar: Btw, the correct name of the toolchain is "Arm Compiler" - with "C" upper case - see links.

theotherjimmy: If we know something will be discouraged in favor of an alternative, how is that not discouraging it's use already? That's my point really: that telling people it'll be discouraged in the future is the same as discouraging it's use now. I'm pretty sure this verbiage is common, so I'm not going to block on such a bike shed.

theotherjimmy (approved):
My only remaining issue is that the local modifications to `targets.json` is really for porting targets, not for applications. I don't think that's something that needs to be addressed in this PR.

docs/tools/tools_intro.md
TeroJaasko: Should a modification above to mbed_app.json be enough? Just testing mbed-os-example-blinky with that on mbed OS master (b80c961da) and it still fails to build on ARMC5.
```
mbed compile -m K64F -t ARMC5
<..>
make.py: error: argument -t/--tool: ARMC5 is not a supported toolchain. Supported toolchains are:
ARM,     ARMC6,   GCC_ARM, IAR,     uARM     
[mbed] ERROR: "/usr/bin/python" returned error.
```

SenRamakri: @TeroJaasko - Yes, please follow the changes as mentioned here to enable ARMC5 and try. By the way "-t ARMC5" is an invalid option. After adding ARMC5 to supported_toolchains you should still do "-t ARM". Please let me know if you need help with making the changes.

TeroJaasko: I already copied the ```"target.supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"]``` to mbed_app.json and the results are no better with "-t ARM".
```
mbed compile -m K64F -t ARM
<..>
make.py: error: Could not find executable for ARMC6.
```

Diff on mbed-os-example-blinky:
```
mbed-os-example-blinky$ git diff
diff --git i/mbed_app.json w/mbed_app.json
index 3b191ff..c106c06 100644
--- i/mbed_app.json
+++ w/mbed_app.json
@@ -1,6 +1,7 @@
 {
     "target_overrides": {
         "*": {
+            "target.supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"],
             "platform.stack-stats-enabled": true,
             "platform.heap-stats-enabled": true,
             "platform.cpu-stats-enabled": true,
```

SenRamakri: @TeroJaasko - Although we tested this in past, I tried the same thing again with the exact version you are using(b80c961da) and it appears to be working fine on my machine, and yes, the compilation works with ARMC5. And yet, I don't see anything wrong with what you are trying to do.

SenRamakri: I also tried the same on Ubuntu and that worked fine as well. So I wonder whats causing your failure :-(. One thing you can make sure is that you are in mbeb-os-example-blinky dir and not in mbed-os when doing mbed compile.

bridadan (approved):

docs/tools/tools_intro.md
TeroJaasko: Have you tried removing ARMC6 from paths? I don't have ARMC6 at my paths by default, there are at least one ticket open where wrong compiler is used: https://github.com/ARMmbed/mbed-os/issues/10069

0xc0170 (approved):
I would also move porting part out of this intro document and reference it here. The porting detail can be in porting docs.

The rest LGTM

bulislaw (approved):

docs/tools/tools_intro.md
```diff
@@ -44,3 +45,30 @@ For more information, please see the [Online Compiler page](developing-mbed-onli
...
+{
+  "target_overrides": {
+      "*": {
+          "target.supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"]
+      }
+  }
+}
+```
+
+- Modify the `supported_toolchains` entry in targets.json to replace all `ARM`, `ARMC6` entries with `ARMC5`:
```
AnotherButler: Shall I create a new PR moving this to porting?

AnotherButler: Or would config be a better location? We do talk about target.json here: https://os.mbed.com/docs/mbed-os/development/reference/adding-and-configuring-targets.html

theotherjimmy: That's porting documentation. It talks about how to port targets.

docs/tools/tools_intro.md
```diff
@@ -38,9 +39,21 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
...
+
+```
+{
+  "target_overrides": {
+      "*": {
+          "target.supported_toolchains": ["ARMC5", "GCC_ARM", "IAR"]
+      }
+  }
+}
+```
```
AnotherButler, line 62: I moved the porting content out in another PR: https://github.com/ARMmbed/mbed-os-5-docs/pull/1013

bridadan (approved):
Changes still look good, the clarification around the porting section in particular looks really good :+1:

docs/tools/tools_intro.md
```diff
@@ -44,3 +45,30 @@ For more information, please see the [Online Compiler page](developing-mbed-onli
...
+
+- For porting a target that cannot use Arm Compiler 6 at this time, modify the `supported_toolchains` entry in `targets.json` to replace all `ARM` and `ARMC6` entries with `ARMC5`:
```
theotherjimmy, line 64: It's not really a modification if you're doing new target porting...

theotherjimmy (approved):

AnotherButler (approved):

docs/tools/tools_intro.md
SenRamakri: Ah ok, that makes sense if you dont have ARMC6 configured. I believe the tools would check for ARMC6 as in this PR https://github.com/ARMmbed/mbed-os/pull/10044 , have you tried adding ARMC6 and see if that fixes the problem?

AnotherButler:
Is this ready to merge?

SenRamakri:
@AnotherButler - I think so, as we have it approved by all of our active reviewers.

PR #1003, opened 2019-03-08
Add docs about new test macro

bridadan:
This feature was added in Mbed CLI version 1.9.0 by @screamerbg. This adds documentation on how to use it.

### DIFF
--- a/docs/tools/testing/testing_greentea.md
+++ b/docs/tools/testing/testing_greentea.md
@@ -14,6 +14,16 @@ You can run tests throughout Mbed OS and for your project's code. They are locat
 
 The fact that the code is located under this directory means that it is ignored when building applications and libraries. It is only used when building tests. This is important because all tests require a `main()` function, and building them with your application would cause multiple `main()` functions to be defined.
 
+The macro `MBED_TEST_MODE` is defined when building tests with Mbed CLI versions 1.9.0 and later. You can wrap your application's `main()` function in a preprocessor check to prevent multiple `main()` functions from being defined:
+
+```c++
+#if !MBED_TEST_MODE
+int main() {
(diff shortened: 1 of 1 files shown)


AnotherButler (approved):
Thanks for the PR 👍

docs/tools/testing/testing_greentea.md
```diff
@@ -14,6 +14,16 @@ You can run tests throughout Mbed OS and for your project's code. They are locat
 
 The fact that the code is located under this directory means that it is ignored when building applications and libraries. It is only used when building tests. This is important because all tests require a `main()` function, and building them with your application would cause multiple `main()` functions to be defined.
 
+The macro `MBED_TEST_MODE` is defined when building tests with Mbed CLI versions 1.9.0 and later. You can wrap your application's `main()` function in a preprocessor check to prevent multiple `main()` functions from being defined:
+
+```c++
+#if !MBED_TEST_MODE
```
screamerbg, line 20: For consistency shall this be `#ifndef MBED_TEST_MODE`? See https://github.com/ARMmbed/pelion-ready-example/blob/master/main.cpp

screamerbg (approved):
Thanks for this. See comment.

docs/tools/testing/testing_greentea.md
bridadan: I went with `#if !MBED_TEST_MODE` so you could also set `MBED_TEST_MODE=0` for testing purposes. Specifying it as `#ifndef` would always evaluate to `false`, even if the macro was set to 0. A similar pattern is used within Mbed OS: https://github.com/ARMmbed/mbed-os/pull/8957

PR #1007, opened 2019-03-11
Add tag based configuration parameter splitting

kegilbert:
https://github.com/ARMmbed/mbed-os-5-docs/pull/981 (mesh configuration file) logically splits up the `mbed_mesh_api` configuration parameters to improve readability. The automatic config update script could not handle this previously. 

Now (as in the case of the Mesh API), if there is a uniquely common tag in how each segment is split up, the script all lump common tags together if requested by including the tag in the configuration page's markdown as in the following example (see test and test2 tags):

```
### This is a configuration parameter doc page!
Here's how you use the `test` parameters in the lib1 library.

'''test
Configuration parameters
--------------------------
Name: lib1.test-a
    Description: I'm just a test
    ....
Name: lib1.test-b
    Description: I'm another test parameter
    ....
....
'''

### This is another section
Here's how you use the `test2` parameters in the lib1 library.

'''test2
Configuration parameters
--------------------------
Name: lib1.test2-a
    Description: I'm just a test
    ....
Name: lib1.test2-b
    Description: I'm another test parameter
    ....
....
'''

```

### DIFF
--- a/check_tools/config-update.py
+++ b/check_tools/config-update.py
@@ -105,17 +105,19 @@ def main(file):
                     print("\t------- Tag: %s -------" % tag)
 
                     start_of_config_block = file.find('Name:', start)
-                    updated_config = str(file[ : start_of_config_block]) # + "Configuration parameters\n------------------------\n")
+                    updated_config = str(file[ : start_of_config_block])
                     for line in out.splitlines():
                         if 'Name' in line and tag in line:
                             updated_config += line
 
                             # Collect all text until next parameter name. If there's no following 'Name:' token, its the last
-                            # config option, match to 'Macros' instead to termiante the block
-                            if out.find('Name:', out.find(line) + 4) > 0:
-                                updated_config += out[out.find('\n', out.find(line)) : out.find('Name:', out.find(line) + 4)]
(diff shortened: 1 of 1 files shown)


AnotherButler (approved):
Thanks for the PR 👍

kegilbert:
Thanks for the edits, no idea what I was typing there...

check_tools/config-update.py
```diff
@@ -45,6 +45,21 @@ def split_into_pairs(l):
...
+    """ Determine if the provided string contains
+        alphabetical characters (case insensitive)
+    Args:
+    line - string to scan
+
+    Returns:
+    True if the string contains [a-z], else False
+    """
+    regexp = re.compile(r'[a-z]', re.IGNORECASE)
+    if regexp.search(line):
```
cmonr: Why not

```suggestion
    return regexp.search(line)
```

cmonr (commented):
A nit, but otherwise LGTM

kegilbert:
@cmonr Thanks! Updated `is_string` to just return the match object.
@AnotherButler Should be good to go.

cmonr (approved):

PR #1009, opened 2019-03-12
QuickStart toolchain version and location update

kegilbert:
@AnotherButler Should I even mention the version or just say the Arm compiler? Then we'd only have to update the compiler page in the future. IAR and GCC_ARM also don't specify a version here.

I was also going to add a link to https://os.mbed.com/docs/mbed-os/development/tools/index.html but wasn't sure the best linking mechanic when the note mentions the "supported compiler page"


AnotherButler:
Let's be consistent and remove the number and just link "supported compiler page" to the tools landing page you linked above.

kegilbert:
@AnotherButler Updated, let me know if that link formatting is correct.

AnotherButler (approved):
Looks great 👍 Thanks for the PR

PR #1013, opened 2019-03-15
Update note in mbed_targets.md

AnotherButler:
Update note to include method to force ArmC5.

### DIFF
--- a/docs/reference/configuration/mbed_targets.md
+++ b/docs/reference/configuration/mbed_targets.md
@@ -198,9 +198,17 @@ If you specify `ARMC5` in `supported_toolchains`, it means the corresponding tar
 
 If a target lists both `ARMC5` and `ARM` (or `ARMC6`) in `supported_toolchains`, the Arm Compiler 6.11 will be used when compiling with `ARM` option for `--toolchain`.
 
-<span class="notes">**Note:** Although you can specify `ARMC5` in `supported_toolchains` in `targets.json`, it's not a valid option for `--toolchain` when compiling using [Mbed CLI](../tools/developing-mbed-cli.html).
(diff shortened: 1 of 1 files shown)


SenRamakri (changes requested):
@AnotherButler - I'm not sure we have to split that into 2 pages. This makes it harder to find. Now we have 2 ways of doing same thing explained in 2 pages. And also we eventually have to remove this and individual modifications to these section in separate pages makes it harder. I would prefer this to go in the same place as other doc and but its ok to make it shorter as you recommended without losing the main message that we will be deprecating ARMC5 in future.

theotherjimmy:
@SenRamakri This does the same thing, but it's for different users. In particular `mbed_app.json` modification is for an application developer. This is for someone porting a target. I would not expect them to look in the same location for this information.

SenRamakri:
@theotherjimmy - Although in general that is the case, I don't think anyone porting new targets should use AC5. Also compiler documentation will be applicable to both porting and app development not just app development.

theotherjimmy:
> I don't think anyone porting new targets should use AC5.

The this should not be documented.

theotherjimmy:
> Also compiler documentation will be applicable to both porting and app development not just app development.

This is documentation about `targets.json` which only applies to porting. Application developers should not be encouraged to modify Mbed OS.

theotherjimmy (approved):

SenRamakri:
I think there is some confusion, it was never meant to be a "porting" related doc, its only meant to document how to use AC5. But yes, it does refer targets.json which is one way of doing it.

theotherjimmy:
@SenRamakri @AnotherButler We could instead place this in the toolchain section, so long as this advice is clearly marked as "For porting a target that cannot use ArmC6 at this time". That way Application developer are not presented with an option that encourages modification of Mbed OS, as they're not doing target porting. (from offline discussion)

theotherjimmy:
Part of the confusion is that when `targets.json` gets involved, it's immediately about target porting.

SenRamakri:
@AnotherButler - I'll make the changes to the PR as @theotherjimmy suggested and push to other PR. Thanks for the quick review @theotherjimmy and @AnotherButler.

SenRamakri:
@AnotherButler - Please close this PR as I have modifications to the other PR as we discussed.

PR #1014, opened 2019-03-15
Update uvision version to 6

aashishc1988:
Since we have moved to arm compiler 6, during export the uvision version should be changed to 6. In future, we will track it based on supported toolchain instead of hardcoding a value.


AnotherButler (approved):
Thanks for the PR 👍

bridadan (approved):

PR #1015, opened 2019-03-18
Add document on using TLS-on-PSA

Patater:
Document at a high level how one would go about using Mbed TLS with with
PSA Crypto (TLS-on-PSA) to use opaque ECDSA keys for TLS client
authentication, use opaque ECDSA keys to generated certificate signing
requests (CSRs), use opaque pre-shared keys with TLS PSK ciphersuites,
for both Mbed OS 5.11 and 5.12.

### Is there a code dependency?
All necessary code has been merged to Mbed OS master and 5.12 via https://github.com/ARMmbed/mbed-os/commit/5285b39c01f03a05f9aafcfd489889e3ee73bd80


AnotherButler:
This is good content, but it doesn't seem to fit well here. Because the note states the content is likely to change and because most of the document compares and contrasts 5.11 with 5.12, I wonder if this is better suited for a blog post. @ndevillard What are your thoughts?

AnotherButler:
If we keep this in the technical documentation, I recommend we move it from API references because it's not one. We could move it to migrations in the tutorials, perhaps, as it's all about moving to PSA. I'd like to remove version numbers from it in that case because it's weird to me to keep a guide that is only relevant to one feature release.

Patater:
@RonEld Where do you think it best for this to go? Does Mbed OS have KBs?

RonEld:
@Patater This is the Mbed OS KBs .
But I agree with @AnotherButler that this is not the correct location.
I think it should be in the `porting` sub folder. Perhaps `porting/psa` ? or perhaps `porting/target`?

ndevillard:
Agreed: this is a porting guide to migrate to PSA. porting/psa sounds like a good place if we can create it, will certainly get more docs later.

Patater:
OK, I moved it to the porting/psa folder.

docs/porting/psa/TLS-on-PSA.md
```diff
@@ -0,0 +1,114 @@
...
+1. Build normally, for example (with GNU Make) make all test or (with CMake) `mkdir build && cd build && cmake .. && make all test`.
+
+### Switching from pre-PSA Mbed TLS to PSA-enabled Mbed TLS
+
+You can use applications built with pre-PSA Mbed TLS with PSA-enabled Mbed TLS by recompiling them against a PSA-enabled version of Mbed TLS. (Note that relinking is not enough, though.) In other words, PSA-enabled Mbed TLS is currently fully API-compatible with non-PSA-enabled Mbed TLS, but not ABI compatible. All existing (pre-PSA) APIs are preserved and continue to exist in parallel with new PSA-based APIs, to ease the transition for users at this stage.
+
+_Please note that this is likely to change at a later date though, as existing interfaces are likely to be deprecated in favor of the new, PSA-based APIs - if, when and how it happens is outside the scope of this document._
+
+### Switching from PSA-enabled Mbed TLS in Mbed OS 5.11 to PSA-enabled Mbed TLS in Mbed OS 5.12
```
AnotherButler: I don't want this to be a document just about going from 5.11 to 5.12. That narrows it to being relevant to only one quarter. Could we instead position it as how to enable PSA in Mbed TLS, without mentioning version numbers?

docs/porting/psa/TLS-on-PSA.md
```diff
@@ -0,0 +1,114 @@
...
+
+If your application is not using any of the new PSA-specific APIs described below, then it can be used with the version in Mbed OS 5.12 by just recompiling it. (Note that relinking is not enough, though.)
```
AnotherButler: Similarly, let's remove references to "new" and "old" and instead focus on using PSA-specific APIs.

docs/porting/psa/TLS-on-PSA.md
```diff
@@ -0,0 +1,114 @@
...
+#### Application flow without PSA
+
+1. At application startup, make sure `mbedtls_platform_setup()` is called if relevant.
+1. Declare (and allocate) an object of type `mbedtls_pk_context`.
+1. Load a key into that PK context, presumably using `mbedtls_pk_parse_key()` (or `mbedtls_pk_parse_keyfile()`).
+1. Configure the key for use in client authentication with a matching certificate by calling `mbedtls_ssl_conf_own_cert()` on the PK context.
+1. Use the resulting SSL configuration for TLS connections as usual.
+1. When you're done using TLS connections based on that configuration, _and no longer want to use that key for anything else_, free the PK context using `mbedtls_pk_free()`.
+
+#### Application flow with PSA in Mbed OS 5.11 (new/changed parts emphasized):
```
AnotherButler: This says we've emphasized the new/changed parts. Do we include the unchanged parts elsewhere in the Mbed OS docs, or are they just in the TLS docs? Can we link to the unchanged parts instead of including them here and bolding the new ones?

AnotherButler (commented):
I've left some general comments and queries.

AnotherButler:
I edited this to make sense in the context of 5.12. Please review.

AnotherButler:
@Patater @ndevillard Please review my edits, so this can be merged.

AnotherButler:
I'm going to merge this into 5.12. If you'd like any changes made, please open a new PR.

PR #1021, opened 2019-03-20
OOB_5.12 Clarification regarding label directories

GuyWi:
This PR is related to a documentation enhancement requested by @alzix as part of https://github.com/ARMmbed/mbed-os-5-docs/pull/1018.

### DIFF
--- a/docs/reference/build_rules.md
+++ b/docs/reference/build_rules.md
@@ -32,7 +32,7 @@ Label directories are directories that follow a naming convention: an upper case
 
 The Mbed OS build tools skip all label directories, unless you configure them explicitly, as described above.
 
-For more information about using, extending, and removing labels, see (Adding and configuring targets)[../reference/adding-and-configuring-targets.html].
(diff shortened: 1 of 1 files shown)


theotherjimmy (changes requested):
This is incorrect. `target.extra_labels_add` is not a different configuration variable, and `target.extra_labels_remove` is also available.

Please document that you may use both add and remove to modify the `target.extra_labels` value in addition to directly setting it.

GuyWi:
Thanks for your review and clarification, @theotherjimmy !
Please approve.

theotherjimmy (changes requested):
The value `extra_labels_add` is not something a user can configure; and the `TARGET` labels are all built from a single configuration value. Same for features and components.

GuyWi:
Thanks again!
I failed to modify that. 
Updated.

theotherjimmy (approved):

PR #1029, opened 2019-03-22
Revert "Using Mbed Studio's ARMC6 with Mbed CLI"

arekzaluski:
Reverts ARMmbed/mbed-os-5-docs#1028

@bridadan I do not think that documenting the location of the tools that we are shipping with Mbed Studio is a good idea. They are designed to be isolated and to work only with Mbed Studio.
Mbed CLI should not reuse tools that we are shipping inside of Mbed Studio. The way that it should be done is for mebd CLI installers to ship its own ARMC6.

We also can't commit to the location of tools. In fact release 0.5 (next release) is going to change tools location. It is going to be specific on each of the platform supported by Mbed Studio.


bulislaw (changes requested):
Too late now, we need to give people way of using ARMC6 for free as we made it the default compiler.

iriark01:
Closing as agreed with Bartek

PR #1033, opened 2019-03-26
Rejoin bare-metal example docs with bare-metal folder 

kegilbert:
Did a little backend cleanup to move the bare-metal example to the correct folder location following the bare-metal book relocating.


AnotherButler (approved):
Thanks for the PR 👍

PR #1036, opened 2019-03-29
Fix default path to toolchains

MarceloSalazar:
These are simple tweaks to:
- Replace references to wrong AC6 toolchain with the correct one
- Use default and the most common path to AC6 toolchain
- Clarify version of IAR (as they use a different path name that doesn't match with the tools version)
- Adds ARMC6_PATH option

@AnotherButler


AnotherButler (approved):
Thanks for the PR 👍

SenRamakri (approved):

bridadan (approved):
Nice!

PR #1047, opened 2019-04-26
Connectivity review

melwee01:


docs/reference/technology/connectivity/connectivity.md
```diff
@@ -32,41 +35,41 @@ Our Socket API standardizes all of our connectivity options. It supports both IP
 
 Mbed OS provides network drivers, such as Ethernet, Wi-Fi and cellular.
 
-You can find descriptions of supported interfaces and protocols in the [IP networking](ip-networking.html) section.
+You can find descriptions of supported interfaces and protocols in the [IP networking](#ip-networking.html) section.
```
iriark01: This is a page line, you can't use # in it. It's just missing the the ../book-name/ bit

docs/reference/technology/connectivity/connectivity.md
```diff
@@ -32,41 +35,41 @@ Our Socket API standardizes all of our connectivity options. It supports both IP
...
+BLE is optimized for cheap, battery-based devices, and has a low memory footprint. It is widely adopted for connected spaces, but requires setting up an infrastructure. Its range is up to 100 meters, with 1 Mbps bandwidth. You need an edge gateway so devices in your network can connect to the internet.
 
 Typical applications of BLE are health care, fitness trackers, beacons, smart homes, security, entertainment, proximity sensors, industrial applications, and automotives.
 
 To learn how to use BLE on Mbed OS, please refer to the [Bluetooth overview](../apis/ble.html).
 
 ###  IP-based mesh (Thread, Wi-SUN and 6LoWPAN-ND)
 
-As a technology designed with large-scale deployment in mind, mesh is optimized for long battery life and low cost.
+Originally designed with large-scale deployment in mind, mesh is optimized for long battery life and low cost.
```
iriark01: This implies that the goal has since changed

melwee01:
```suggestion
Designed with large-scale deployment in mind, mesh is optimized for long battery life and low cost.
```

docs/reference/technology/connectivity/connectivity.md
melwee01:
```suggestion
You can find descriptions of supported interfaces and protocols in the [IP networking](../reference/ip-networking.html) section.
```

docs/reference/technology/connectivity/connectivity.md
```diff
@@ -32,41 +35,41 @@ Our Socket API standardizes all of our connectivity options. It supports both IP
...
 
 End devices for Wi-Fi are complicated due to hardware requirements and limitations. The network cost depends on whether the deployment location already includes a well maintained Wi-Fi infrastructure. Wi-Fi is ubiquitous indoors, and Wi-Fi devices have a range of tens of meters, so for certain applications, like those for homes and offices, the infrastructure likely already exists.
 
 ### Cellular
 
-Cellular has a variable throughput, reaching 10 mbps. Its coverage is extensive, and its range is 25-100km, so it is commonly used in transportation and logistics. It is also well suited for proof of concept, because, like Wi-Fi, it relies on an existing infrastructure. It's suitable for a wide variety of use cases, from monitoring temperature sensors, to streaming video.
+Cellular has a variable throughput, reaching 10 Mbps. Its coverage is extensive, and its range is 25-100km, so it is commonly used in transportation and logistics. It is also well suited for proof of concept, because, like Wi-Fi, it relies on an existing infrastructure. It's suitable for a wide variety of use cases, from monitoring temperature sensors, to streaming video.
```
peknis01: Always have a space between the number and unit -> 25-100 km

docs/reference/technology/connectivity/connectivity.md
```diff
@@ -32,41 +35,41 @@ Our Socket API standardizes all of our connectivity options. It supports both IP
...
+**CAT-M1** has a higher data rate than NB-IoT (closer to existing cellular behavior), but is not proven to be more power efficient. Any efficiency gains can be lost because of operator configurations.
 
 It can be deployed on a single spectrum, reducing antenna configuration complexity, but isn't deployed world-wide, and can take years to be deployed in countries using 3G.
 
 ### Bluetooth Low Energy (BLE)
 
-BLE is optimised for cheap, battery-based devices, and has a low memory footprint. It is widely adopted for connected spaces, but requires setting up an infrastructure. Its range is up to 100 meters, with 1 mbps bandwidth. You need an edge gateway so devices in your network can connect to the internet.
+BLE is optimized for cheap, battery-based devices, and has a low memory footprint. It is widely adopted for connected spaces, but requires setting up an infrastructure. Its range is up to 100 meters, with 1 Mbps bandwidth. You need an edge gateway so devices in your network can connect to the internet.
```
peknis01: Edge, if we are talking about our Edge. But are we?

docs/reference/technology/connectivity/connectivity.md
```diff
@@ -98,21 +101,21 @@ The [LoRa](lora-tech.html) section and [LoRa tutorial](../tutorials/LoRa-tutoria
...
+Near-field communication (NFC) is a short range (few centimeters) wireless technology standard for personal area networks. Typical uses of NFC are commissioning, easy sharing of small content, and Bluetooth connection initiation/out-of-band pairing.
 
 To learn how to use NFC with Mbed OS, please refer to the [Mbed OS NFC overview](../apis/nfc.html).
 
 ### NB-IoT cellular
 
-Non-IP Data Delivery (NIDD) is a new feature for communication over NB-IoT. It is enabled by Control Plane cellular IoT EPS optimization and meant to provide improved support of small data transfer. It does this by transporting user data over the control channel, thus reducing the total number of control plane messages when handling a short data transaction.
+Non-IP Data Delivery (NIDD) is a new feature for communication over NB-IoT. It is enabled by Control Plane cellular IoT EPS optimization, to provide improved support for small data transfer. It does this by transporting user data over the control channel, thus reducing the total number of control plane messages when handling a short data transaction.
```
peknis01: It does this by transporting... -> It transports user data...
How about that?

peknis01 (changes requested):

docs/reference/technology/connectivity/connectivity.md
melwee01: Ack. Been putting those in everywhere else, but missed one. 😅

docs/reference/technology/connectivity/connectivity.md
melwee01: Don't know.

docs/reference/technology/connectivity/lora-tech.md
```diff
@@ -51,7 +51,7 @@ The base stations transmit a beacon every 128 seconds, and all Class B nodes are
 
 <span class="images">![](https://s3-us-west-2.amazonaws.com/mbed-os-docs-images/class_b_timing.png)<span>Figure 3: Class B timing diagram</span></span>
 
-Beacon guard time precedes each beacon, and no ping slot can be placed in that time period. A ping slot is a 30ms unit that you can assign to a Class B device. Beacon reserved is a time period when the actual beacon is sent. Beacon window is the time period when you can open and assign ping slots.
+Beacon guard time precedes each beacon, and no ping slot can be placed in that time period. A ping slot is a 30 ms unit that you can assign to a Class B device. Beacon reserved is a time period when the actual beacon is sent. Beacon window is the time period when you can open and assign ping slots.
```
melwee01, line 54: Unfortunately not 'bacon guard'.

docs/reference/technology/connectivity/wifi.md
```diff
@@ -20,7 +20,7 @@ The original version of the standard IEEE 802.11 was released in 1997. Updates a
 | 802.11n | 2009 | 6.5 - 600 | Both 2.4 Ghz and 5 Ghz |
 | 802.11ac | 2013 | 6.5 - 1000 | 5 Ghz |
 
-Today, new access points and computers support all a/b/g/n/ac standards, and most embedded devices offer only a/b/g/n capability. The device and the access point negotiate speed and standard automatically without requiring user configuration.
+Today, new access points and computers support all a/b/g/n/ac standards, and most embedded devices offer only a/b/g/n capability. The device and the access point negotiate speed and standard automatically without requiring user configuration.git check
```
melwee01:
```suggestion
Today, new access points and computers support all a/b/g/n/ac standards, and most embedded devices offer only a/b/g/n capability. The device and the access point negotiate speed and standard automatically without requiring a user configuration.git check.
```

docs/reference/runtime/static_memory_optimization.md
```diff
@@ -18,12 +18,12 @@ For a simple program like [Blinky](https://github.com/ARMmbed/mbed-os-example-bl
...
 
 #### `Printf` and UART
 
-The linker can also remove other modules that your program does not use. For example, [Blinky's](https://github.com/ARMmbed/mbed-os-example-blinky) `main` program doesn't use `printf` or UART drivers. However, every Mbed OS module handles traces and assertions by redirecting their error messages to `printf` on serial output - forcing the `printf` and UART drivers to be compiled in and requiring a large amount of flash memory.
+The linker can also remove other modules that your program does not use. For example, [Blinky's](https://github.com/ARMmbed/mbed-os-example-blinky) `main` program doesn't use `printf` or UART drivers. However, every Mbed OS module handles traces and assertions by redirecting their error messages to `printf` on serial output. This forces the `printf` and UART drivers to compile in and requires a large amount of flash memory.
```
AnotherButler: Query: Can we change "to compile in" to "to compile" and keep the meaning?

docs/reference/runtime/static_memory_optimization.md
```diff
@@ -40,15 +40,15 @@ To disable error logging to serial output, set the `NDEBUG` macro and the follow
 }
 ```
 
-<span class="notes">**Note:** Different compilers, different results; compiling with one compiler yields different memory usage savings than compiling with another.</span>
-
+<span class="notes">**Note:** Different compilers yield different results. Compiling with one compiler yields different memory usage savings than compiling with another.</span>
+<!---Snappy, but redundant--->
```
AnotherButler: What about "Note: Different compilers yield different memory savings."?

docs/reference/technology/USB.md
```diff
@@ -2,11 +2,11 @@
 
 ## Introduction
 
-A functional Mbed OS USB device consists of three parts - an implementation of USBPhy, the USBDevice stack and a USB component code:
+A functional Mbed OS USB device consists of three parts: an implementation of USBPhy, the USBDevice stack and a USB component code.
```
AnotherButler: We like to keep colons before bulleted lists, but I don't like two colons in a row. What do you suggest?

docs/reference/technology/USB.md
```diff
@@ -16,9 +16,9 @@ You can see the interaction of these three components in this diagram:
 
 ## Synchronization
 
-The class USBDevice is an interrupt-safe class. It uses a critical section to provide thread- and interrupt-safe locking. USB components inheriting from USBDevice can use this lock, but it is not required.
+The class *USBDevice* is an interrupt-safe class. It uses a critical section to provide thread- and interrupt-safe locking. USB components inheriting from USBDevice can use this lock, but it is not required.
```
AnotherButler: Why bold the class name?

docs/reference/technology/USB.md
```diff
@@ -46,7 +46,7 @@ void USBComponent::do_something_internal()
 
 ## USB device state
 
-USB defines 5 separate states a device can be in - Attached, Powered, Default, Address and Configured. Each state adds functionality. The Attached state has the least functionality, and the Configured state has the most functionality.
+USB defines five separate states of a device: **Attached**, **Powered**, **Default**, **Address** and **Configured**. Each state adds functionality. The **Attached** state has the least functionality, and the **Configured** state has the most functionality.
```
AnotherButler, line 49: I love the rewording, but why bold the states?

docs/reference/technology/USB.md
```diff
@@ -56,11 +56,11 @@ USB defines 5 separate states a device can be in - Attached, Powered, Default, A
...
-
+The USB device can enter a state with less functionality at any time. This could be due to a loss of power or surprise disconnect. When leaving or outside of the Configured state, USBDevice ignores writes to and reads from all endpoints other than endpoint 0.
+<!---initially confusing wording--->
 ## USB component callbacks
 
-All callbacks USBDevice sends to its children are prefixed with callback_*. USBDevice calls these callbacks with the USB lock held. One notable callback is `callback_state_change`, which USB components can use generically to handle leaving the Configured state. The USB stack automatically exits the Configured state on disconnect, power loss or USB reset.
+All callbacks that USBDevice sends to its children are prefixed with `callback_*`. USBDevice calls these callbacks with the USB lock held. One notable callback is `callback_state_change`, which USB components can use generically to handle leaving the Configured state. The USB stack automatically exits the Configured state on disconnect, power loss or USB reset.
```
AnotherButler: This "that" is not necessary. Good catch on the tick marks.

docs/reference/technology/USB.md
```diff
@@ -93,7 +93,7 @@ Below is a diagram showing the typical state machine for read (OUT) and write (I
 
 To ensure a USB component runs on all supported devices, the USB component must select the configuration descriptor's endpoints based on the current device. This is because endpoint number and endpoint functionality can differ by device. A USB component can determine the features of USBPhy by examining its endpoint table.
 
-To simplify the process of selecting endpoints, we recommend you use the EndpointResolver class. A USB component constructs the class with an endpoint table. The USB component can then call the class to find an endpoint of the given type and size. After the component finds all required endpoints, it can call the function `EndpointResolver::valid()` to determine whether this device supports this configuration. Below is an example of this:
+To simplify the process of selecting endpoints, use the *EndpointResolver* class. A USB component constructs the class with an endpoint table. The USB component can then call the class to find an endpoint of the given type and size. After the component finds all required endpoints, it can call the function `EndpointResolver::valid()` to determine whether this device supports this configuration. Below is an example of this:
```
AnotherButler: Why bold the class name?

docs/reference/technology/storage.md
```diff
@@ -1,16 +1,16 @@
 <h1 id="storage-tech">Storage</h1>
 
-We designed the architecture of the Mbed OS storage solution with the following properties:
+The architecture of the Mbed OS storage solution has the following properties:
```
AnotherButler, line 3: Much better

docs/reference/technology/storage.md
```diff
@@ -39,7 +39,7 @@ An example of implementations of the interfaces are:
 - FileHandle.
    - File - Class that provides read and write access to files on a file system.
    - BlockDeviceFile - Class that provides read and write access to raw block device storage in the form of a FileHandle.
-   - UARTSerial - The UARTSerial class provides read and write access to a device's serial hardware. It is an example of a nonstorage device that you can access as a file.
+   - UARTSerial - The UARTSerial class provides read and write access to a device's serial hardware. It is an example of a non-storage device that you can access as a file.
```
AnotherButler: I don't think this should be hyphenated. "Non" ends with an "n", and "storage" doesn't start with an "n". There's no proper noun or acronym there, either.

AnotherButler (commented):
Nice work 👍 

I left a few comments and queries.

docs/reference/technology/USB.md
melwee01: Typically do that for introducing key words and names.

docs/reference/technology/USB.md
melwee01: Editorial choice on my part; highlighting key words and newly introduced names.

docs/reference/technology/storage.md
melwee01: True, nonstorage just strikes me as... off.

docs/reference/technology/connectivity/connectivity.md
melwee01:
```suggestion
BLE is optimized for cheap, battery-based devices, and has a low memory footprint. It is widely adopted for connected spaces, but requires setting up an infrastructure. Its range is up to 100 meters, with 1 Mbps bandwidth. You need an Edge gateway so devices in your network can connect to the internet.
```

docs/reference/runtime/static_memory_optimization.md
melwee01: I think so.

melwee01:
```suggestion
The linker can also remove other modules that your program does not use. For example, [Blinky's](https://github.com/ARMmbed/mbed-os-example-blinky) `main` program doesn't use `printf` or UART drivers. However, every Mbed OS module handles traces and assertions by redirecting their error messages to `printf` on serial output. This forces the `printf` and UART drivers to compile and requires a large amount of flash memory.
```

docs/reference/technology/USB.md
melwee01:
```suggestion
The class **USBDevice** is interrupt-safe. It uses a critical section to provide thread- and interrupt-safe locking. USB components inheriting from USBDevice can use this lock, but it is not required.
```

docs/reference/technology/USB.md
melwee01:
```suggestion
To simplify the process of selecting endpoints, use the **EndpointResolver** class. A USB component constructs the class with an endpoint table. The USB component can then call the class to find an endpoint of the given type and size. After the component finds all required endpoints, it can call the function `EndpointResolver::valid()` to determine whether this device supports this configuration. Below is an example of this:
```

docs/reference/runtime/static_memory_optimization.md
```diff
@@ -49,6 +49,6 @@ You can also take advantage of the fact that these programs only run on embedded
 - The code that the compiler injects consumes memory.
 - It implies dynamic memory allocation and thus requires the binary to include `malloc`, even when the application does not use it.
 
-When you run an application on an embedded device, you don't need handlers to destroy objects when the program exits, because the application will never end. You can save more RAM and flash memory usage by [removing destructor registration](https://github.com/ARMmbed/mbed-os/pull/2745) on application startup and by eliminating the code to destruct objects when the operating system calls `exit()` at runtime. 
+When you run an application on an embedded device, you don't need handlers to destroy objects when the program exits, because the application will never end. You can save more RAM and flash memory usage by [removing destructor registration](https://github.com/ARMmbed/mbed-os/pull/2745) on application startup and by eliminating the code to destruct objects when the operating system calls `exit()` at runtime.
```
peknis01: You can save more... -> To save more RAM and... remove destructor registration... and eliminate code...

docs/reference/technology/connectivity/connectivity.md
```diff
@@ -13,66 +13,71 @@ For IP devices:
...
 
 <span class="images">![](https://s3-us-west-2.amazonaws.com/mbed-os-docs-images/ip-networking-simple.png)</span>
 
+<!---This could use a caption, IMO--->
 Our Socket API standardizes all of our connectivity options. It supports both IPv4 and IPv6. Applications are portable regardless of final connectivity option.
 
 Mbed OS provides network drivers, such as Ethernet, Wi-Fi and cellular.
 
-You can find descriptions of supported interfaces and protocols in the [IP networking](ip-networking.html) section.
+You can find descriptions of supported interfaces and protocols in the [IP networking](../reference/ip-networking.html) section.
```
peknis01: Descriptions of supported interfaces and protocols are in...

docs/reference/technology/connectivity/connectivity.md
```diff
@@ -98,21 +103,21 @@ The [LoRa](lora-tech.html) section and [LoRa tutorial](../tutorials/LoRa-tutoria
...
+Non-IP Data Delivery (NIDD) is a new feature for communication over NB-IoT. It is enabled by Control Plane cellular IoT EPS optimization, to provide improved support for small data transfer. It transports user data over the control channel, thus reducing the total number of control plane messages when handling a short data transaction.
 
 To learn how to use this feature with Mbed OS, please refer to [CellularNonIPSocket](../apis/non-ip-cellular-socket.html).
 
 ### Memory needs for Pelion-connected devices
 
-Mbed OS's baseline memory footprint, without Pelion connectivity, is 2.8kb of RAM and 8.2kb of flash. When you add connectivity and full functionality, the footprint grows:
+Mbed OS's baseline memory footprint, without Pelion connectivity, is 2.8 KB RAM and 8.2 KB flash. When you add connectivity and full functionality, the footprint grows:
```
peknis01: @teetak01, can you please check the numbers here.

teetak01: I cannot say on the Mbed OS numbers (which look to be unrealistically low, especially for flash).

For client, I would just link here: https://www.pelion.com/docs/device-management/current/cloud-requirements/requirements-for-production-devices.html

docs/reference/runtime/static_memory_optimization.md
melwee01:
```suggestion
When you run an application on an embedded device, you don't need handlers to destroy objects when the program exits, because the application will never end. To save more RAM and flash memory, [remove destructor registration](https://github.com/ARMmbed/mbed-os/pull/2745) on application startup and eliminate the code to destruct objects when the operating system calls `exit()` at runtime.
```

docs/reference/technology/connectivity/connectivity.md
melwee01:
```suggestion
Descriptions of supported interfaces and protocols are in [IP networking](../reference/ip-networking.html).
```

docs/reference/technology/connectivity/securesocket.md
```diff
@@ -20,7 +20,7 @@ socket->connect(HOST_NAME, PORT)
 socket->send(data, size);
 ```
 
-Please note that internal TLS structures require over 1 kB of RAM, so you need to allocate each TLSSocket from the heap by using the `new` command, instead of using stack or statically allocating it.
+Please note that internal TLS structures require over 1 KB of RAM, so you need to allocate each TLS Socket from the heap by using the `new` command, instead of using stack or statically allocating it.
```
melwee01:
```suggestion
<span class="notes">**Note:** Internal TLS structures require over 1 KB of RAM. Allocate each TLS Socket from the heap by using the `new` command, instead of using stack or statically allocating it.</span>
```

docs/reference/technology/connectivity/securesocket.md
```diff
@@ -102,7 +102,7 @@ This destroys the memory the TLS library allocates. It also closes the transport
 virtual nsapi_error_t connect(const SocketAddress &address);
 ```
 
-The code above initiates the TCP connection and continues to TLS hanshake. If [transport mode](#transport-modes) is either `TRANSPORT_KEEP` or `TRANSPORT_CLOSE`, TCP is assumed to be open and state directly goes into TLS handshake. This is currently forced to blocking mode. After succesfully connecting, you can set it to nonblocking mode:
+The code above initiates the TCP connection and continues to TLS handshake. If [transport mode](#transport-modes) is either `TRANSPORT_KEEP` or `TRANSPORT_CLOSE`, TCP is assumed to be open and state directly goes into TLS handshake. This is currently forced to blocking mode. After successfully connecting, you can set it to nonblocking mode:
```
melwee01:
```suggestion
The code above initiates the TCP connection and continues to the TLS handshake. If [transport mode](#transport-modes) is either `TRANSPORT_KEEP` or `TRANSPORT_CLOSE`, TCP is assumed to be open and state directly goes into TLS handshake. This is currently forced to blocking mode. After connecting, you can set it to nonblocking mode:
```

docs/reference/technology/firmware_update.md
```diff
@@ -14,17 +14,17 @@ The active firmware, made up of Mbed OS and an application, can be updated only
...
 1. Mbed OS with the Device Management Client (which includes the Update client). The clients allow your device to communicate with the Device Management Update service, receive update manifests and firmware, and verify the binary's validity. They are [reviewed in details in our Pelion Device Management documentation](https://cloud.mbed.com/docs/latest/updating-firmware/index.html).
 
-    Note: Because some embedded devices don't require remote update capabilities, Mbed OS does not include the Device Management Client by default; you need to explicitly import the client to your application.
+  <span class="notes">**Note:** Because some embedded devices don't require remote update capabilities, Mbed OS does not include the Device Management Client by default; you need to explicitly import the client to your application.</span>
```
melwee01:
```suggestion
  <span class="notes">**Note:** Because some embedded devices don't require remote update capabilities, Mbed OS does not include the Device Management Client by default. You must explicitly import the client to your application.</span>
```

docs/reference/technology/firmware_update.md
```diff
@@ -82,14 +82,14 @@ When a device downloads new firmware, it stores it locally (in the storage area)
...
 
-<span class="notes">**Note**: The default workflow of the development tools is intended only for development and testing purposes. It is not considered secure for production.</span>
+<span class="notes">**Note**: The default workflow of the development tools is intended only for development and testing purposes. It is not secure for production.</span>
 
-Your development tool needs to use your Device Management account's API key to call the Pelion Device Management APIs. Once it has access to the APIs, it can generate a manifest and upload the manifest and binary to the server, and deliver the manifest to a device or group of devices. For more information, see [the update API documentation](https://cloud.mbed.com/docs/latest/service-api-references/update-service.html).
+Your development tool needs to use your Device Management account's API key to call the Device Management APIs. Once it has access to the APIs, it can generate a manifest and upload the manifest and binary to the server, and deliver the manifest to a device or group of devices. For more information, see [the update API documentation](https://cloud.mbed.com/docs/latest/service-api-references/update-service.html).
```
melwee01:
```suggestion
Your development tool needs to use your Device Management account API key to call the APIs. Once it has access to the APIs, it can generate a manifest, upload the manifest and binary to the server, and deliver the manifest to a device or group of devices. For more information, see [the update API documentation](https://cloud.mbed.com/docs/latest/service-api-references/update-service.html).
```

docs/reference/technology/storage.md
```diff
@@ -26,7 +26,7 @@ Interfaces from low-level to high-level:
 An example of implementations of the interfaces are:
 
 - BlockDevice.
-   - QSPIBlockDevice - Block device driver for NOR-based QSPI Flash devices that support the SFDP standard. 
+   - QSPIBlockDevice - Block device driver for NOR-based QSPI Flash devices that support the SFDP standard.
```
melwee01, line 29: I don't like these dashes. Would rather have two layers of : than inconsistency.

docs/reference/technology/storage.md
melwee01:
```suggestion
   - UARTSerial - The UARTSerial class provides read and write access to a device's serial hardware. It is an example of a nonstorage device that you can access as a file.
```

docs/reference/technology/technology.md
```diff
@@ -1,3 +1,5 @@
 # Technology
 
 This section of the book talks about technologies that Mbed OS implements or supports. The content in this section details the architecture and protocols, design decisions and historical context of these technologies.
+
+<!---Should we have links to the subsections here?--->
```
melwee01: This section is quite empty, and I'd rather combine it with an index of the pages below.

docs/reference/technology/USB.md
melwee01:
```suggestion
All callbacks USBDevice sends to its children are prefixed with `callback_*`. USBDevice calls these callbacks with the USB lock held. One notable callback is `callback_state_change`, which USB components can use generically to handle leaving the Configured state. The USB stack automatically exits the Configured state on disconnect, power loss or USB reset.
```

docs/reference/technology/USB.md
melwee01: I changed this so the double colon was no longer necessary, but there's another one below that I can't do that kind of restructure with.

docs/reference/technology/technology.md
AnotherButler: This page needs a lot of help.

docs/reference/technology/storage.md
AnotherButler: I don't know that I agree.

docs/reference/technology/firmware_update.md
```diff
@@ -47,14 +47,14 @@ The Mbed tools (Mbed CLI, Online Compiler) can manage bootloaders where:
...
-If your bootloader does not meet the two requirements of a manageable bootloader, you will need an unmanaged bootloader. The Mbed tools will not automatically combine this bootloader with the application image; you will need to write your own scripts to build your full image.
+If your bootloader does not meet the two requirements of a manageable bootloader, you will need an unmanaged bootloader. The Mbed tools do not automatically combine this bootloader with the application image; you need to write your own scripts to build your full image.
 
 For more information about [managed and unmanaged bootloaders, and how to create bootloaders, see the tutorial](../tutorials/bootloader.html).
 
 ### Porting a bootloader
 
 If you're interested in porting the Update client and bootloader to new hardware, please [review the porting section](../porting/bootloader.html).
-
+<!---Are there really two of these guides, or should we just have one link to the bootloader cloud-side docs in the porting section, as below?--->
```
AnotherButler: Because we're specifically talking about porting the Update client and bootloader, I'm fine linking to the cloud-side doc here.

docs/reference/technology/USB.md
```diff
@@ -56,11 +56,11 @@ USB defines 5 separate states a device can be in - Attached, Powered, Default, A
 | Address    |    No new functionality                  |
 | Configured |    All enabled endpoints are functional  |
 
-At any time, the USB device can enter a state with less functionality. This could be due to a loss of power event or a surprise USB disconnect. When leaving or outside of the Configured state, USBDevice ignores writes to and reads from all endpoints other than endpoint 0.
-
+The USB device can enter a state with less functionality at any time. This could be due to a loss of power or surprise disconnect. When leaving or outside of the Configured state, USBDevice ignores writes to and reads from all endpoints other than endpoint 0.
```
AnotherButler: I'm still not sure what this means. Does it mean it can turn off or parts of it can stop working at any time? Is this specific to USB?

docs/reference/technology/connectivity/connectivity.md
```diff
@@ -13,66 +13,71 @@ For IP devices:
...
-There is a wide variety of possibilities for connected devices, and there is no globally correct connectivity method. Choose a method that fits your application needs, then choose a device that can support that method. If you're already committed to a device type, you may need to adjust your application to work within that device's connectivity constraints.
+There is a wide variety of possibilities for connected devices, and no globally correct connectivity method. Choose a method that fits your application needs, then choose a device that can support that method. If you're already committed to a device type, you may need to adjust your application to work within that device's connectivity constraints.
 
 ## Internet protocol (IP) networking
 
 Mbed OS supports various IP-based connectivity options and IP stacks.
 
 <span class="images">![](https://s3-us-west-2.amazonaws.com/mbed-os-docs-images/ip-networking-simple.png)</span>
 
+<!---This could use a caption, IMO--->
```
AnotherButler: What do you think about "Connectivity stack architecture in Mbed OS"?

docs/reference/technology/connectivity/wifi.md
```diff
@@ -33,6 +33,6 @@ As with Ethernet, Wi-Fi networks are only used for carrying IP traffic in Mbed O
 For more information, please see the following sections:
 
 - [IP networking](ip-networking.html).
-- [Ethernet](ethernet-technology.html).
+- [Ethernet](ethernet-technology.html). <!---Can links be like this, or do they need ../reference/?--->
```
AnotherButler: If they're in the same book, they work this way. If the inconsistency really bothers anyone, we can change it.

docs/reference/technology/firmware_update.md
```diff
@@ -82,14 +82,14 @@ When a device downloads new firmware, it stores it locally (in the storage area)
...
+<span class="notes">**Note**: The default workflow of the development tools is intended only for development and testing purposes. It is not secure for production.</span>
 
-Your development tool needs to use your Device Management account's API key to call the Pelion Device Management APIs. Once it has access to the APIs, it can generate a manifest and upload the manifest and binary to the server, and deliver the manifest to a device or group of devices. For more information, see [the update API documentation](https://cloud.mbed.com/docs/latest/service-api-references/update-service.html).
+Your development tool needs to use your Device Management account API key to call the APIs. Once it has access to the APIs, it can generate a manifest, upload the manifest and binary to the server, and deliver the manifest to a device or group of devices. For more information, see [the update API documentation](https://cloud.mbed.com/docs/latest/service-api-references/update-service.html).
 
 <span class="notes">**Note**: The tools currently support the update flow for Device Management Client, not Device Management Client Lite.</span>
-
+<!---Is this still true?--->
```
AnotherButler: Who can answer this?

docs/reference/technology/connectivity/connectivity.md
```diff
@@ -13,66 +13,71 @@ For IP devices:
...
-Non-IP devices:
+<!--- Order and headings do not reflect what're in the lists below--->
+Non-IP devices require a gateway:
 
 - LoRaWAN.
 - Cellular.
 
+<!--- What about NFC?--->
+
+<!---What if we move this to the ../technology.md file?--->
```
AnotherButler: We separated networking architecture docs from non-networking architecture docs. The technology.md file is for both. This content is networking-specific.

AnotherButler (commented):
I've replied to the remaining queries (usually with more questions).

docs/reference/technology/USB.md
melwee01: Looking at it again, I think we're talking more *random* occurrences of power loss or disconnects, rather than 'at any time', which seems to imply it's a specific fault of USB.

melwee01:
```suggestion
Random power loss or disconnect can cause the USB device to lose functionality. When leaving or outside of the Configured state, USBDevice ignores writes to and reads from all endpoints other than endpoint 0.
```

melwee01: Does that communicate the idea more clearly?

docs/reference/technology/connectivity/connectivity.md
melwee01: Sure.

docs/reference/technology/firmware_update.md
melwee01: @yogpan01 or @teetak01, perhaps?

teetak01: yes, support is still missing.

docs/reference/technology/technology.md
AnotherButler: I'm going to remove this comment for now. We can address this page in another PR.

AnotherButler (approved):
Thanks for the PR and all of the responses to our questions. 👍

PR #1057, opened 2019-05-05
Fix AC6 path on Windows

arekzaluski:
## Changes

- Remove extra dot in path to AC6


GuyWi (approved):

iriark01:
This needs to be cheery picked into 5.12 as well, if it hasn't been

GuyWi:
I just checked. 5.12 is good. 👍

PR #1060, opened 2019-05-07
Erase function for SDBlockDevice

iriark01:
For IOTDOCS-953


AnotherButler (approved):

linlingao (changes requested):
Instead of saying "erase doesn't work", which implies it attempts to do something but fails, I would say "erase is not applicable on SDBlock device, hence it's a no-op operation". This makes it clear this API doesn't do anything at all.

linlingao (approved):

PR #1063, opened 2019-05-10
Add semihosting and SWO examples to mbed_override_console().

korjaa:
Related to https://github.com/ARMmbed/mbed-os/pull/10553

### DIFF
--- a/docs/api/platform/FileHandle.md
+++ b/docs/api/platform/FileHandle.md
@@ -69,14 +69,15 @@ If a target has serial support, by default a serial port is used for the console
 The target can override this by providing `mbed::mbed_target_override_console` to specify an alternative `FileHandle`. For example, a target using SWO might have:
 
 ```
-    namespace mbed
+namespace mbed
+{
+    FileHandle *mbed_target_override_console(int)
     {
-        FileHandle *mbed_target_override_console(int)
-        {
-            static SerialWireOutput swo;
-            return &swo;
-        }
+        // SerialWireOutput
+        static SerialWireOutput swo;
+        return &swo;
     }
+}
 ```
 
 Then any program using `printf` on that target sends its output over the SWO, rather than serial.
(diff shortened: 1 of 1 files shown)


korjaa:
@TeroJaasko semihosting example in docs.

docs/api/platform/FileHandle.md
```diff
@@ -69,17 +69,50 @@ If a target has serial support, by default a serial port is used for the console
...
+    FileHandle *mbed_target_override_console(int)
+    {
+        // SerialWireOutput
+        static SerialWireOutput swo;
+        return &swo;
+    }
+}
+```
+
+Then any program using `printf` on that target sends its output over the SWO, rather than serial. It is also possible to redirect the console by using semihosting which enables the console logs to be captured by a debugger:
```
kjbracey: The above example's text was talking about the case where a *target* is using SerialWireOutput.

Both the new examples are probably/definitely the case where the *developer/application* is deciding to redirect. (The semihosting would crash without debugger attached, so it's probably only a temporary thing by the developer, and the other one is talking about the target default not being suitable).

So those two should be overriding `mbed_override_console`, and/or maybe the text should be shuffled a bit. Otherwise you'd have a horrible collision in the case where the target had the `SerialWireOutput` override above, and you tried to override to semihosting.

In an out-of-the-box Mbed OS, the WEAK `target` defaults may or may not be already overridden by the target's support code - they're reserved for target code.

The WEAK non-`target` defaults are always ready to be overridden by application code on any target. Or some libraries may offer an option to do the override for you - eg "esp8266.provide_default". 

We could have `platform.localfilesystem.provide_default_console` or something to provide this example behaviour.

korjaa: Good good, I'll update this.

0xc0170: was this updated?

korjaa: Not yet.

korjaa: Now it makes sense to me, I was initially using the wrong `mbed_target_override_console()` in my `main.cpp` . I juggled the text around to better support this new understanding.

@kjbracey-arm any comments?

korjaa:
~~Please don't merge this until I rebase the fixup commits. I'll rebase after the changes are approved.~~
Nvm, decided to rebase already.

kjbracey (approved):

AnotherButler:
Is this relevant to both 5.12 and 5.13?

korjaa:
`mbed_console_override()` was introduced in https://github.com/ARMmbed/mbed-os/commit/f79bfddcad7ec5891f6e13cb919a712cf42ebdc3 (tools-release-5.8.5) so if I'm not mistaken, this is relevant starting from 5.8.5.

PR #1068, opened 2019-05-10
Mbed CLI installation steps refactor

bridadan:
Addresses part of IOTDOCS-778. This is mainly addressing the installation page and tried to make it simpler with less duplication.

### DIFF
--- a/docs/tools/CLI/cli-setup/cli-reqs.md
+++ b/docs/tools/CLI/cli-setup/cli-reqs.md
@@ -151,7 +151,7 @@ To install `mbed-cli` bash completion:
 1. Clone the Mbed CLI repository: `git clone https://github.com/ARMmbed/mbed-cli`.
 1. Navigate to the `mbed-cli/tools/bash_completion` directory.
 1. Copy the `mbed` script into the  `~/.bash_completion.d` directory (you may need to create this directory first).
-1. Restart the terminal
+1. Restart the terminal.
 
 # Configuration options
 
--- a/docs/tools/CLI/cli-setup/cli-reqs.md
+++ b/docs/tools/CLI/cli-setup/cli-reqs.md
@@ -5,7 +5,7 @@ The installers are a great way to get started with Mbed OS quickly if you are us
 - For Windows, download the installer [here](https://github.com/ARMmbed/mbed-cli-windows-installer/releases/latest).
 - For macOS, download the installer [here](https://github.com/ARMmbed/mbed-cli-osx-installer/releases/latest).
 
-<span class="notes">**Note:** The installer installs the GNU Arm embedded toolchain (GCC). If you want to compile using the Arm Compiler or IAR, visit the [supported compilers page](../tools/index.html#compiler-versions).</span>
+<span class="notes">**Note:** The GNU Arm embedded toolchain (GCC) is bundled with the installers. If you want to compile using the Arm Compiler or IAR, visit the [supported compilers page](../tools/index.html#compiler-versions).</span>
 
 # Manual installation
 
--- a/docs/tools/CLI/cli-setup/cli-reqs.md
+++ b/docs/tools/CLI/cli-setup/cli-reqs.md
@@ -11,33 +11,13 @@ There is no installer for Linux; please follow the manual installation guide.
 
 # Manual installation
 
-## Dependencies
-
-- [Python](https://www.python.org/) - we support versions 2.7.x, 3.6.x, and 3.7.x.
-
-    We recommend using Python 2, since [Python 3 support for Mercurial is still in the early stages](https://www.mercurial-scm.org/wiki/Python3).
-
-    <span class="notes">Python 3 is not supported for Mbed OS versions earlier than 5.9.0.</span>
-
-- [pip](https://pip.pypa.io/en/stable/)
-
-    <span class="notes">**Note:** Mbed CLI versions 1.5.0 and earlier are **not compatible** with `pip` version 10.0 (or newer). Please use Mbed CLI versions 1.5.1 or later with recent versions of `pip`.</span>
-
-- [Git](https://git-scm.com/) - versions 1.9.5 or later are supported.
-
-- [Mercurial](https://www.mercurial-scm.org/) - versions 2.2.2 or later are supported.
-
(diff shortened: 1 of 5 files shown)


docs/tools/CLI/cli-setup/cli-reqs.md
```diff
@@ -84,39 +91,49 @@ $ sudo apt-get install libc6:i386 libncurses5:i386 libstdc++6:i386
...
+1. Mbed CLI config (local first, then global).
+1. Environment variables.
+1. System path.
+
+<!--the order in the list is the same as the original text, but the content has been changed so that PATH is first rather than last. Why?-->
 
 ### Install compilers in the system PATH
 
-Installing the compilers in the system PATH is most often the easiest method, as the compiler installation process will often place the executables in your system's PATH automatically.
+Installing the compilers in the system PATH is most often the easiest method, as the compiler installation process will often place the executables in your system's PATH automatically.<!--often? what does it depend on?-->
```
bridadan: It is usually an option within the installer weather the compiler is placed in the PATH or not. I am not sure what the default is, I'd need to check that.

docs/tools/CLI/cli-setup/cli-reqs.md
```diff
@@ -84,39 +91,49 @@ $ sudo apt-get install libc6:i386 libncurses5:i386 libstdc++6:i386
...
+
+<!--what was wrong with the original text
```
bridadan: I suppose it's personal preference. I typically don't care for sentences that start with "This means". I find having to explain a previous sentence (that isn't a directly quoting something) usually means the original sentence could be clearer.

docs/tools/CLI/cli-setup/cli-reqs.md
```diff
@@ -84,39 +91,49 @@ $ sudo apt-get install libc6:i386 libncurses5:i386 libstdc++6:i386
...
 
 ```
 $ mbed config -G ARM_PATH "C:\Program Files\ARM"
 [mbed] C:\Program Files\ARM now set as global ARM_PATH
 ```
 
-Use the `-G` argument to se the compiler location globally. Omit the `-G` argument to use this compiler location for a specific project.
+The `-G` argument sets the location globally, so that it applies to all projects. You can omit the argument if you are configuring a compiler location for a specific project.
+
+<!--Each toolchain requires a specific directory to be used when setting the location:--> <!--I don't think this sentence improves on the original. What it implies is "each toolchain has one universally correct path no all computers everywhere" which we know isn't true. The original version was "Mbed CLI supports a setting for each toolchain path:" which isn't great, but is far less confusing.-->
```
bridadan: Fair enough!

docs/tools/CLI/cli-setup/cli-reqs.md
```diff
@@ -84,39 +91,49 @@ $ sudo apt-get install libc6:i386 libncurses5:i386 libstdc++6:i386
...
 
-Each toolchain requires a specific directory to be used when setting the location:
+Each toolchain has its own setting name, and must have a path to <!--where? some of the example paths are down to `bin`, some are to a different level. what's the rule?-->
```
bridadan: There isn't really much a rule unfortunately. The toolchain paths were not implemented consistently initially so we're left with this inconsistent paths. I tried to make the settings less picky but it turned out to be more work than I had for at the time.

bridadan (commented):
Edits seem ok to me!

docs/tools/CLI/cli-setup/cli-reqs.md
```diff
@@ -153,28 +153,6 @@ To install `mbed-cli` bash completion:
...
-
-1. Set the environment variable `ARMLMD_LICENSE_FILE`:
-
-    For Mbed Studio 0.5 and later, the Arm Compiler 6 license file, `ac6-license.dat`, is located in the following path:
-
-    - Windows: `C:\Users\<user>\AppData\Local\Mbed Studio\mbed-studio-tools`
-    - Mac: `/Library/Application Support/Mbed Studio/mbed-studio-tools`
-
-    Note that this environment variable is not managed by Mbed CLI, so you must set it appropriately in your environment.
-
```
bridadan: Any reason we're pulling this out? During the 5.12 release this particular section was explicitly asked for.

iriark01: It's not supported, and will break soon if it hasn't already. Someone (I think Andy P) said it's okay to just pull it out.

iriark01:
Hey @bridadan , can we work on the open questions on this today?

docs/tools/CLI/cli-setup/cli-reqs.md
```diff
@@ -1,236 +1,139 @@
...
-Mbed CLI may not work properly without some manual configuration.
+To build project, Mbed CLI needs to know where the compiler is installed. It checks three configuration options in the following order:
 
-## Mandatory: setting PATH variables
+1. Mbed CLI config (local first, then global).
+1. Environment variables.
+1. System path.
 
-Add the following to the system `PATH`:
+<!--the order in the list is the same as the original text, but the content has been changed so that PATH is first rather than last. Why?-->
```
bridadan: I've changed the order because I think installing compilers to the PATH is the most common scenario that people will use. It also doesn't require any additional Mbed CLI commands.

docs/tools/CLI/cli-setup/cli-reqs.md
bridadan: Well crud, it isn't the default. I will switch the order back.

docs/tools/CLI/cli-setup/cli-reqs.md
bridadan: Ah I was mistaken, I thought installing it to the PATH was on by default for GCC atleast, turns out this isn't the case. I will switch the order back.

bridadan:
Ok, think I've addressed everything.

AnotherButler:
I see no further comments that need to be addressed.

@bridadan Thanks for all the help 👍 Is this OK to merge to development and cherry pick to 5.13?

bridadan:
I have no reservations! The people to really check with would be @madchutney and @mark-edgeworth.

madchutney (approved):
I see no need to second guess @bridadan, good to go in my opinion.

AnotherButler (approved):

PR #1086, opened 2019-06-13
Fix description of DataFlashBlockDevice

geky:
Also crosslinked SPIFBlockDevice <-> DataFlashBlockDevice because this is a common pain point for users

cc @AnotherButler, @bulislaw


bulislaw (approved):

AnotherButler (approved):
Thanks for the PR 👍

PR #1088, opened 2019-06-14
Relax watchdog timing requirements

fkjagodzinski:
Update the docs to reflect the new timing requirements for watchdog (from https://github.com/ARMmbed/mbed-os/issues/8857#issuecomment-467838213):
* The watchdog should trigger at or after the timeout value.
* The watchdog should trigger before twice the timeout value.


AnotherButler (approved):
Thanks for the PR 👍

fkjagodzinski:
I can see @AnotherButler has already marked this with `code dependency`. :+1:

This has to wait until https://github.com/ARMmbed/mbed-os/pull/10839 is merged in Mbed OS repo.

bulislaw (approved):

PR #1090, opened 2019-06-14
Add C++14 content to common_issues.md

AnotherButler:
Add section about C++14 mode being the default and the implications of that.

### DIFF
--- a/docs/tutorials/debug/common_issues.md
+++ b/docs/tutorials/debug/common_issues.md
@@ -22,7 +22,7 @@ If you are using an old version of Mbed CLI, you may see compile-time errors. Ma
 
 Mbed OS 5 can be built with various toolchains. Make sure you are using the latest versions of the toolchains. You can find the currently supported versions on [our tools page](../tools/index.html).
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/debug/common_issues.md
```diff
@@ -22,6 +22,30 @@ If you are using an old version of Mbed CLI, you may see compile-time errors. Ma
 
 Mbed OS 5 can be built with various toolchains. Make sure you are using the latest versions of the toolchains. You can find the currently supported versions on [our tools page](../tools/index.html).
 
+The default profiles for the GCC and Arm toolchains now select C++14 and C11. Some applications may fail to compile because they use constructs that were valid in C++98 but not in C++14. This may also apply to your applications if you are migrating from older Mbed OS versions that used IAR 7. To fix these common compatibilitiy issues:
```
kjbracey: If this is general info, rather than a "Mbed OS 5.13 release note", then the IAR distinction isn't really so important - all toolchains are now C++14 and C11. IAR switched in 5.12, the other two in 5.13.

I'd simplify it down to:

> The default toolchain profiles now select C++14 and C11. Some applications that worked with older versions of Mbed OS may fail to compile because they use constructs that were valid in C++98 but not in C++14. To fix these common compatibility issues:

("compatibilitiy" -> "compatibility")

kjbracey (approved):

PR #1091, opened 2019-06-14
Remove Arm Compiler 5 from style.md

AnotherButler:
Remove Arm Compiler 5 requirement.

### DIFF
--- a/docs/reference/contributing/guidelines/style.md
+++ b/docs/reference/contributing/guidelines/style.md
@@ -169,10 +169,10 @@ All functions and methods should contain documentation using Doxygen.
 
 ## Compiler settings
 
-All C and C++ code submitted to Mbed OS must compile with GCC Arm Embedded and IAR EWARM. Mbed OS:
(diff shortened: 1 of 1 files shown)


AnotherButler:
@kjbracey-arm @bulislaw Should this requirement be removed now?

bulislaw (changes requested):
We need to replace it with 6. Also @kjbracey-arm can you update the C/C++ standards below to match your recent changes?

kjbracey:
Yes, next bit should be

    - Uses the GNU11 standard for C.
    - Uses the GNU++14 standard for C++.

But that's assuming we *are* outright dropping the "must compile for ARM Compiler 5" . If that remains, ARM C 5 selects GNU99/GNU++11, and its "GNU++11" is only a subset (about 70% of language additions, no library).

I'm currently still uncertain whether I'm okay to push something to master that will not work in ARM C 5's C++11 subset. Changing these docs would give me permission, so that's a higher-level-than-me decision. Personally, I'm in favour.

AnotherButler:
@bulislaw Is this OK to merge, or should we talk about this in person?

bulislaw (approved):

PR #1094, opened 2019-06-24
Review of Device Management flow for CLI

melwee01:
https://jira.arm.com/browse/IOTDOCS-1079


docs/tools/CLI/cli-device-management.md
```diff
@@ -33,6 +33,7 @@ $ mbed dm init -d "<company domain name>" --model-name "<product model identifie
 <span class="notes">**Note:** If you do not want to enter the subject information for your update certificate (country, state, city, organization and so on), add the `-q` flag to the command above.</span>
 
 This command asks for information about your update certificate. After completing the prompts, Mbed CLI creates several files:
+>This is actually more detailed and nicer than the information we have in the DM docs. Can we steal it?
```
AnotherButler: Please do. :)

docs/tools/CLI/cli-device-management.md
```diff
@@ -44,8 +45,8 @@ This command asks for information about your update certificate. After completin
    - The path of the update certificate and private key.
 - Device Management developer credentials in `mbed_cloud_dev_credentials.c`
 
-<span class="notes">**Note:** The certificate created in `mbed dm init` is not suitable for production. Use it for testing and development only. To create a certificate for production purposes, use an air-gapped computer or a Hardware Security Module. When going to production, conduct a security review on your manifest signing infrastructure because it is the core of the security guarantees for update client.</span>
-
+<span class="notes">**Note:** The certificate created in `mbed dm init` is not suitable for production. Use it for testing and development only. To create a certificate for production purposes, use an air-gapped computer or a Hardware Security Module. When going to production, conduct a security review on your manifest signing infrastructure because it is the core of the security guarantees for Update client.</span>
+<!--- All of this is on the DM site. Do we link out?
```
AnotherButler: I would think so. @iriark01 , what are your thoughts?

docs/tools/CLI/cli-device-management.md
```diff
@@ -55,7 +56,7 @@ $ mbed compile
 ```
 
 This generates a payload to update the device with. After generating the payload, update the device through Device Management with:
-
+<!---We talk about an image rather than a payload.
```
AnotherButler: Because this is the only place we talk about "payload" (at least in Mbed OS), I think we should standardize to match Device Management. @iriark01 , what do you think?

docs/tools/CLI/cli-device-management.md
iriark01: No harm in pointing people at more info, but don't eliminate the paragraph - they may not care and this info is enough.

docs/tools/CLI/cli-device-management.md
iriark01: Is payload clearer? Although I guess that ship's sailed, on the DM side

melwee01: This might be one of those instances where both are in the code. I'll poke around.

melwee01: Yes--payload is in the SDKs and various references throughout the DM docs, such as advanced Client config; it seems to refer more to the size of a block of information that you transmit, rather than the firmware *image* as such. Though there looks to be enough overlap that we still have competing terms for the same thing.

AnotherButler: This sounds like something we should confirm with engineering.

melwee01: @yogpan01, do you have an opinion on this?

AnotherButler:
@iriark01 This should probably be revisited.

iriark01:
@melwee01 I want to replace this with links to the DM docs. Where should it point?

melwee01:
https://www.pelion.com/docs/device-management/current/connecting/mbed-os.html This is the CLI material, with a couple of sentences about Update at the bottom.

iriark01:
I'm curious as to why this PR wasn't merged. Did engineering disagree with any of it?

iriark01:
@melwee01 why was this PR not merged? Are there issues with the content?

melwee01:
Not that I'm aware, but once the Update restructure goes in, a few links might change. I'll make a note to check that.

iriark01:
Okay. I'm closing this, as I'm working on DM content in another PR.

PR #1118, opened 2019-08-01
Add redirects validation step to the build

muchzill4:


micque01 (approved):
Don't forget to do the Pelion docs also 😜

PR #1131, opened 2019-08-12
Update build recognition list in mbed_targets.md

AnotherButler:
Update list the build recognizes, as requested.


0xc0170 (changes requested):
What about `bootloader` feature, should it be in the list?

`lwip` feature was deprecated, should not be listed anymore

Isn't storage feature the one deprecated and we provide storage always?

cc @SeppoTakalo

SeppoTakalo:
This is the up to date list: https://github.com/ARMmbed/mbed-os/blob/master/tools/config/__init__.py#L72

From that list, `NANOSTACK` and `LWIP` are left there just to produce deprecation warning and not break the build. See: https://github.com/ARMmbed/mbed-os/tree/master/features/deprecated_warnings

`FEATURE_STORAGE` seems to have disappeared from some point. It doesn't do anything, so should be removed from list.


So, the list should be:
* BOOTLOADER
* BLE
* CRYPTOCELL310

AnotherButler:
Thanks for the help, @SeppoTakalo and @0xc0170  👍 Does that look better?

0xc0170 (approved):

Issue #1139, opened 2019-08-20
Specify full directory path on "Working with mbed CLI" user guide 


TheNageek:
# Full path to MBED_OS_DIR not specific in guide
Minor clarification, but when attempting to follow the guide, at the link below, specifically for managing multiple projects, could clarify that the path to the projects directory should be the full path, not just the immediate path when setting MBED_OS_DIR global, example below

`/home/user/workspace/mbed-os`

NOT

`workspace/mbed-os`

https://os.mbed.com/docs/mbed-os/v5.13/tools/working-with-mbed-cli.html

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTDOCS-1208*

AnotherButler:
@CuriousEngineering Thanks for raising this issue 👍 

If you're interested in contributing a fix to this or any other documentation issue, our documentation contributions follow the [same guidelines as our code](https://os.mbed.com/docs/mbed-os/v5.13/contributing/index.html) (as applicable).

AnotherButler:
We have merged PR #1141 to resolve this issue.

PR #1141, opened 2019-08-21
#1139: Improving description of MBED_OS_DIR config option

mark-edgeworth:
Add some text to make it clear that this config option should be an absolute path.
This should fix the issue raised in #1139


AnotherButler (approved):
Thanks so much for the PR 👍

PR #1144, opened 2019-08-23
add docs for integration tests

jamesbeyond:
This the docs for integration tests.
ARMmbed/mbed-os#11276

### DIFF
--- a/docs/porting/porting_full_process/integration tests.md
+++ b/docs/porting/porting_full_process/integration tests.md
@@ -1,6 +1,6 @@
 # Integration tests
 
-Mbed OS provides a set of integration tests to help you identify integration issues around storage devices and connectivity devices.
+Mbed OS provides a set of integration tests to help you identify integration issues around storage devices and connectivity devices. If you're using a custom board, these tests can also help you verify your board works well in your environment.
 
 ## Test suites
 
@@ -22,7 +22,7 @@ If a platform has both of them enabled by default (in `targets.json`), then you 
     $ mbed test -t <toolchain> -m <platform> -n *integration-* -DINTEGRATION_TESTS -v
     ```
     
-If a platform doesn't have default storage or connectivity device and may require extra shield to run the integration test. In this case, you can pass a `target_extended.json` file in the command-line to configure the target:
+If a board doesn't have default connectivity and storage, it requires an extra shield to run the integration test. In this case, you can pass a `target_extended.json` file in the command-line to configure the target:
 
    ```
    $ mbed test -t <toolchain> -m <platform> -n *integration-* -DINTEGRATION_TESTS --app-config TESTS\integration\COMMON\target_extended.json -v
@@ -50,7 +50,7 @@ If required, edit the `target_extended.json` file, and create a new entry under 
             "target.components_add" : ["SD"],
    ```
 
-   Valid options are `SD`, `SPIF`, `QSPIF` and `FLASHIAP` (not recommended). For more available options, please see the [block device components](https://github.com/ARMmbed/mbed-os/tree/master/components/storage/blockdevice).
+   Valid options are `SD`, `SPIF` and `QSPIF`. For more available options, please see the [block device components](https://github.com/ARMmbed/mbed-os/tree/master/components/storage/blockdevice).
 
    You also have to specify the block device pin configuration, which may vary from one block device type to another. Here's an example for `SD`:
    
--- a/docs/porting/porting_full_process/integration tests.md
+++ b/docs/porting/porting_full_process/integration tests.md
@@ -1,9 +1,6 @@
-# Integration tests for Mbed OS
+# Integration tests
 
-
-
-
-Mbed OS provides a set of integration tests to help users identify integration issues around storage device and connectivity device.
+Mbed OS provides a set of integration tests to help you identify integration issues around storage devices and connectivity devices.
 
 ## Test suites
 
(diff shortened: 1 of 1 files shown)


docs/porting/porting_full_process/integration tests.md
```diff
@@ -0,0 +1,65 @@
+# Integration tests
+
+Mbed OS provides a set of integration tests to help you identify integration issues around storage devices and connectivity devices.
```
AnotherButler: This is only for people porting, right?

docs/porting/porting_full_process/integration tests.md
```diff
@@ -0,0 +1,65 @@
...
+
+The integration test requires you to configure both the connectivity and storage devices.
+
+If a platform has both of them enabled by default (in `targets.json`), then you can compile the tests and run them from inside the `mbed-os` directory: 
+
+    ```
+    $ mbed test -t <toolchain> -m <platform> -n *integration-* -DINTEGRATION_TESTS -v
+    ```
+    
+If a platform doesn't have default storage or connectivity device and may require extra shield to run the integration test. In this case, you can pass a `target_extended.json` file in the command-line to configure the target:
```
AnotherButler: Does this mean if a platform requires an extra shield, you have have to pass `target_extended.json` in the command-line? Or that if it doesn't have storage and connectivity by default, it requires an extra shield?

docs/porting/porting_full_process/integration tests.md
```diff
@@ -0,0 +1,65 @@
...
+
+   ```
+   $ mbed test -t <toolchain> -m <platform> -n *integration-* -DINTEGRATION_TESTS --app-config TESTS\integration\COMMON\target_extended.json -v
+   ```
+   
+A preconfiguration file is defined in the COMMON folder, and that file covers some of the platform connectivity and storage combinations. However, this preconfiguration file might contain some compoments drivers not included in `mbed-os` (for example, `WIFI_WIZFI310` and `NUSD`). You must add those drivers manually. If you have your own platform or want to define your own configuration, please follow the [configuration guide](#tests-configuration).
```
AnotherButler, line 31: When would it contain drivers not included in `mbed-os`? To add them manually, do they just follow the normal porting targets sections?

docs/porting/porting_full_process/integration tests.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Tests configuration
+
+If required, edit the `target_extended.json` file, and create a new entry under `target_overrides` with the target name for your device:
+
+- **Connectivity** - Specify the default connectivity type for your target. It's essential with targets that lack default connectivity set in `targets.json` or for targets that support multiple connectivity options. For example:
```
AnotherButler, line 37: Is it optional for targets that don't lack default connectivity or support multiple connectivity options?

docs/porting/porting_full_process/integration tests.md
```diff
@@ -0,0 +1,65 @@
...
+   
+   Depending on connectivity type, you might have to specify more configuration options. For an example, please see the defined `CELLULAR` targets in `mbed_app.json`.
+
+- **Storage** - Specify the storage block device type, which dynamically adds the block device driver you specified at compile time. For example:
+
+   ```
+            "target.components_add" : ["SD"],
+   ```
+
+   Valid options are `SD`, `SPIF`, `QSPIF` and `FLASHIAP` (not recommended). For more available options, please see the [block device components](https://github.com/ARMmbed/mbed-os/tree/master/components/storage/blockdevice).
```
AnotherButler: Why isn't FLASHIAP recommended? Is it the only one of the group not recommended?

AnotherButler (changes requested):
Thanks for the PR. I left some questions.

docs/porting/porting_full_process/integration tests.md
MarceloSalazar: Yes, this is mainly for people porting, but can also help customers and developers with custom targets and to verify that the platform works well in their environment (in case there are some kind of network limitations).

docs/porting/porting_full_process/integration tests.md
MarceloSalazar: Both statements are correct. The usage of a shield and `target_extended.json` is only required if the platform doesn't connectivity and storage. The user can pass `target_extended.json` or any similar file with the similar of configuration.

docs/porting/porting_full_process/integration tests.md
MarceloSalazar: There are multiple reasons why we haven't included a driver in `mbed-os`, maybe we simply haven't had the chance to do it.
To add a library, a user can just follow the mbed-cli docs: https://os.mbed.com/docs/mbed-os/latest/tools/working-with-mbed-cli.html

docs/porting/porting_full_process/integration tests.md
MarceloSalazar: It's required for targets/platforms that don't have any type of IP connectivity.
It's also required in case the user wants to override the default connectivity option (replace by another available onboard or external shield). Hope this helps?

docs/porting/porting_full_process/integration tests.md
MarceloSalazar: It's not recommended because of the risks of wearing out the flash memory of the device.
We should probably remove this statement as afaik we're not commenting on the possibility of its usage.
I understand it's the only not recommended.

MarceloSalazar (commented):
@AnotherButler find comments - let me know if you need further information.

AnotherButler (approved):
@MarceloSalazar Thanks for the quick responses 👍 Please review my latest changes to make sure they are accurate.

MarceloSalazar (approved):
This looks really good - thanks!

PR #1151, opened 2019-10-03
Update release_process.md

AnotherButler:
Update document to reflect current release process.

### DIFF
--- a/docs/introduction/release_process.md
+++ b/docs/introduction/release_process.md
@@ -13,7 +13,7 @@ They can include:
 
 ## Feature release
 
-Feature releases occur once a quarter. As their name suggests, feature releases introduce new features to the code base. Unlike major releases, feature releases are backward compatible (source compatible). The second digit in the release number indicates the feature release. For example, Mbed OS 5.1.0 indicates major version 5, feature version 1.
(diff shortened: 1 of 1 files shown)


docs/introduction/release_process.md
```diff
@@ -25,7 +25,7 @@ They can include:
 
 ## Patch release
 
-Patch releases occur every two weeks. They are also backward compatible (source compatible). The last digit in the release number indicates the patch release. In Mbed OS 5.2.3, `3` shows the patch release is the third release in Mbed 5.2.
+Patch releases occur on the third Wednesday of every month. They are also backward compatible (source compatible). The last digit in the release number indicates the patch release. In Mbed OS 5.2.3, `3` shows the patch release is the third release in Mbed 5.2.
```
adbridge, line 28: Not strictly true. From January onwards we will make 'a' release on the 3rd Wednesday of every month. This could be a patch, feature or major release depending on what has landed on Master since the last release.

adbridge (changes requested):

docs/introduction/release_process.md
AnotherButler: I thought we should wait until January to change that in the docs and that for now we should indicate what the current release cadence is. What do you think?

adbridge: Well you could update it now and then have to update it again in 2 months time :)

adbridge: I'm fine with that too

adbridge (approved):

PR #1153, opened 2019-10-04
Update USBHID snippet as  API as changed

hugueskamba:
The previous C++ code snippet was outdated and could not be compiled (uses outdated and deprecated APIs).
Also simply the Python code.


AnotherButler:
Many thanks for the PR 👍 If this is meant to compile (and it sounds like it does), could you please remove the "TODO" at the top of the code? This will run the script that makes sure it compiles.

bulislaw (approved):

AnotherButler (approved):
Thanks 👍

PR #1156, opened 2019-10-16
Add the porting guide for the explicit pin-map extension.

mprse:
Add the porting guide for the explicit pin-map extension.

### DIFF
--- a/docs/porting/target/static_pinmap.md
+++ b/docs/porting/target/static_pinmap.md
@@ -1,39 +1,41 @@
-<h2 id="explicit-pinmap-port">Explicit Pinmap extension</h2>
+<h2 id="static-pinmap-port">Static Pinmap extension</h2>
 
-The **Explicit Pinmap extension** allows the peripheral configuration (pin/periheral/function) to be explicitly specified in the HAL API function.
+The **Static Pinmap extension** allows the peripheral configuration (pin/periheral/function) to be staticly specified in the HAL API function.
 
 ### Overview and background
 
-HAL APIs making use of pins take these pins in their constructor and use those pins to lookup which peripheral/function to use. The process of looking up the peripheral/function requires there to be a pinmap table that maps pins to peripherals/functions. This pinmap table takes up ROM which could be saved if the pinmap wasn't used. Explicit pinmap extension provides additional HAL API/constructors which takes pinmap as a parameter where pin/peripheral/function is specified explicitly and there is no need to use the pinmap tables. 
+In modern MCUs peripherals often can be mapped to different pins and each pin can have multiple functions. Mbed supports dynamic pin mapping, meaning that pins can be reconfigured at run time to be used by different driver. That provides great flexibility, but it's not free. There's non trivial ROM cost to maintain the pinmap tables and infrastructure to parse it. In some use cases this flexibility is worth the cost. Quite often pin configuration is frozen at hw design stage and doesn't require runtime modification. Shifting this configuration to compile time will allow us free memory associated with the dynamic approach.
 
-Supported peripherals:  
- - `PWM`  
- - `AnalogIn`  
- - `AnalogOut`  
- - `SPI`  
- - `I2C`  
- - `UART`  
- - `QSPI`  
- - `CAN`  
+HAL APIs making use of pins take these pins in their constructor and use those pins to lookup which peripheral/function to use. The process of looking up the peripheral/function requires there to be a pinmap table that maps pins to peripherals/functions. This pinmap table takes up ROM which could be saved if the pinmap wasn't used. Static pinmap extension provides additional HAL API/constructors which takes pinmap as a parameter where pin/peripheral/function is specified staticly and there is no need to use the pinmap tables.
+
+Supported peripherals:
+ - `PWM`
+ - `AnalogIn`
+ - `AnalogOut`
+ - `SPI`
+ - `I2C`
+ - `UART`
+ - `QSPI`
+ - `CAN`
  
 ### Requirements and assumptions
 
-1. Provide types which will hold explicit pinmaps for peripherals(`PWM`, `AnalogIn`, `AnalogOut`, `SPI`, `I2C`, `UART`, `QSPI`, `CAN`).
-2. Provide `xxx_init_direct(xxx_t *obj, explicit_pinmap_t *)` functions to HAL API (these functions will not use pinmap tables).
-3. Provide additional constructors in drivers layer which will use the `xxx_init_direct(xxx_t *obj, explicit_pinmap_t*)` HAL functions.
-4. Provide default weak implementations of `xxx_init_direct(explicit_pinmap_t *)` functions. These functions will call standard `xxx_init(xxx_t *obj, PinName, ...)` function (backward compatibility for targets which do not support explicit pinmap mechanism).
+1. Provide types which will hold static pinmaps for peripherals(`PWM`, `AnalogIn`, `AnalogOut`, `SPI`, `I2C`, `UART`, `QSPI`, `CAN`).
+2. Provide `xxx_init_direct(xxx_t *obj, static_pinmap_t *)` functions to HAL API (these functions will not use pinmap tables).
+3. Provide additional constructors in drivers layer which will use the `xxx_init_direct(xxx_t *obj, static_pinmap_t*)` HAL functions.
+4. Provide default weak implementations of `xxx_init_direct(static_pinmap_t *)` functions. These functions will call standard `xxx_init(xxx_t *obj, PinName, ...)` function (backward compatibility for targets which do not support static pinmap mechanism).
 5. Provide `constexpr` utility functions to lookup for pin mapping in compile time (requires C++14).
 6. Provide `constexpr` pin-map tables in the header file.
 7. Provide macros for the pin-map tables.
-8. Provide `EXPLICIT_PINMAP_READY` macro in `PinNames.h`.
+8. Provide `STATIC_PINMAP_READY` macro in `PinNames.h`.
 
-### Implementing explicit pin-map extension
+### Implementing static pin-map extension
 
-Most of the above points are already implemented. If you want to make explicit pinmap available on your platform please perform the following steps:  
+Most of the above points are already implemented. If you want to make static pinmap available on your platform please perform the following steps:  
 
-- Provide implementation of `xxx_init_direct(xxx_t *obj, explicit_pinmap_t *)` function (which does not use pinmap tables).  
-    - `xxx_init()` will use pinmap tables to determine associated peripheral/function with the given pins, populate the pin-map structure and call void `xxx_init_direct()`.  
-    - `xxx_init_direct()` will perform peripheral initialization using given explicit pinmap structure.  
+- Provide implementation of `xxx_init_direct(xxx_t *obj, static_pinmap_t *)` function (which does not use pinmap tables).
+    - `xxx_init()` will use pinmap tables to determine associated peripheral/function with the given pins, populate the pin-map structure and call void `xxx_init_direct()`.
+    - `xxx_init_direct()` will perform peripheral initialization using given static pinmap structure.
 
 Example implementation below:
 
@@ -54,18 +56,18 @@ void xxx_init(xxx_t *obj, PinName pin)
     int peripheral = (int)pinmap_peripheral(pin, PinMap_xxx);
     int function = (int)pinmap_find_function(pin, PinMap_xxx);
 
-    const PinMap explicit_pinmap = {pin, peripheral, function};
+    const PinMap static_pinmap = {pin, peripheral, function};
 
-    xxx_init_direct(obj, &explicit_pinmap);
+    xxx_init_direct(obj, &static_pinmap);
 }
 ```
 
 - Provide `constexpr` pin-map tables in the header file.
 
-Move pinmap tables from `PeripheralPins.c` to `PeripheralPinMaps.h` (create new file) and add `constexpr` specifier in the pin-map table declarations.  
-The tables are required in the header file, so can be included and used by constant expression utility functions to find and return mapping without pulling the pin-map table into the image.  
+Move pinmap tables from `PeripheralPins.c` to `PeripheralPinMaps.h` (create new file) and add `constexpr` specifier in the pin-map table declarations.
+The tables are required in the header file, so can be included and used by constant expression utility functions to find and return mapping without pulling the pin-map table into the image.
 
-**Note:**  
+**Note:**
 Please include `<mstd_cstddef>` module and use `MSTD_CONSTEXPR_OBJ_11` macro instead `constexpr` specifier. This must be done for backward compatibility with `ARM 5` compiler which does not support constant expressions. When `ARM 5` compiler is in use `MSTD_CONSTEXPR_OBJ_11` will be translated to `const`.  
 
 Example pin-map table below:
@@ -114,18 +116,18 @@ Since pin-map table names are not common across all targets the following macros
 #define PINMAP_CAN_TD [PinMap CAN RD]
 ```
 
-- Provide `EXPLICIT_PINMAP_READY` macro in `PinNames.h`  
+- Provide `STATIC_PINMAP_READY` macro in `PinNames.h`  
 
-Adding this macro will enable the explicit pin-map support for the target.
+Adding this macro will enable the static pin-map support for the target.
 
 ```
 /* If this macro is defined, then constexpr utility functions for pin-map seach can be used. */
-#define EXPLICIT_PINMAP_READY 1
+#define STATIC_PINMAP_READY 1
 ```
 
 ### Example usage/testing
 
-Use code below to check if explicit pinmap extension works. 
+Use code below to check if static pinmap extension works.
 
 ```
 int main()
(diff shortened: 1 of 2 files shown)


mprse:
cc @jamesbeyond

AnotherButler:
ping @jamesbeyond

AnotherButler:
ping @jamesbeyond

jamesbeyond:
Sorry for the later review.
My comments would be same as ARMmbed/mbed-os#11569
1. we want to decide the name of the feature, 'static pinmap' or `explicit pinmap`
2. rephrase words about background, make clear when user need to use `static pinmap`, when need `dynamic pinmap`

>In modern MCUs peripherals often can be mapped to different pins and each pin can have multiple functions. Mbed supports dynamic pin mapping, meaning that pins can be reconfigured at run time to be used by different driver. That provides great flexibility, but it's not free. There's non trivial ROM cost to maintain the pinmap tables and infrastructure to parse it. In some use cases this flexibility is worth the cost. Quite often pin configuration is frozen at hw design stage and doesn't require runtime modification. Shifting this configuration to compile time will allow us free memory associated with the dynamic approach.

mprse:
Fixed as suggested.

jamesbeyond (approved):
I am happy about the contents of the docs. @bulislaw please review

docs/porting/target/static_pinmap.md
```diff
@@ -0,0 +1,183 @@
...
+
+HAL APIs making use of pins take these pins in their constructor and use those pins to lookup which peripheral/function to use. The process of looking up the peripheral/function requires there to be a pinmap table that maps pins to peripherals/functions. This pinmap table takes up ROM which could be saved if the pinmap wasn't used. Static pinmap extension provides additional HAL API/constructors which takes pinmap as a parameter where pin/peripheral/function is specified staticly and there is no need to use the pinmap tables.
+
+Supported peripherals:
+ - `PWM`
```
bulislaw: What about digitalio? I mean it would make a lot of sense to say `Pins p23, p24 DigitalOut; p12 DigitalIn; Pins p34, 35 UART; all the other pins not used` In this case we would only need descriptors for 5pins and ignore the remaining ones saving further memory.

docs/porting/target/static_pinmap.md
```diff
@@ -0,0 +1,183 @@
...
+/* If this macro is defined, then constexpr utility functions for pin-map seach can be used. */
+#define STATIC_PINMAP_READY 1
+```
+
+### Example usage/testing
+
+Use code below to check if static pinmap extension works.
+
+```
+int main()
```
bulislaw, line 121: Do we need (or maybe already have) a test for that in HAL suite?

bulislaw (approved):
Looks good, just couple of queries.

docs/porting/target/static_pinmap.md
kjbracey: I'm not sure which, if any, platforms would benefit.

Most perform a fairly simple computation from pin number to GPIO bit/peripheral (because selecting GPIO is generally uniform across all pins), and the static information they write into their `gpio_t` is a bit more varied than doing the basic muxing of "set function 3" when selecting an alternate function - they're not using our `PinMap` type. The "pin/peripheral/function" 3-ple probably wouldn't cut it.

What might make sense is the option of a target `constexpr` `gpio_t`-returning function, and then have something like

    struct gpio_pinmap_t {
    #ifdef HAVE_STATIC_GPIO
        gpio_t gpio;
    #else
        PinName pin;
    #endif
    };
    
    constexpr gpio_pinmap_t gpio_get_pinmap(PinName pin)
    {
    #ifdef HAVE_STATIC_GPIO
        return { hal_get_gpio(PinName pin) };
    #else
        return { pin };
    #endif
    }

    DigitalIn::DigitalIn(const gpio_pinmap_t &map)
    #ifdef HAVE_STATIC_GPIO
        : gpio(map.gpio)
    #endif
    {
    #ifndef HAVE_STATIC_GPIO
        gpio_init(&gpio, map.pin);
    #endif
        ...
    }


    constexpr gpio_pinmap_t pin_a_map = gpio_get_pinmap(PIN_A);
    DigitalIn pin_a(pin_a_map);

No idea whether that really saves anything significant in practice. It might be worse on some platforms.

mprse: As far as I remember only some Freescale targets have pin-maps for GPIOs, but even these targets do not use them during initialization. Probably they are redundant.

docs/porting/target/static_pinmap.md
mprse: I added the test cases which use static pinmap to the FPGA tests. Unfortunately, we do not have tested `CAN`, `QSPI` and `AnalogOut` (this should be added later, because for now `Analogout` test has been removed due to some issue with the FPGA test shield).

AnotherButler:
What's the relevant code PR for this? Has it already merged?

docs/porting/target/static_pinmap.md
```diff
@@ -0,0 +1,180 @@
...
+Total flash memory (text + data): 44554(-1010) bytes
+```
+
+Run FPGA tests to check whether your implementation is valid:
+
+```
+ mbed test -t ARM -m K64F -n tests-mbed_hal_fpga_ci_test_shield*
+```
+
+<span class="notes">**Note:** Your target must be ready to run FPGA-Test-Shield tests. You can test the following peripherals: `Analogin`, `SPI`, `I2C`, `PWM`, `UART`.</span>
```
AnotherButler: Is this list likely to change? If so and it's easy for it to become out of date, I suggest removing it.

docs/porting/target/static_pinmap.md
```diff
@@ -0,0 +1,180 @@
...
+| hal\mbed_pinmap_common.o        |      0(-272) |   0(+0) |      0(+0) | // remove pinmap lib (this is common for all peripherals)
+| hal\mbed_ticker_api.o           |      978(+0) |   0(+0) |      0(+0) |
+| hal\mbed_us_ticker_api.o        |      114(+0) |   4(+0) |     65(+0) |
+| main.o                          |      70(+32) |   0(+0) |      0(+0) | // extra space for static pin map structure in application
+| platform\source                 |    5683(+46) |  64(+0) |    249(+0) | // extra space for UART static pin map structure to initialize the console
+| rtos\source                     |     8990(+0) | 168(+0) |   6626(+0) |
+| targets\TARGET_Freescale        |  16581(-816) |  12(+0) |    340(+0) | // remove pin maps and driver code reduction
+| Subtotals                       | 44290(-1010) | 264(+0) | 205518(+0) |
+Total static RAM memory (data + bss): 205782(+0) bytes
+Total flash memory (text + data): 44554(-1010) bytes
```
AnotherButler, line 166: Are these numbers constant, or do they differ by board, configuration and so on? If the latter, I recommend removing it.

docs/porting/target/static_pinmap.md
```diff
@@ -0,0 +1,180 @@
...
+3. Provide additional constructors in the drivers layer, which will use the `xxx_init_direct(xxx_t *obj, static_pinmap_t*)` HAL functions.
+4. Provide default weak implementations of `xxx_init_direct(static_pinmap_t *)` functions. These functions call standard `xxx_init(xxx_t *obj, PinName, ...)` function (backward compatibility for targets that don't support the static pin map mechanism).
+5. Provide `constexpr` utility functions to lookup for pin mapping in compile time (requires C++14).
+6. Provide `constexpr` pin map tables in the header file.
+7. Provide macros for the pin map tables.
+8. Provide `STATIC_PINMAP_READY` macro in `PinNames.h`.
+
+## Implementing static pin map extension
+
+Most of the above points are already implemented. To make static pin map available on your platform, please perform the following steps:
```
AnotherButler: If the above points are already implemented, do we still need to tell people to do them? Should we remove them?

docs/porting/target/static_pinmap.md
```diff
@@ -0,0 +1,180 @@
...
+5. Provide `constexpr` utility functions to lookup for pin mapping in compile time (requires C++14).
+6. Provide `constexpr` pin map tables in the header file.
+7. Provide macros for the pin map tables.
+8. Provide `STATIC_PINMAP_READY` macro in `PinNames.h`.
+
+## Implementing static pin map extension
+
+Most of the above points are already implemented. To make static pin map available on your platform, please perform the following steps: 
+
+- Provide implementation of `xxx_init_direct(xxx_t *obj, static_pinmap_t *)` function (which does not use pin map tables).
```
AnotherButler: Do these bulleted steps need to happen in order?

docs/porting/target/static_pinmap.md
```diff
@@ -0,0 +1,180 @@
...
+   }
+   ```
+
+- Provide `constexpr` pin-map tables in the header file.
+
+   Move pin map tables from `PeripheralPins.c` to `PeripheralPinMaps.h` (create new file) and add `constexpr` specifier in the pin map table declarations.
+   
+   The tables are required in the header file, so constant expression utility functions can include and use them to find and return mapping without pulling the pin map table into the image.
+   
+   <span class="notes">**Note:** Please include the `<mstd_cstddef>` module, and use the `MSTD_CONSTEXPR_OBJ_11` macro instead of the `constexpr` specifier. This must be done for backward compatibility with the Arm 5 compiler, which does not support constant expressions. When you use the Arm 5 compiler, `MSTD_CONSTEXPR_OBJ_11` translates to `const`.</span>
```
AnotherButler: Do you only need to do this when using the Arm 5 compiler?

docs/porting/target/static_pinmap.md
```diff
@@ -0,0 +1,180 @@
...
+ - `AnalogOut`.
+ - `SPI`.
+ - `I2C`.
+ - `UART`.
+ - `QSPI`.
+ - `CAN`.
+ 
+## Assumptions
+
+1. Provide types that will hold static pin maps for peripherals (`PWM`, `AnalogIn`, `AnalogOut`, `SPI`, `I2C`, `UART`, `QSPI`, `CAN`).
```
AnotherButler: Do these numbered steps need to happen in order?

docs/porting/target/static_pinmap.md
```diff
@@ -0,0 +1,180 @@
+<h1 id="static-pinmap-port">Static Pinmap extension</h1>
+
+The **Static Pinmap extension** allows you to statically specify the peripheral configuration (pin, peripheral or function) in the HAL API function.
```
AnotherButler, line 3: Is this a HAL API, or is an extension something different? If it is an API, we should link to the Doxygen in the implementing section.

docs/porting/target/static_pinmap.md
```diff
@@ -0,0 +1,180 @@
...
+
+In modern MCUs, you can often map peripherals to different pins, and each pin can have multiple functions. Mbed supports dynamic pin mapping, meaning you can reconfigure pins at run time for different drivers to use. That provides great flexibility, but it's not free. There's a nontrivial ROM cost to maintain the pin map tables and infrastructure to parse it. In some use cases, this flexibility is worth the cost. Often, pin configuration is frozen at the hardware design stage and doesn't require run time modification. Shifting this configuration to compile time allows free memory associated with the dynamic approach.
+
+HAL APIs using pins take these pins in their constructor and use those pins to look up which peripheral or function to use. The process of looking up the peripheral or function requires there to be a pin map table that maps pins to peripherals or functions. This pin map table takes up ROM. Static pinmap extension provides additional HAL API constructors, which take pin map as a parameter where the pin, peripheral or function is specified statically, and there is no need to use the pin map tables.
```
AnotherButler, line 7: So does this mean the benefit of the static pin map extension is that it saves ROM?

AnotherButler (commented):
I've left some initial queries.

docs/porting/target/static_pinmap.md
mprse: The HAL API will be extended: `xxx_init_direct()` functions will be added (e.g. `spi_init_direct()`). Also, the driver's layer will be extended and we will have additional constructors that will use `xxx_direct()` HAL API.

Which link should be added?
https://os.mbed.com/docs/mbed-os/v5.14/apis/spi.html 

I'm not sure how these docs are updated. At the moment there is no constructor that uses `xxx_direct()` HAL API on the list. Will it be added automatically when static pinmap is merged into the master?

docs/porting/target/static_pinmap.md
mprse: Correct. This is the main reason for adding this.

docs/porting/target/static_pinmap.md
AnotherButler: Which HAL API will be extended? SPI?

When new APIs are added with the right Doxygen tags and doxyfile options (I think here: https://github.com/ARMmbed/mbed-os/blob/master/doxygen_options.json), they show up in the Doxygen: https://os.mbed.com/docs/mbed-os/v5.14/mbed-os-api-doxy/annotated.html

mprse:
> What's the relevant code PR for this? Has it already merged?

https://github.com/ARMmbed/mbed-os/tree/feature-hal-spec-explicit-pinmap

This is on the feature branch. One PR is left and when it is merged we can bring this to master.

docs/porting/target/static_pinmap.md
mprse: The following HAL and Drivers (classes) APIs will be extended:
- SPI
- I2C
- Serial
- PWM
- AnalogIn
- AnalogOut
- CAN
- QSPI

docs/porting/target/static_pinmap.md
mprse: I will remove the list.

docs/porting/target/static_pinmap.md
mprse: For sure these are not constant, but the specific values are irrelevant here. The comments are important which shows in which modules we expect memory changes when static pinmap is used.

docs/porting/target/static_pinmap.md
mprse: I will try to modify this part and leave only the information required for porting.

docs/porting/target/static_pinmap.md
mprse: Nope, but all together :-)

docs/porting/target/static_pinmap.md
mprse: This is required generally, but it is used to deal with `ARM 5` compiler. If we use `constexpr`, then we will get an error in the case of `ARM 5` compiler. If we use `MSTD_CONSTEXPR_OBJ_11` macro we will get in case of  `ARM 5` `const` (instead `constexpr` which will be used in all other cases).

docs/porting/target/static_pinmap.md
mprse: No. This will be changed as you suggested.

docs/porting/target/static_pinmap.md
kjbracey: This wasn't actually anything to do with ARM C 5 - it was to handle the case where the header was going to be included from C code (inside the target code).

When we include it from the Mbed OS C++ code, we need to see it as `constexpr`, but if the target code includes it from C, it has to have it as `const`.

https://github.com/ARMmbed/mbed-os/pull/11399#discussion_r322608118

mprse: Yes, but I remember that this was also related to ARMC5:

https://github.com/ARMmbed/mbed-os/pull/11399#discussion_r322610020

But this was about the `constexpr` functions and here we have `constexpr` tables.

I mixed this. Thanks for catching this.

mprse:
@AnotherButler I addressed all the review comments except adding a link to the Doxygen.

AnotherButler:
I think my changes were overwritten. Could you please reinstate them?

mprse:
> I think my changes were overwritten. Could you please reinstate them?

Ohhh, sorry about this. I couldn't cherry-pick the lost commit so I tried to add your edits manually. Hopefully, it is ok now.

AnotherButler:
Thanks. That looks good 👍 

When do we expect the code to merge to master? 5.15 or later?

mprse:
PR https://github.com/ARMmbed/mbed-os/pull/11892.
It looks like its tagged for 5.15.

AnotherButler:
In that case, could an engineer please add it to the Doxyfile options in the `mbed-os` code? Once that merges, this will exist in Doxygen, and I can add a link to it.

mprse:
> When new APIs are added with the right Doxygen tags and doxyfile options (I think here:  https://github.com/ARMmbed/mbed-os/blob/master/doxygen_options.json), they show up in the Doxygen: https://os.mbed.com/docs/mbed-os/v5.14/mbed-os-api-doxy/annotated.html

PR  ARMmbed/mbed-os#11892 has been merged and it's tagged for `5.15`. I believe that doxyfile options are ok since `xxx_init_direct()` functions have been added to existing header files and the same Doxygen tags are used as for other functions. Example:
https://github.com/ARMmbed/mbed-os/blob/7177d8fefe2626d36bb576917cff70f466eb0fdf/hal/spi_api.h#L172-L189

So as far as I understand we expect that new direct init functions will show up here after release:
https://os.mbed.com/docs/mbed-os/v5.15/mbed-os-api-doxy/annotated.html

Then I will be able to add a link to them to the porting guide. @AnotherButler It that correct?

AnotherButler (approved):

PR #1160, opened 2019-11-03
Update example on integration test in host test

RonEld:
As described in [this post](https://forums.mbed.com/t/green-tea-example-on-docs-with-host-tests/5844).
Update the name of the executable test, as the path it is located is
different.


AnotherButler (approved):
Thanks for the PR 👍

AnotherButler:
ping @jamesbeyond

jamesbeyond (approved):
good catch. thanks for updating

PR #1165, opened 2019-11-07
Remove example_walkthrough.md from docs.json

AnotherButler:
Remove example walkthrough for quick start because it's no longer relevant to the quick start document.


hugueskamba (approved):
LGTM

PR #1166, opened 2019-11-08
Update Thread.md

AnotherButler:
Add helpful tip from discussion in another repository (internal ticket IOTDOCS-1220).

### DIFF
--- a/docs/api/rtos/Thread.md
+++ b/docs/api/rtos/Thread.md
@@ -10,6 +10,8 @@ The Thread class allows defining, creating and controlling parallel tasks.
 
 All the internal thread data structures are part of the C++ class, but by default, the thread stack is allocated on the heap. Memory is allocated at the run time during the call to `start` method. If you don't want to use dynamic memory, you can provide your own static memory using the constructor parameters.
 
+The main thread stack size is specified as `rtos.main-thread-stack-size` in the configuration .json file. That defines the main thread for `mbed_rtos_start` in `mbed_rtos_rtx.c`.
+
 The default stack size is 4K. However, the application can override it by using the configuration system and setting the `THREAD_STACK_SIZE` option to the required size in `mbed_app.json`. For details, please see the [configuration documentation](../../reference/configuration/configuration.md).
 
 ## Thread class reference
--- a/docs/api/rtos/Thread.md
+++ b/docs/api/rtos/Thread.md
@@ -91,9 +91,9 @@ The Callback API provides a convenient way to pass arguments to spawned threads.
 
 ## Debugging tips
 
-When debugging Thread, check the full RTX state (ready, blocked and so on) for a deadlock (two things waiting for each other) or a missed signal leading to an event machine stall. You can confirm a deadlock if you see two threads in the "blocked" state waiting for a resource that the other is supposed to signal.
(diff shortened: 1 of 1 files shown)


AnotherButler:
@kjbracey-arm Could you please review this to make sure it's right?

Is this also where information about thread state (Ready, Blocked and so on) and `info threads` would go?

docs/api/rtos/Thread.md
```diff
@@ -10,6 +10,8 @@ The Thread class allows defining, creating and controlling parallel tasks.
 
 All the internal thread data structures are part of the C++ class, but by default, the thread stack is allocated on the heap. Memory is allocated at the run time during the call to `start` method. If you don't want to use dynamic memory, you can provide your own static memory using the constructor parameters.
 
+The main thread stack size is specified as `rtos.main-thread-stack-size` in the configuration .json file. That defines the main thread for `mbed_rtos_start` in `mbed_rtos_rtx.c`.
```
kjbracey: Correct, but I think this needs to be the third paragraph. The paragraphs above and below are talking about extra threads you create using the `Thread` API, but this paragraph is a side point about the memory for the starting main thread (which is not created using `Thread`).

Possibly it should be a "Note", like in the section above.

kjbracey:
> Is this also where information about thread state (Ready, Blocked and so on) and `info threads` would go?

It could go here, yes, I can't immediately think of anywhere better. They're both about debugging threads, so if we have a general "debugging tips" page they could go there, but could be notes here.

(The `Thread` API does have the `get_state` call, but it's really just a debugging API. I've never seen it used except during printing out the thread list during a crash dump).

kjbracey (approved):

docs/api/rtos/Thread.md
```diff
@@ -89,6 +89,12 @@ The Callback API provides a convenient way to pass arguments to spawned threads.
 
 [![View code](https://www.mbed.com/embed/?url=https://os.mbed.com/teams/mbed_example/code/rtos_threading_with_callback/)](https://os.mbed.com/teams/mbed_example/code/rtos_threading_with_callback/file/5938bdb7b0bb/main.cpp)
 
+## Debugging tips
+
+When debugging Thread, check the full RTX state (ready, blocked and so on) for a deadlock (two things waiting for each other) or a missed signal leading to an event machine stall. You can confirm a deadlock if you see two threads in the "blocked" state waiting for a resource that the other is supposed to signal.
```
kjbracey: "debugging threads", I think is better. Otherwise it should be `Thread`.

"waiting for" -> "each waiting for"?

docs/api/rtos/Thread.md
```diff
@@ -89,6 +89,12 @@ The Callback API provides a convenient way to pass arguments to spawned threads.
...
+
+To reduce deadlocks, proactively code in a safe way by maintaining a mental ordering of your mutexes and never claiming a high-level mutex while holding a low-level one.
```
kjbracey: "mental ordering" sounds weird. Maybe just "ordering"?

I guess the point is that the system doesn't enforce this, so there's no way to follow this rule apart from self-discipline. Feels like you needs some sort of adjective to suggest that, but can't think of anything better.

(For comparison, in Symbian, to help with this, every mutex had a program-specified level number, and debug builds would stop with a diagnostic if you ever claimed a high-level one while holding a low-level one.  Although being a debug-build-only check was annoying - if someone broke the rule and never tried a debug build, then other people found out when they made a debug build to look at their stuff and found the system stopped with a mutex diagnostic in debug mode. Debug builds weren't tested in CI.)

PR #1167, opened 2019-11-08
Add suppressing output content to FileHandle.md

AnotherButler:
Add helpful tips from discussion in another repository (internal ticket IOTDOCS-1220).

### DIFF
--- a/docs/api/platform/FileHandle.md
+++ b/docs/api/platform/FileHandle.md
@@ -130,6 +130,12 @@ namespace
 
 Alternatively, an application could use the standard C `freopen` function to redirect `stdout` to a named file or device while running. However there is no `fdreopen` analog to redirect to an unnamed device by file descriptor or `FileHandle` pointer. 
 
+### Suppressing console output
+
+To suppress output, you can create a class derived from FileHandle to act as a mux and supply that to the console. That leverages the Mbed OS "redirect" `stdout` logic.
+
+More portably, you could ensure all your code uses your own `FILE *output_stream` rather than `stdout`, and you could dynamically change that between `stdout` and `fdopen(Sink)`. That then doesn't rely on "under C library" retargeting of `stdout`. Instead, you must switch streams at the application level.
+
 ### Polling and nonblocking
 
 By default, `FileHandle`s conventionally block until a `read` or `write` operation completes. This is the only behavior supported by normal `File`s, and is expected by the C library's `stdio` functions.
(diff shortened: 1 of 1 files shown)


AnotherButler:
ping @kjbracey-arm Could you please review this one, too?

docs/api/platform/FileHandle.md
```diff
@@ -130,6 +130,12 @@ namespace
 
 Alternatively, an application could use the standard C `freopen` function to redirect `stdout` to a named file or device while running. However there is no `fdreopen` analog to redirect to an unnamed device by file descriptor or `FileHandle` pointer. 
 
+### Suppressing console output
+
+To suppress output, you can create a class derived from FileHandle to act as a mux and supply that to the console. That leverages the Mbed OS "redirect" `stdout` logic.
```
kjbracey: I don't recall where this suggestion came from, but it looks familiar - my feeling is that it's a bit specialised and contrived, as a potential motivation for the API.

The simplest answer to totally suppressing console on most platforms is now to just turn off `target.console-uart` in JSON options. If the console would normally go to a UART on a target, that deactivates it and leaves no output. (By redirecting to a `Sink` `FileHandle` internally for you).

We've also just described the `mbed_overrride_console` mechanism above, so we can also trim a little here - maybe this should be a subsection of "Redirecting the console"?

I think add an extra first paragraph:

> To always suppress output, you can provide an `mbed_override_console` that returns a sink class that discards output. (Or for UART-console targets, you can simply set `target.console-uart` to false in your mbed_app.json).

Then this becomes 

>To temporarily suppress output, your `mbed_override_console` can return a class that acts as a switchable mux between a sink and the real output.

docs/api/platform/FileHandle.md
```diff
@@ -130,6 +130,14 @@ namespace
 
 Alternatively, an application could use the standard C `freopen` function to redirect `stdout` to a named file or device while running. However there is no `fdreopen` analog to redirect to an unnamed device by file descriptor or `FileHandle` pointer. 
 
+#### Suppressing console output
+
+To always suppress output, you can provide an `mbed_override_console` that returns a sink class that discards output. (For UART console targets, you can simply set `target.console-uart` to `false` in `your mbed_app.json`.)
```
kjbracey:
```suggestion
To always suppress output, you can provide an `mbed_override_console` that returns a sink class that discards output. (For UART console targets, you can simply set `target.console-uart` to `false` in your `mbed_app.json`.)
```

kjbracey (approved):

AnotherButler:
To avoid the CI failures I introduced, I'm closing this PR and making these changes directly.

AnotherButler:
Commit: https://github.com/ARMmbed/mbed-os-5-docs/commit/6e0a63d29b795d00ab46d8a80a8966fab5e8205c

PR #1170, opened 2019-11-19
Deprecate Wait.md

AnotherButler:
Replace content with deprecation warning.

### DIFF
--- a/docs/api/platform/Wait.md
+++ b/docs/api/platform/Wait.md
@@ -1,20 +1,3 @@
 # Wait
 
-Wait functions provide simple wait capabilities. The OS scheduler puts the current thread in `waiting state`, allowing another thread to execute. Even better: If there are no other threads in `ready state`, it can put the whole microcontroller to `sleep`, saving energy.
-
-## Avoiding OS delay
-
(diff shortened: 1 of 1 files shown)


bulislaw (changes requested):
Not all calls were deprecated.

AnotherButler:
So should we keep the body content and add a note that says "**Warning:** The `wait` and `wait_ms` calls of this API have been deprecated. In their place, please use [explicit sleep functions](../apis/power-management-sleep.html) and `wait_us`."?

AnotherButler:
ping @bulislaw

bulislaw (commented):
I think we should basically say the same as the deprecation notice (maybe with links): ```MBED_DEPRECATED_SINCE("mbed-os-5.14",
                      "'wait' is deprecated in favor of explicit sleep functions. To sleep, 'wait' should be replaced by "
                      "'ThisThread::sleep_for' (C++) or 'thread_sleep_for' (C). If you wish to wait (without sleeping), call "
                      "'wait_us'. 'wait_us' is safe to call from ISR context.")```

I'll update the examples not to use deprecated calls.

AnotherButler:
@bulislaw Does this resolve your concerns?

bulislaw (approved):

PR #1171, opened 2019-11-19
Add deprecation tip to PowerManagement.md

AnotherButler:
Add tip about replacing deprecated API.

### DIFF
--- a/docs/api/platform/PowerManagement.md
+++ b/docs/api/platform/PowerManagement.md
@@ -1,5 +1,7 @@
 # Power management (sleep)
 
+<span class="tips">**Tip:** This API replaces the deprecated Wait API.</span>
+
 ## Sleep
 
 There is only one sleep function in Mbed OS:


bulislaw (changes requested):
This is not accurate. The wait API was repleced by 
> MBED_DEPRECATED_SINCE("mbed-os-5.14",
                      "'wait' is deprecated in favor of explicit sleep functions. To sleep, 'wait' should be replaced by "
                      "'ThisThread::sleep_for' (C++) or 'thread_sleep_for' (C). If you wish to wait (without sleeping), call "
                      "'wait_us'. 'wait_us' is safe to call from ISR context.")

But it's a bit dispersed so adding a notice to each place would be probably too noisy.

AnotherButler:
@bulislaw Yes, I pulled the wording for this tip from that quote in the code. Are you saying we should keep the tip here and also add it to the ThisThread API page?

bulislaw:
I think this notice shouldn't be there. We don't want people in RTOS environment to call the system sleep right now. In most cases it won't change anything and it doesn't achieve the same result as `sleep_for`.

kjbracey:
There is the existing line 13 note saying you shouldn't normally use (system) `sleep` directly, but use RTOS API functions that can lead to system sleep (which includes `ThisThread::sleep_for`). That still applies, and possibly should be made stronger.

I'd say the same tip could be added to just the `ThisThread` page, and not add anything here.

AnotherButler:
Closing in favor of #1180

PR #1177, opened 2019-11-29
Enabling bare metal

evedon:
Information on how to enable bare metal is missing from this page (although it can be found here: https://os.mbed.com/docs/mbed-os/v5.14/tutorials/migrating-to-mbed-os-5.html)

### DIFF
--- a/docs/reference/bare_metal/baremetal_example.md
+++ b/docs/reference/bare_metal/baremetal_example.md
@@ -1,16 +1,6 @@
 # Bare metal example
 
-## Enabling Mbed OS bare metal
-
-You can enable Mbed OS bare metal by creating an `mbed_app.json` with the following contents:
-
-```
-{
-    "requires": ["bare-metal"]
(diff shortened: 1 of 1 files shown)


evedon:
@AnotherButler Please review asap. This information is now difficult to find in the docs and should be corrected.

AnotherButler:
@evedon I've replaced the duplication with a link to the specific section. If you think this is still too difficult to find, please let me know.

AnotherButler (approved):

evedon:
> @evedon I've replaced the duplication with a link to the specific section. If you think this is still too difficult to find, please let me know.

I think it should be the other way round. This section in the docs is about bare metal so it should describe how to enable it. The other page is a porting doc.

AnotherButler:
Good point. I switch it, so the other page points to this one, instead.

0xc0170 (approved):

PR #1183, opened 2019-12-09
SharedPtr: NULL -> nullptr

kjbracey:
https://github.com/ARMmbed/mbed-os/pull/12048 will require that `SharedPtr` users use `nullptr` rather than `NULL` or `0`, which will no longer compile.

Update example in code.

Change can be made to docs whenever, as `nullptr` form works now.


0xc0170 (approved):

AnotherButler (approved):
Thanks for the PR 👍

PR #1184, opened 2019-12-09
Callback: nullptr adjustments

kjbracey:
Adjust wording to talk of Callback's being "empty", which is clearer
than "null", and avoids thought of `NULL`. Add explicit note about using
`nullptr` to reset.

Preparation for https://github.com/ARMmbed/mbed-os/pull/12036 - can be
merged now as adjustments are valid for existing code.

### DIFF
--- a/docs/api/platform/Callback.md
+++ b/docs/api/platform/Callback.md
@@ -24,7 +24,7 @@ void run_timer_event(Callback<void(float)> on_timer) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


40Grit:
@aglass0fmilk

AnotherButler (approved):
Thanks for the PR 👍

0xc0170 (approved):

PR #1188, opened 2019-12-11
Update Platform configuration page and reference platform.stdio-minimal-console-only

hugueskamba:


AnotherButler (approved):
Thanks for the PR 👍

PR #1189, opened 2019-12-31
Add SD card content to SDBlockDevice.md

AnotherButler:
Add content about wiring an SD card module to a developer board.

### DIFF
--- a/docs/api/storage/SDBlockDevice.md
+++ b/docs/api/storage/SDBlockDevice.md
@@ -10,6 +10,21 @@ To configure this class, please see our [BlockDevice configuration documentation
 
 <span class="images">![](../images/sd-wiring.png)<span>How to wire an external SD-card module to a developer board</span></span>
 
+1. In `mbed_app.json`, override the SD pin mapping based on the wiring. For example: 
+
+   ```
+   "sd.SPI_MOSI" : "PC_3",
+   "sd.SPI_MISO" : "PC_2",
+   "sd.SPI_CLK"  : "PC_7",
+   "sd.SPI_CS"   : "PB_9",
+   ```
+   
+1. Add the `SD` component to `mbed_app.json`: 
+
+   ```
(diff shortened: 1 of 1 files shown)


docs/api/storage/SDBlockDevice.md
```diff
@@ -6,6 +6,17 @@ You can use the Mbed OS SD card block device, so applications can read and write
 
 To configure this class, please see our [BlockDevice configuration documentation](../reference/storage.html#blockdevice-default-configuration).
 
+## How to wire an external SD-card module to a developer board
+
+<span class="images">![](../images/sd-wiring.png)<span>How to wire an external SD-card module to a developer board</span></span>
```
teetak01: Should this PR also include the .png or how does that work?

AnotherButler: I added the PR separately: https://github.com/ARMmbed/mbed-os-5-docs/tree/development/docs/images

teetak01 (commented):
Should this also mention that application needs to override the SD pin mapping based on the wiring?

For. ex. as is done in https://github.com/ARMmbed/mbed-cloud-client-example/blob/master/mbed_app.json#L123-L126

docs/api/storage/SDBlockDevice.md
teetak01: Ok, a bit unclear why you would do those in seperate PRs.

VeijoPesonen:
Not to forget that the "SD" component needs to be added - [Component SD](https://github.com/ARMmbed/mbed-cloud-client-example/blob/fd495ae5d9c94e3501cc17bc2a71d51ed1487179/mbed_app.json#L109)

AnotherButler:
@VeijoPesonen and @teetak01 Do these changes look correct?

VeijoPesonen (approved):

teetak01 (approved):

PR #1215, opened 2020-01-28
Add note to RawSerial.md

AnotherButler:
Add deprecation note.

### DIFF
--- a/docs/api/io/RawSerial.md
+++ b/docs/api/io/RawSerial.md
@@ -1,5 +1,7 @@
 # RawSerial
 
+<span class="notes">**Note:** Some functions, variables or types have been deprecated. Please see the class reference linked below for details.</span>
+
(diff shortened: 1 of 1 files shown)


docs/api/io/RawSerial.md
```diff
@@ -1,5 +1,7 @@
 # RawSerial
 
+<span class="notes">**Note:** Some functions, variables or types have been deprecated. Please see the class reference linked below for details.</span>
```
hugueskamba, line 3: All of it has been deprecated.

AnotherButler: Is there an API we should direct people to use in its place?

hugueskamba: Yes, [`UnbufferedSerial`](https://github.com/ARMmbed/mbed-os/blob/master/drivers/UnbufferedSerial.h). Currently working on the docs.

AnotherButler:
Duplicates work of #1241

PR #1216, opened 2020-01-31
initial commit of 'four-ways-to-blinky' tutorial.

cvasilak:
This PR integrates a 'Four ways to blinky tutorial' into our doc explaining concepts such as 'Ticker/Timers/EventQueues' and some pitfalls and areas to be careful by the developer while using them. Mainly written to introduce an unfamiliar user to those concepts (such as me when I started using the platform).

I chose the '/APIS / RTOS and event handling' section in our 'development' site cause I believe it's the most appropriate place. Feel free though to comment if you want to be located somewhere else.


cvasilak:
Travis failure on code snippet is because they include line numbers (used in the tutorial text). At the moment this is not supported by the web tool but a ticket was raised to the web team to bring support (together with better syntax highlighting).

*Internal Jira
[#IOTWEB-5863](https://jira.arm.com/projects/IOTWEB/issues/IOTWEB-5863)

AnotherButler:
@jamesbeyond Could you please work with @cvasilak to put these snippets into the snippet repo?

AnotherButler:
@cvasilak What's the difference between this tutorial and the flow control tutorial, which also uses busy wait, ticker and thread to blink an LED? Are they similar enough they should be combined?

cvasilak:
> @cvasilak What's the difference between this tutorial and the flow control tutorial, which also uses busy wait, ticker and thread to blink an LED? Are they similar enough they should be combined?

They are similar albeit in my version I go over with much more detail on the concepts plus introducing the EventQueue. 

IMHO,  I think we can remove the flow-control tutorial and leave this new version only and use it as a base for future additions to the concepts. (eg. power consumption graphs, etc).

WDYT?

cvasilak:
> @jamesbeyond Could you please work with @cvasilak to put these snippets into the snippet repo?

Unfortunately, the [View Code] markdown doesn't support embedding the code with line numbers from external repositories. (Line numbers are used in the description to explain the code)

I see two possible solutions to this:
1) Either leave the code embedded in the document as it is at the moment. The code is pretty straightforward and extracted from our examples in the Class documentation.
2) or wait till [IOTWEB-5863](https://jira.arm.com/browse/IOTWEB-5863) is resolved.

I propose to go for [1] at the moment, but we should look to resolve [IOTWEB-5863](https://jira.arm.com/browse/IOTWEB-5863) in our new documentation site since upcoming tutorials (at least from my side) will use the line numbers mechanism to describe the code.

cc/ @jamesbeyond

AnotherButler:
@cvasilak We don't want to create any new hardcoded snippets. We're trying to get away from that. Line numbers should not block this tutorial from moving forward. Please follow our process with @jamesbeyond to transclude a snippet in the docs_only repo.

Also, what's the internal JIRA # for this task?

jamesbeyond:
Hi @cvasilak, I had a look at your examples, here are my comments

1)  As Amanda explained, we found it is very difficult to manage mixed code snippets and documentation, because we can't test those code snippets easily, and verify whether they are still working with latest mbed-os. So all the code snippets need to be checked in https://github.com/ARMmbed/mbed-os-examples-docs_only. 

2) Reference code by line number is not very reliable, because code changes could happen without your awareness. The final doc website will NOT show the line number anyway, no matter it is a hard-codded snippet or a referenced code from [code view]. Only the Github show line numbers. Also requesting webteam to add a new feature to show line number, is it not going happen any time soon as I can imagine. So what I can recommend is use inline C comments inside your code or use `prinf line inside flip( ) fucntion` to reference it.

3) Serial Class in your example will be deprecated soon. The core team is working on that at the moment. So in your case, I assume `printf()` can be used the same way. Or the proposed new class is `UnbufferedSerial` which can be used inside interrupt mode. 
for more details please check:
https://confluence.arm.com/pages/viewpage.action?pageId=361056431
https://github.com/ARMmbed/mbed-os/blob/master/docs/design-documents/drivers/serial/serial.md
https://jira.arm.com/browse/IOTCLOUDPR-3294

cvasilak:
@jamesbeyond sounds good, will go ahead with the inline-comments and the external repo then. Let me rework the PR and update it.

cc/ @AnotherButler

AnotherButler:
Before we focus too much on the snippets, I want to flag some bigger-picture concerns. I'm not sure why we're replacing an existing tutorial that works with one that does the same thing. I don't want us to sink too much time into duplicated content.

Could we instead apply the new pieces to the flow control tutorial?

cvasilak:
@AnotherButler agree with your concern.
In essence, my tutorial was written from a beginners perspective when I was trying to learn the concepts of 'Ticker/Timers/EventQueues' using the blinky-example as a base.
In this version, I cover in a bit more detail all the concepts explained in the 'flow-control' plus the addition of EventQueues and some pitfalls that the programmer may face when using them.

IMHO, I believe we can remove the 'flow-control' tutorial and keep this one since it's a bit more detailed in the concepts.

cc) @jamesbeyond

AnotherButler:
No, we don't want to completely replace good, functioning content. Please instead apply the new pieces to the flow control tutorial.

cvasilak:
@AnotherButler I can understand, let me close this for now as to not to have overlap with the existing tutorial and re-focus mine.

PR #1223, opened 2020-02-06
Move changes in custom-target-porting.md

AnotherButler:
Move changes above snippet to make it clear what changed.


ladislas:
More verbose but I really like that! :)

big 👍 for not just listing changes but also explaining why we made them :)

ladislas:
More verbose but I really like that! :)

big 👍 for not just listing changes but also explaining why we made them :)

PR #1227, opened 2020-02-07
Remove leading slash from redirect value v5.13

alidav01:
The redirects functionality does not handle redirects that have a leading slash such as this one (this will just lead to a page error at the moment).


iriark01:
FYI @AnotherButler

PR #1231, opened 2020-02-09
Fix config option check example

RonEld:
Fix the configuration check of `MBED_CONF_RTOS_PRESENT` to
`MBED_CONF_RTOS_API_PRESENT` to align to the [code](https://github.com/ARMmbed/mbed-os/blob/master/mbed.h#L21).
Reported by Email from 王高腾


AnotherButler (approved):
Thanks for the PR 👍

Patater (approved):

PR #1239, opened 2020-02-21
Update DeviceKey document after Root of trust refactoring.

tymoteuszblochmobica:
DeviceKey class has been modified to not allow automatic Root of trust  generation. 
This requires   docs/api/security/Devicekey.md  document update.

### DIFF
--- a/docs/api/security/Devicekey.md
+++ b/docs/api/security/Devicekey.md
@@ -19,13 +19,17 @@ The characteristics required by this root of trust are:
 
 The DeviceKey feature keeps the root of trust key in internal storage, using the NVStore component. Internal storage provides protection from external physical attacks to the device.
 
-The root of trust is generated at the first use of DeviceKey if the true random number generator is available in the device. If no true random number generator is available, you must pass the injected root of trust key to the DeviceKey before you call the key derivation API.
+The root of trust is not generated at the first use of DeviceKey. It must be created explicitly before. If the true random number generator is available in the device it is used for root of trust generation. Otherwise injected root of trust must be passed to the DeviceKey before key derivation API call.
 
 ### Key derivation API
 
-`generate_derived_key`: This API generates a new key based on an array of data ([salt](https://en.wikipedia.org/wiki/Salt_(cryptography)) the caller provides. A single salt value always generates the same key, so if you need a new key, you must use a new salt value. The salt can have any value - array, string and so on.
+`generate_derived_key`: This API generates a new key based on an array of data ([salt](https://en.wikipedia.org/wiki/Salt_(cryptography)) the caller provides. A single salt value always generates the same key, so if you need a new key, you must use a new salt value. The salt can have any value - array, string and so on. Root of trust must be generated or injected earlier, otherwise error will be returned. 
 
 The generated keys can be 128 or 256 bits in length.
+
+#### Root of Trust generation
+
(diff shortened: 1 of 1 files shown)


docs/api/security/Devicekey.md
```diff
@@ -19,14 +19,18 @@ The characteristics required by this root of trust are:
 
 The DeviceKey feature keeps the root of trust key in internal storage, using the NVStore component. Internal storage provides protection from external physical attacks to the device.
 
-The root of trust is generated at the first use of DeviceKey if the true random number generator is available in the device. If no true random number generator is available, you must pass the injected root of trust key to the DeviceKey before you call the key derivation API.
+The root of trust is not generated at the first use of DeviceKey. It must be created explicitly before. If the true random number generator is available in the device it is used for root of trust generation. Otherwise injected root of trust must be passed to the DeviceKey before key derivation API call.
```
SeppoTakalo, line 22: Please remove the first sentence.
It is not helpful to tell what the API does not do, just remove the reference and start with `The root of trush must be created before its first use` or something like that.

Also `If the true random number generator is available in the device it is used ` is also slightly missleading. We TRNG may be used.. but you can also inject the security.
So rewrite the sentence to better indicate that we have two options.

docs/api/security/Devicekey.md
```diff
@@ -19,14 +19,18 @@ The characteristics required by this root of trust are:
...
-`generate_derived_key`: This API generates a new key based on an array of data ([salt](https://en.wikipedia.org/wiki/Salt_(cryptography)) the caller provides. A single salt value always generates the same key, so if you need a new key, you must use a new salt value. The salt can have any value - array, string and so on.
+`generate_derived_key`: This API generates a new key based on an array of data ([salt](https://en.wikipedia.org/wiki/Salt_(cryptography)) the caller provides. A single salt value always generates the same key, so if you need a new key, you must use a new salt value. The salt can have any value - array, string and so on. Root of trust must be generated or injected earlier, otherwise error will be returned. 
 
 The generated keys can be 128 or 256 bits in length.
 
+#### Root of Trust generation
+
+`generate_root_of_trust`: You must call this API once in the lifecycle of the device, before any call to key derivation, if the device supports true random number generator (`DEVICE_TRNG` is defined).
+
 #### Root of Trust Injection API
```
SeppoTakalo, line 34: Please combine these two sections into one.
```
#### Root of Trust generation

Root of Trust must be present in the device before its first use. You have two options to generate Root of Trust:

1. Use random number generator to create the key
2. Inject the key manually

<then explain both>
```

Also, both requires their own code examples. You have not demonstrated just the generation part.

docs/api/security/Devicekey.md
```diff
@@ -41,6 +45,12 @@ To instantiate DeviceKey, you need to call its `get_instance` member function as
     DeviceKey &deviceKey = DeviceKey::get_instance();
 ```
 
+To generate root of trust before key derivation:
+
+```c++ TODO
+    DeviceKey &deviceKey = DeviceKey::get_instance().generate_root_of_trust();
```
SeppoTakalo, line 51: This code example does not seem right. We don't get the reference as a return value, but an error code instead. Maybe you meant:
```C++

int err = DeviceKey::get_instance().generate_root_of_trust();
if (err == 0)
    // Success
```

SeppoTakalo (changes requested):

PR #1243, opened 2020-02-24
Update the version of example to 5.15

RonEld:
Update the link of the example code for TLS Socket to 5.15,
as current documentation refers to this version.

Note that this version code has a bug in it, but already fixed, and will probably be updated with 5.16 version, so need to  stay tuned on this.
References:
* https://github.com/ARMmbed/mbed-os-example-tls-socket/issues/48
* https://forums.mbed.com/t/warning-connect-is-deprecated-string-based-apis-are-deprecated-since-mbed-os-5-15/7347


AnotherButler (approved):
Thanks for the PR 👍

PR #1244, opened 2020-02-24
Update DeviceKey document after root of trust refactoring

tymoteuszblochmobica:
DeviceKey class has been modified to not allow automatic Root of trust generation.
This requires docs/api/security/Devicekey.md document update.

### DIFF
--- a/docs/api/security/Devicekey.md
+++ b/docs/api/security/Devicekey.md
@@ -19,27 +19,24 @@ The characteristics required by this root of trust are:
 
 The DeviceKey feature keeps the root of trust key in internal storage, using the KVStore component. Internal storage provides protection from external physical attacks to the device.
 
-The root of trust must be created before its first use. Otherwise key  derivation API  will fail.
+The root of trust must be created before its first use. Otherwise, the key derivation API fails.
 
 ## Key derivation API
 
 `generate_derived_key`: This API generates a new key based on an array of data ([salt](https://en.wikipedia.org/wiki/Salt_(cryptography)) the caller provides. A single salt value always generates the same key, so if you need a new key, you must use a new salt value. The salt can have any value - array, string and so on.
 
-The generated keys can be 128 or 256 bits in length.
+The generated keys can be 128b or 256b in length.
 
 ### Root of Trust generation API
 
-DeviceKey class needs ROT ready to use  before derivation API first call. There are two options to achieve it:
+DeviceKey class needs root of trust ready to use before the derivation API's first call. There are two options to achieve this:
 
+- Create a device key using a built-in random number generator.
+- Manually fill the device key data array.
 
-- Create device key using built-in random number generator 
+Both cases requires injecting this key data to the KVStore reserved area.
 
-- Manually fill device key data array 
-
-Both cases requires injecting this key data to kvstore  reserved area.  
-
-The first way is used when device supports random number generator -  ` DEVICE_TRNG` is defined.
-Then  `generate_root_of_trust` must be called only once.
+Use this first method when the device supports a random number generator - when ` DEVICE_TRNG` is defined. Then, call `generate_root_of_trust` only once.
 
 ```c++ NOCI
 int status = DeviceKey::get_instance().generate_root_of_trust();
@@ -50,7 +47,7 @@ if(status == DEVICEKEY_SUCCESS) {
 }
 ```
 
-If `DEVICE_TRNG` is not defined  then key buffer must be filled manually and followed by `device_inject_root_of_trust` call. The example below shows an injection of a dummy key.  
+If `DEVICE_TRNG` is not defined, the key buffer must be filled manually and followed by `device_inject_root_of_trust` call. The example below shows an injection of a dummy key:
 
 ```c++ NOCI
 uint32_t key[DEVICE_KEY_32BYTE / sizeof(uint32_t)];
@@ -67,9 +64,9 @@ if(status == DEVICEKEY_SUCCESS) {
 
 ### Using DeviceKey
 
-DeviceKey is a singleton class, meaning that the system can have only a single instance of it.
+DeviceKey is a singleton class, meaning the system can have only a single instance of it.
 
-To instantiate DeviceKey, you need to call its `get_instance` member function as following:
+To instantiate DeviceKey, call its `get_instance` member function:
 
 ```c++ TODO
     DeviceKey &deviceKey = DeviceKey::get_instance();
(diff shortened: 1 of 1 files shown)


docs/api/security/Devicekey.md
```diff
@@ -19,17 +19,48 @@ The characteristics required by this root of trust are:
...
 
 The generated keys can be 128 or 256 bits in length.
 
-### Root of Trust Injection API
+### Root of Trust generation API
 
-`device_inject_root_of_trust`: You must call this API once in the lifecycle of the device, before any call to key derivation, if the device does not support true random number generator (`DEVICE_TRNG` is not defined).
+DeviceKey class needs ROT ready to use  before derivation API first call. There are two options to achieve it:
+
+-Create device key using built-in random number generator
```
ohadhawk: @tymoteuszblochmobica tymoteuszblochmobicaI think you meant to have bulleted items, but instead the dash simply connects to the first word:
image

tymoteuszblochmobica: Done

docs/api/security/Devicekey.md
```diff
@@ -19,17 +19,49 @@ The characteristics required by this root of trust are:
 
 The DeviceKey feature keeps the root of trust key in internal storage, using the KVStore component. Internal storage provides protection from external physical attacks to the device.
 
-The root of trust is generated at the first use of DeviceKey if the true random number generator is available in the device. If no true random number generator is available, you must pass the injected root of trust key to the DeviceKey before you call the key derivation API.
+The root of trust must be created before its first use. Otherwise key  derivation API  will fail.
```
AnotherButler: Does the user need to create it?

SeppoTakalo: In development phase, probably yes. In production device it should be injected in the factory.

docs/api/security/Devicekey.md
```diff
@@ -19,17 +19,49 @@ The characteristics required by this root of trust are:
...
+The first way is used when device supports random number generator -  ` DEVICE_TRNG` is defined.
+Then  `generate_root_of_trust` must be called only once.
+
+
+	int status = DeviceKey::get_instance().generate_root_of_trust();
+	    if(status == DEVICEKEY_SUCCESS) {
+            //success
+        } else {
+            //error
+        }
```
SeppoTakalo: This code block does not seem to be aligned correctly.
Can you use the backticks, so it can be properly aligned?

SeppoTakalo (approved):
Documentation looks OK.

Just a personal preference, I would prefer to use three backticks to format my codeblocks.

docs/api/security/Devicekey.md
tymoteuszblochmobica: Changed

SeppoTakalo (approved):
LGTM

SeppoTakalo:
Oh.. you probably need to start the code snippets with
<pre>
```C++ NOCI
</pre>

Because otherwise the CI will try to build those, and fail as they are not complete.

**Review by AnotherButler — APPROVED** [review]
Thanks for the PR 👍 Please review my edits to make sure I didn't accidentally change the meaning of anything.

File: docs/api/security/Devicekey.md [hunk 4]
```diff
@@ -19,27 +19,24 @@ The characteristics required by this root of trust are:
...
-- Create device key using built-in random number generator 
+Both cases requires injecting this key data to the KVStore reserved area.
 
-- Manually fill device key data array 
-
-Both cases requires injecting this key data to kvstore  reserved area.  
-
-The first way is used when device supports random number generator -  ` DEVICE_TRNG` is defined.
-Then  `generate_root_of_trust` must be called only once.
+Use this first method when the device supports a random number generator - when ` DEVICE_TRNG` is defined. Then, call `generate_root_of_trust` only once.
```
SeppoTakalo: This is now misleading (as was the original as well). The Call of `generate_root_of_trust()` is the method of generating the key.
So maybe the sentence should be like:
> When `DEVICE_TRNG` is defined, the device supports a random number generator and you may generate the key by calling `generate_root_of_trust()`.
> The call succeed only if the key does not already exist. Subsequent calls will fail, without modifying the existing key.

docs/api/security/Devicekey.md
```diff
@@ -50,7 +47,7 @@ if(status == DEVICEKEY_SUCCESS) {
 }
 ```
 
-If `DEVICE_TRNG` is not defined  then key buffer must be filled manually and followed by `device_inject_root_of_trust` call. The example below shows an injection of a dummy key.  
+If `DEVICE_TRNG` is not defined, the key buffer must be filled manually and followed by `device_inject_root_of_trust` call. The example below shows an injection of a dummy key:
```
SeppoTakalo: Similarly as above
> If `DEVICE_TRNG` is not defined, the key buffer must be filled manually by calling `device_inject_root_of_trust()`. The example below shows an injection of a dummy key:

SeppoTakalo (changes requested):
I propose change to two sentences which are a bit misleading.
Otherwise looks OK.

Just a personal preference, do we need to shorten bits to just `b`? For example https://en.wikipedia.org/wiki/Advanced_Encryption_Standard uses `AES is a variant of Rijndael which has a fixed block size of 128 bits, and a key size of 128, 192, or 256 bits.` then later on, if they talk about key length they leave the `bit` out completely.

AnotherButler:
@SeppoTakalo Thanks for the review and suggested changes 👍 Does this look better?

docs/api/security/Devicekey.md
```diff
@@ -19,31 +19,62 @@ The characteristics required by this root of trust are:
...
 
-`device_inject_root_of_trust`: You must call this API once in the lifecycle of the device, before any call to key derivation, if the device does not support true random number generator (`DEVICE_TRNG` is not defined).
+DeviceKey class needs root of trust ready to use before the derivation API's first call. There are two options to achieve this:
+
+- Create a device key using a built-in random number generator.
+- Manually fill the device key data array.
+
+Both cases requires injecting this key data to the KVStore reserved area.
+
+When `DEVICE_TRNG` is defined, the device supports a random number generator, and you may generate the key by calling `generate_root_of_trust()`. The call succeeds only if the key does not already exist. Subsequent calls will fail if you don't modify the existing key.
```
SeppoTakalo: > Subsequent calls will fail if you don't modify the existing key.

Sorry, I was probably unclear here. There is no way to modify the key, this sentence needs rewrite.

I tried to impress that you can only generate the key once. Modifying it, or regenerating it is not possible.

So subsequent call to this API will fail. And those calls don't modify the key.

Would it be clearer with:
```
The call succeeds only if the key does not already exist. Existing key cannot be changed.
```

SeppoTakalo (changes requested):
Just a minor change.
Otherwise looks OK.

SeppoTakalo (approved):

SeppoTakalo:
Looks good now. Thanks @AnotherButler

JammuKekkonen (approved):

PR #1248, opened 2020-02-25
update time example links

maciejbocianski:
examples move PR: https://github.com/ARMmbed/mbed-os-examples-docs_only/pull/76


AnotherButler (approved):
Thanks for the PR 👍

PR #1282, opened 2020-03-18
Update drivers.md

AnotherButler:
Replace with new section name to reflect new structure.

### DIFF
--- a/docs/api/io/drivers.md
+++ b/docs/api/io/drivers.md
@@ -1,23 +1,23 @@
-# Drivers overview
+# Input/Output (I/O) overview
 
-Drivers provide access to general purpose microcontroller hardware. Drivers typically access a microcontroller's hardware peripherals directly, and may form the lowest lever of a protocol or network stack.
+I/O APIs provide access to general purpose microcontroller hardware. I/O APIs typically access a microcontroller's hardware peripherals directly, and may form the lowest lever of a protocol or network stack.
 
-The Mbed OS driver APIs include analog and digital inputs and outputs on development boards, as well as digital interfaces, which allow your board to interface with a computer or external devices. With these drivers, you can read or set the voltage of analog pins, control digital pins individually or grouped into a bus and trigger an event when more data is available on the bus. You can also control the frequency and mark to space ratio, or duty cycle, of a Pulse Width Modulated (PWM) digital output.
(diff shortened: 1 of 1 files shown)


AnotherButler:
@hugueskamba @rajkan01 When you get the chance, could one of you please review this to make sure the content is still accurate with these changes?

AnotherButler:
@iriark01 This is waiting on engineering review. I changed things to match our new section name, but I need confirmation I didn't accidentally change the meaning of anything.

iriark01:
@rajkan01 and @hugueskamba will one of you have time to look at this?

hugueskamba (commented):
Why renaming drivers to I/0?
Are we referring to the `mbed-os/drivers` directory? Not all drivers are for I/O (e.g watchdog) and the directory is called `drivers` not `io`.
@bulislaw @evedon Can you please comment?

iriark01:
From the restructure work we did, we learned:
* It's more intuitive to users.
* It's task-based, rather than code-based.
* Yes, we still have a devices directory in the code.
* If there are drivers that aren't I/O, you may need to move them elsewhere. No on had brought that up previously, but I could see that slipping through.

evedon:
This page should be for drivers in general. If we need a specific page for I/O drivers then we should write a new one.

iriark01:
In this structure, do you think it's worth splitting I/O and Drivers to two?
https://os.mbed.com/docs/mbed-os/development/introduction/index.html

iriark01:
I've split the group into two: 
https://os.mbed.com/docs/mbed-os/development/apis/drivers.html
https://os.mbed.com/docs/mbed-os/development/apis/input-output-i-o.html
For the intro page, I think I need a bit of help. I currently have the same page twice, and for I/O it simply says "this is a subset of Drivers". Before I try rewriting the intro, I would like to know if there are parts of it that are only correct for I/O or that are never correct for it.

iriark01:
@hugueskamba could you help me figure out what the intros should cover?
https://os.mbed.com/docs/mbed-os/development/apis/drivers.html
https://os.mbed.com/docs/mbed-os/development/apis/input-output-i-o.html

hugueskamba:
> @hugueskamba could you help me figure out what the intros should cover?
> https://os.mbed.com/docs/mbed-os/development/apis/drivers.html
> https://os.mbed.com/docs/mbed-os/development/apis/input-output-i-o.html

Perhaps something similar to http://www.keil.com/pack/doc/CMSIS/Driver/html/index.html for the Drivers page.
Perhaps something similar to https://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_IO/FreeRTOS_Plus_IO.html for the I/O page

iriark01:
Rejecting the PR; I will work on intros for API categories when we have them all sorted out

PR #1289, opened 2020-04-08
Target, board and build

iriark01:
Clarifying that:

- Target isn't another word for board; it's a build tools representation of it.
- What a build is.
- What board is, physically.

### DIFF
--- a/docs/introduction/terms.md
+++ b/docs/introduction/terms.md
@@ -4,7 +4,7 @@
 
 **API** - [Application programming interface](../apis/index.html).
 
-**Application** - An executable module built out of libraries that acts as a final product.
+**Application** - An executable software module built out of libraries that acts as a final product.
 
 **Arm Mbed CLI** - The name of the Arm Mbed [command-line tool](../tools/developing-mbed-cli.html), packaged as `mbed-cli`.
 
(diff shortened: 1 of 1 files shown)


docs/introduction/terms.md
```diff
@@ -26,16 +28,20 @@
 
 **Block Device** - A [file](../porting/porting-storage.html) that represents a device, with data that can be read or written to it in blocks.
 
-**Board** - An Mbed Enabled development board. You can see a [complete list](https://os.mbed.com/platforms/) of Mbed Enabled boards.
+**Board** - Hardware that can run an Mbed OS application. Usually refers to an [Mbed Enabled development board](https://os.mbed.com/platforms/). Its software representation in the build tools is as a target.
```
rosiemiller2: I'm not 100% on a target being a software representation. My understanding was that a target represented a 'destination' for code built by the tools (either physical or abstract). This seems to include boards but I think for convenience rather than having to specify exactly which bits on the board you are flashing to, though you could potentially have a target that was only a part of a board, or a particular mode eg. secure/non-secure?

I am trying to work out if this is consistent with CMSIS packs, since they also use 'target', 'board' and 'device' and you seem to be able to have both a board and a target.

However, I am not sure that including all this complexity here is necessary so maybe this is good as a roughly-correct overview.

iriark01:
On hold until meeting.

rosiemiller2 (approved):
👍

iriark01:
Right. Merging this, and will also add it to 5.15.
There's a lot of work in the docs themselves to get consistent wording, but I didn't want to overload this PR.

PR #1293, opened 2020-04-13
Added Mbed 2 to Mbed OS bare metal porting guide

evedon:


iriark01:
Waiting for engineering review

MarceloSalazar:
Looks good - thanks!

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
+# Porting an Mbed 2 target to Mbed OS 6 bare metal
+
+Mbed OS bare metal provides fewer features compared to a full RTOS platform which makes it suitable for constrained microcontrollers. This document describes how to configure an Mbed 2 target to run in bare metal mode and how to validate the port.
```
hugueskamba: So an RTOS is not suitable for constrained MCUs? A term used on the [bare metal page](https://os.mbed.com/docs/mbed-os/v5.15/reference/mbed-os-bare-metal.html) is "ultraconstrained" maybe consider using it instead.
Also were you trying to compare the bare metal mode with RTOSes in general or with Mbed OS's RTOS?

hugueskamba: When we say Mbed **OS** bare metal do we imply there is an OS? Should it be "Mbed bare metal" or "Mbed OS bare metal mode" instead?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Configuring your target
+
+You will use the blinky bare metal Mbed OS example and make it run on your target. The configuration will be validated upon successful compilation of the example.
```
hugueskamba: Not only should it compile, it should also run to be validated.

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+
+The main steps are:
+1. Clone [mbed-os-example-blinky-baremetal](https://github.com/ARMmbed/mbed-os-example-blinky-baremetal).
+1. Edit `targets.json`. In this step, you will configure your target to run in Mbed OS 6 and then add configuration parameters for the bare metal mode.
```
hugueskamba:
```suggestion
1. Edit `targets.json`. In this step, you will configure your target to support Mbed OS 6 and add configuration parameters for the bare metal mode. 
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+Mbed OS bare metal provides fewer features compared to a full RTOS platform which makes it suitable for constrained microcontrollers. This document describes how to configure an Mbed 2 target to run in bare metal mode and how to validate the port.
+
+## Configuring your target
+
+You will use the blinky bare metal Mbed OS example and make it run on your target. The configuration will be validated upon successful compilation of the example.
+
+The main steps are:
+1. Clone [mbed-os-example-blinky-baremetal](https://github.com/ARMmbed/mbed-os-example-blinky-baremetal).
+1. Edit `targets.json`. In this step, you will configure your target to run in Mbed OS 6 and then add configuration parameters for the bare metal mode. 
+1. Compile mbed-os-example-blinky-baremetal. In this step, you will validate the configuration and troubleshoot any issue.
```
hugueskamba:
```suggestion
1. Compile and run mbed-os-example-blinky-baremetal. In this step, you will validate the configuration and troubleshoot any issue.
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Clone mbed-os-example-blinky-baremetal
+
+Follow these steps:
```
hugueskamba:
```suggestion
Run the following command:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+1. git clone https://github.com/ARMmbed/mbed-os-example-blinky-baremetal
```
hugueskamba:
```suggestion
1. `git clone https://github.com/ARMmbed/mbed-os-example-blinky-baremetal`
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+1. cd mbed-os-example-blinky-baremetal
```
hugueskamba:
```suggestion
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Edit targets.json
+
+Find your target in `targets.json` and update its configuration as described below.
+
+Configure your target to run in Mbed OS 6:
```
hugueskamba:
```suggestion
Configure your target to support Mbed OS 6:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+- Remove the `release_versions` property as it is no longer necessary to specify the list of major versions of Mbed OS that the target supports.
```
hugueskamba:
```suggestion
- Remove the `release_versions` attribute so the target can support any version of Mbed OS post Mbed 2.
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+```
+"release_versions": ["2"]
+```
+
+- Add a new property (this property is not supported by the tools yet therefore it is ignored) `supported_application_modes` to indicate that the application mode supported by this target is bare metal.
```
hugueskamba:
```suggestion
- Add the `"supported_application_modes"` attribute to indicate that the application mode supported by this target is bare metal. This attribute is currently ignored by the build tools as it is not yet supported but will be essential once they are updated.
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+```
+"release_versions": ["2"]
+```
+
+- Add a new property (this property is not supported by the tools yet therefore it is ignored) `supported_application_modes` to indicate that the application mode supported by this target is bare metal. 
+```
+"supported_application_modes" : ["BARE-METAL"]
+```
+
+- Override the default `supported_c_libs` property to link with the smaller C libraries. Both the ARM and GCC_ARM toolchains support optimized versions of their C standard libraries, microlib and newlib-nano. We recommend using them with the bare metal profile. For each toolchain, if there is enough memory to link with the standard library, then add the corresponding "std" library to the list. For example:
```
hugueskamba:
```suggestion
- Override the default `supported_c_libs` property to link with the smaller C libraries. Both the ARM and GCC_ARM toolchains support optimized versions of the C standard libraries, microlib and newlib-nano respectively. We recommend using them with the bare metal profile to achieve lower memory footprints. The missing features of the standard library are not essential in single threaded applications. For each toolchain, if there is enough memory to link with the standard library, then add the corresponding "std" library to the list. For example:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+
+<span class="notes">**Note:** The IAR toolchain does not have a small C library.</span>
+
+- Set the `c_lib` property to indicate which C library should be linked by the build tools. If your target has `"default_lib": "small"` defined, then replace with `"c_lib" : "small"`. Otherwise add `"c_lib" : "small"`. We recommend that ultra constrained devices running bare metal applications link with the small C libraries by default.
+
+<span class="notes">**Note:** If a toolchain does not support a small C library and `c_lib` is set to `small`, the build tools will revert to linking with the standard C library, provided that it is listed in `supported_c_libs` for the given toolchain.</span>
+
+- If `default_toolchain` is set to `uARM`, then replace with `ARM` and remove `uARM` from `supported_toolchains`. The uARM toolchain, which is the ARMC6 toolchain with the micro C library has been deprecated and it is no longer possible to use it. In the previous step, `c_lib` was set to `small` to ensure that the micro C library will be linked by the tools when using the ARM toolchain.
```
hugueskamba:
```suggestion
- If `"default_toolchain"` is set to `uARM` replace it with `ARM` and remove `uARM` from `"supported_toolchains"`. Support for the uARM toolchain, which is the ARM toolchain with microlib, has been removed. Setting `"c_lib"` to `"small"` ensures microlib is used when the ARM toolchain is selected.
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+
+<span class="notes">**Note:** As mentionned above, to successfully build with the ARM micro C library, you might need to replace the content of the TOOLCHAIN_ARM_STD scatter file with that of TOOLCHAIN_ARM_MICRO. This is because the TOOLCHAIN_ARM_MICRO linker file defines a two-region memory model which is necessary to build with microlib.</span>
```
hugueskamba:
```suggestion
<span class="notes">**Note:** As mentioned above, to successfully build with microlib, the target must define a 2-region memory model. You might need to replace the content of the TOOLCHAIN_ARM_STD linker file with that of TOOLCHAIN_ARM_MICRO. The TOOLCHAIN_ARM_MICRO directory was created to host the startup file and linker file to use with the now defunct uARM toolchain (ARM toolchain with microlib), it therefore already includes a linker file that defines a two-region memory model.</span>
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+
+- If your board does not have a low power ticker, ensure that tickless is enabled using the microsecond ticker as follows:
+```
+"overrides": {
+    "tickless-from-us-ticker": true
+}
+```
+
+- It might be necessary to reduce the stack size allocated for your target if it does not have enough RAM. The stack size is configured by setting a value for the `boot-stack-size` attribute; this value must be a multiple of 8 for alignment purposes.By default all targets are configured to have a boot stack size of 0x1000 (4096 bytes) in bare metal mode, however, this must be overridden if inadequate for your target. We recommend to reduce the boot stack size to 0x400 (1096 bytes) if your target has 8KB of RAM and to 0x300 if your target has 4KB of RAM.
```
hugueskamba:
```suggestion
- It might be necessary to reduce the stack size allocated for your target if it does not have enough RAM. The stack size is configured by setting a value for the `boot-stack-size` attribute; this value must be a multiple of 8 for alignment purposes. By default all targets are configured to have a boot stack size of 0x1000 (4096 bytes) in bare metal mode, however, this must be overridden if inadequate for your target. We recommend to reduce the boot stack size to 0x400 (1024 bytes) if your target has 8KB of RAM and to 0x300 (768 bytes) if your target has 4KB of RAM.
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+```
+"overrides": {
+    "boot-stack-size": "0x400"
+}
+```
+
+### Compile mbed-os-example-blinky-baremetal
```
hugueskamba:
```suggestion
### Build and run mbed-os-example-blinky-baremetal
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+
+Compile the example for all three toolchains: `mbed compile -m <YOUR_TARGET> -t <TOOLCHAIN>`.
```
hugueskamba:
```suggestion
Build the application and program your target: `mbed compile -m <YOUR_TARGET> -t <TOOLCHAIN> -f --sterm`.
The application running on the target should print a text on the console. Repeat for all supported toolchains.
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Validating the port
+
+To validate the bare metal target configuration, you will execute the Mbed OS greentea test suite in bare metal mode. A subset of the tests are automatically skipped by the framework because they are not supported in bare metal mode, either because the underlying functionality has not been ported to bare metal or some tests require RTOS features, for examples tests based on multi-threading.
+
+- First, change directory.
+```
+cd mbed-os
+```
+- Then execute the greentea test suite in bare metal mode for both GCC_ARM and ARM toolchains.
```
hugueskamba:
```suggestion
- Then execute the greentea test suite in bare metal mode for the supported toolchains.
```
The [PR to solve the issue with IAR](https://github.com/ARMmbed/mbed-os/pull/12691) has been merged.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: I don't think the sentence implies that an RTOS is not suitable for constrained MCUs. Nevertheless I will rephrase the sentence to remove some of the ambiguity.

evedon: The documentation says "Mbed OS bare metal profile" so I will use "profile" instead of "mode".

docs/program-setup/bare_metal/mbed2_porting.md
evedon: The original sentence is accurate because we no longer support Mbed 2. The suggested sentence could wrongly lead to believe that Mbed 2 is supported.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: I will use "property" instead of "attribute" to use teh same terminology as the documentation.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: I will use "Compile and run" as described at the top of the file.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: You need to cd to mbed-os-example-blinky-baremetal in order to compile at step 3.

hugueskamba: Yes but then mention it at step 3, no?

docs/program-setup/bare_metal/mbed2_porting.md
hugueskamba: Compile is just compile, the linker is also used. The combination of all operations is usually referred to as building.

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
+# Porting an Mbed 2 target to Mbed OS 6 bare metal
+
+The Mbed OS bare metal profile removes the RTOS and provides fewer features compared to Mbed OS 6. This profile offers the same functionality as Mbed 2 and allows targets to access features that we've added to Mbed OS 5 and 6. This document describes how to configure an Mbed 2 target to run in Mbed OS 6 bare metal and how to validate the port.
```
hugueskamba: > The Mbed OS bare metal profile removes the RTOS and provides fewer features compared to Mbed OS 6.

Bare metal is a profile to Mbed OS 6 so the comparison does not quite make sense. Perhaps just stop at «features ».

hugueskamba: > we've added to Mbed OS 5 and 6.

How about « we have added to more recent versions of Mbed OS »?

hugueskamba: > how to configure an Mbed 2 target to run in Mbed OS 6 bare metal 

How about « how to configure an Mbed 2 target to run with Mbed OS using the bare metal profile »?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+The main steps are:
+1. Clone [mbed-os-example-blinky-baremetal](https://github.com/ARMmbed/mbed-os-example-blinky-baremetal).
+1. Edit `targets.json`. In this step, you will configure your target to support Mbed OS 6 and add bare metal configuration parameters.
+1. Compile and run mbed-os-example-blinky-baremetal. In this step, you will validate the configuration and troubleshoot any issue.
+
+### Clone mbed-os-example-blinky-baremetal
+
+Run the following commands:
+1. `git clone https://github.com/ARMmbed/mbed-os-example-blinky-baremetal`
+1. `cd mbed-os-example-blinky-baremetal`
```
hugueskamba: This step has nothing to do with cloning.

docs/program-setup/bare_metal/mbed2_porting.md
hugueskamba: See the updated suggestion.
The problem with the origin text is that it gives unnecessary information that seems out of context. We are talking about the changes to make and why we are making them. The suggest text explain why the change is necessary whilst the original text explains why that attribute was present.

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+
+Find your target in `mbed-os/targets/targets.json` and update its configuration as described below.
+
+Configure your target to support Mbed OS 6:
+- Remove the `release_versions` property as it is no longer necessary to specify the list of major versions of Mbed OS that the target supports.
+```
+"release_versions": ["2"]
+```
+
+- Add the `supported_application_profiles` property to indicate that the application profile supported by this target is bare metal. This property is currently ignored by the build tools as it is not yet supported.
```
hugueskamba: Why is this attribute needed? How will it be used by the tools? Will the target that can work with an RTOS and bare metal also have it? If so, what will be the value of the attribute for those targets?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+
+<span class="notes">**Note:** If a toolchain does not support a small C library and `c_lib` is set to `small`, the build tools will revert to linking with the standard C library, provided that it is listed in `supported_c_libs` for the given toolchain.</span>
+
+- If `default_toolchain` is set to `uARM`, then replace it with `ARM` and remove `uARM` from `supported_toolchains`. Support for the uARM toolchain, which is the ARM toolchain with microlib, has been removed. Setting `c_lib` to `small` ensures that microlib is used when the ARM toolchain is selected.
+
+<span class="notes">**Note:** As mentioned above, to successfully build with microlib, the target must define a two-region memory model. You might need to replace the content of the TOOLCHAIN_ARM_STD linker file with that of TOOLCHAIN_ARM_MICRO which already includes a two-region memory model linker file.
```
hugueskamba: There is something wrong in the content of the span. Please check.

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+you might need to replace the content of the TOOLCHAIN_ARM_STD scatter file with that of TOOLCHAIN_ARM_MICRO. This is because the TOOLCHAIN_ARM_MICRO linker file defines a two-region memory model which is necessary to build with microlib.</span>
+
+- If your board does not have a low power ticker, ensure that tickless is enabled using the microsecond ticker as follows:
+```
+"overrides": {
+    "tickless-from-us-ticker": true
+}
+```
+
+- It might be necessary to reduce the stack size allocated for your target if it does not have enough RAM. The stack size is configured by setting a value for the `boot-stack-size` attribute; this value must be a multiple of 8 for alignment purposes. By default all targets are configured to have a boot stack size of 0x1000 (4096 bytes) in bare metal. However, this must be overridden if inadequate for your target. We recommend to reduce the boot stack size to 0x400 (1096 bytes) if your target has 8KB of RAM and to 0x300 (768 bytes) if your target has 4KB of RAM.
```
hugueskamba: 0x400 is 1024 bytes

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+```
+"overrides": {
+    "boot-stack-size": "0x400"
+}
+```
+
+### Compile and run mbed-os-example-blinky-baremetal
```
hugueskamba: It isn’t just compiling, so I would use Build rather than Compile.

docs/program-setup/bare_metal/mbed2_porting.md
bulislaw: How about we just say `Remove the `release_versions` property as it is no longer required`?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Clone mbed-os-example-blinky-baremetal
+
+Run the following commands:
+1. `git clone https://github.com/ARMmbed/mbed-os-example-blinky-baremetal`
+1. `cd mbed-os-example-blinky-baremetal`
+
+### Edit targets.json
+
+Find your target in `mbed-os/targets/targets.json` and update its configuration as described below.
```
bulislaw, line 31: In previous step you do the `cd` into mbed-os so this path is not correct.

docs/program-setup/bare_metal/mbed2_porting.md
bulislaw: I would drop the second part of this paragraph `This property is currently ignored by the build tools as it is not yet supported.` it doesn't add anything. And it raises questions if it's not supported why do we ask people to add it.

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+
+Configure your target to support Mbed OS 6:
+- Remove the `release_versions` property as it is no longer necessary to specify the list of major versions of Mbed OS that the target supports.
+```
+"release_versions": ["2"]
+```
+
+- Add the `supported_application_profiles` property to indicate that the application profile supported by this target is bare metal. This property is currently ignored by the build tools as it is not yet supported.
+```
+"supported_application_profiles" : ["BARE-METAL"]
```
bulislaw: Why uppercase? I would just use `baremetal`

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+- Override the default `supported_c_libs` property to link with the smaller C libraries. Both the ARM and GCC_ARM toolchains support optimized versions of the C standard libraries, microlib and newlib-nano respectively. We recommend using them with the bare metal profile to achieve lower memory footprints. For each toolchain, if there is enough memory to link with the standard library, then add the corresponding "std" library to the list. For example:
```
bulislaw: This paragraph is confusing:
>  Override the default `supported_c_libs`
What's the default? It's not obvious at this stage with what to overwrite it. Maybe starting with `Define supported variants of standard C library if different from the default.` We should also quote default below. I like rest of the description with simple explanation.

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+
+```
+"supported_c_libs": {
+    "arm":  ["std"], ["small"],
+    "gcc_arm":  ["std"], ["small"],
+    "iar": ["std"]
+}
+```
+<span class="notes">**Note:** For ARM, your target scatter file needs to have a two-region memory model. If the build system throws an error when compiling with microlib, you can find more information [here](https://os.mbed.com/docs/mbed-os/v6.0-preview/reference/using-small-c-libraries.html).</span>
```
bulislaw: What sort of error?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+
+<span class="notes">**Note:** The IAR toolchain does not have a small C library.</span>
+
+- Set the `c_lib` property to indicate which C library should be linked by the build tools. If your target has `"default_lib": "small"` defined, then replace with `"c_lib" : "small"`. Otherwise add `"c_lib" : "small"`. We recommend that ultra constrained devices running bare metal applications link with the small C libraries by default.
```
bulislaw: >should be linked by the build

should be used? The "link" is not important here and simpler words work for both the specialists and newbies

> then replace with `"c_lib" 

teh replace *it* with

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+
+<span class="notes">**Note:** If a toolchain does not support a small C library and `c_lib` is set to `small`, the build tools will revert to linking with the standard C library, provided that it is listed in `supported_c_libs` for the given toolchain.</span>
```
bulislaw: Will we get a warning?

docs/program-setup/bare_metal/mbed2_porting.md
bulislaw: Will that actually work? Just using the uARM linkerfile? What about if someone will want to use full ARM clib?

docs/program-setup/bare_metal/mbed2_porting.md
bulislaw: I'm not convinced, as far as I remember bare metal has only one stack. So the boot stack is the thread stack as well and the ISR stack.

> 0x400 (1096 bytes) if your target has 8KB of RAM 
7K for global data and heap seems excessive. Also we'd like to have the applications pretty much portable if the stacks are different between different boards. The initial idea was that boards won't be allowed to play with stack sizes, only applications will be allowed to change them.

> 0x300 (768 bytes) if your target has 4KB of RAM.

Is it even possible to run Mbed on 4k board? Considering it's probably 8-16k flash...

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,109 @@
...
+
+Troubleshoot any issue.
+
+## Validating the port
+
+To validate the bare metal target configuration, you will execute the Mbed OS greentea test suite with the bare metal configuration. A subset of the tests are automatically skipped by the framework either because the underlying functionality has not been ported to bare metal or because some tests require RTOS features, for examples more complex tests based on multi-threading.
+
+- First, change directory.
+```
+cd mbed-os
```
bulislaw: We did "compile" in the step above so we must be in the right directory.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: > > The Mbed OS bare metal profile removes the RTOS and provides fewer features compared to Mbed OS 6.
> 
> Bare metal is a profile to Mbed OS 6 so the comparison does not quite make sense. Perhaps just stop at «features ».

That does not work for me. If it has less features then we should say compared to what.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: `cd` was to mbed-os-example-blinky-baremetal not mbed-os.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: I will change the formatting then.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: Good catch.

evedon: Yes this works. ARM C library works with a two-region memory model linker file.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: Yes, the tools will issue a warning to indicate that the standard C library is used instead of the small library.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: The error is that there are undefined symbols (Image$$ARM_LIB_HEAP$$ZI$$Base, Image$$ARM_LIB_HEAP$$ZI$$Length, Image$$ARM_LIB_HEAP$$ZI$$Limit). I will list them.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: > I'm not convinced, as far as I remember bare metal has only one stack. So the boot stack is the thread stack as well and the ISR stack.
> 
> > 0x400 (1096 bytes) if your target has 8KB of RAM
> > 7K for global data and heap seems excessive. Also we'd like to have the applications pretty much portable if the stacks are different between different boards. The initial idea was that boards won't be allowed to play with stack sizes, only applications will be allowed to change them.
> 
> > 0x300 (768 bytes) if your target has 4KB of RAM.
> 
> Is it even possible to run Mbed on 4k board? Considering it's probably 8-16k flash...

You are right that in bare metal the boot stack is the thread stack as well and the ISR stack. And that stack is defined to be 4K by default in https://github.com/ARMmbed/mbed-os/blob/master/targets/targets.json/#L60. For ultra constrained devices, it needs to be lowered.

LPC1114FN28: Flash 32kB, RAM 4kB was ported to bare metal https://github.com/ARMmbed/mbed-os/pull/12623/commits/90d7e6266fe2c582f0b938bf4444dec7f425f723;  the recommendation to set `boot-stack-size` to `0x300` comes from that port.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: It isn't just compiling but the mbed command is `compile`. I don't have a strong opinion on this so I can change to build.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: In the previous step we compiled the example.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: I will rephrase;  I would like to introduce the main two parts of this page: `Configuring your target` and `Validating the port`.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: We need to override because the default for all targets does not include "small" for ARM. 
But you are right it will be cleaerer to quote the default.

evedon:
I have addressed all review comments. Could you please re-review?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
+# Porting an Mbed OS 2 target to Mbed OS 6 bare metal
+
+The Mbed OS bare metal profile removes the RTOS and provides fewer features compared to Mbed OS 6. This profile offers the same functionality as Mbed OS 2 and allows targets to access features that we have added to more recent versions of Mbed OS. This document describes how to configure an Mbed OS 2 target for bare metal and validate the port.
```
hugueskamba: functionality or functionalities?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+- Add the `supported_application_profiles` property to indicate that the application profile supported by this target is bare metal.
+```
+"supported_application_profiles" : ["bare-metal"]
+```
+
+- Override `supported_c_libs` property to link with the smaller C libraries. The default for all targets is defined as follows: 
+```
+"supported_c_libs": {
+    "arm":  ["std"],
+    "gcc_arm":  ["std"], ["small"],
```
hugueskamba:
```suggestion
    "gcc_arm":  ["std", "small"],
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+    "iar": ["std"]
+}
+```
+
+Both the ARM and GCC_ARM toolchains support optimized versions of the C standard libraries, microlib and newlib-nano respectively. We recommend using them with the bare metal profile to achieve lower memory footprints. Ultra constrained targets should override `supported_c_libs` as follows:
+```
+"supported_c_libs": {
+    "arm":  ["small"],
+    "gcc_arm":  ["small"]
```
hugueskamba: How about `"iar"`? Were Mbed 2 targets not supported by the IAR toolchain?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+}
+```
+
+For each toolchain, if there is enough memory to link with the standard library, then add the corresponding "std" library to the list. For example:
```
hugueskamba:
```suggestion
For each toolchain, if there is enough memory to link with the standard library, add the corresponding "std" library to the list. For example:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+```
+"supported_c_libs": {
+    "arm":  ["std"], ["small"],
```
hugueskamba:
```suggestion
    "arm":  ["std", "small"],
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+    "gcc_arm":  ["std"], ["small"],
```
hugueskamba:
```suggestion
    "gcc_arm":  ["std", "small"],
```

docs/program-setup/bare_metal/mbed2_porting.md
evedon: If IAR standard library is supported then it will be added in the next step.

docs/program-setup/bare_metal/mbed2_porting.md
iriark01: Probably "functionality"

hugueskamba: So Mbed 2 provides one functionality? What is it?

iriark01: Functionality is a collective term. 

"The range of operations that can be run on a computer or other electronic system: new software with additional functionality."

hugueskamba (approved):

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
+# Porting an Mbed OS 2 target to Mbed OS 6 bare metal
```
iriark01:
```suggestion
# Porting a target from Mbed OS 2 to Mbed OS 6 bare metal
```

docs/program-setup/bare_metal/mbed2_porting.md
bulislaw: OK, but this is an extreme example with target that small it doesn't make much sense to run Mbed and it makes even less sense for us to spend time on making it work. I'm afraid that this exceptions will lead to a mess when people see "oh you can do that as it's already done for this target". It would be more than reasonable to say that we require at least 32/48K of ROM and at 8K of RAM. How big is bare metal blinky these days?

docs/program-setup/bare_metal/mbed2_porting.md
iriark01:
```suggestion
The Mbed OS bare metal profile offers the same functionality as Mbed OS 2 and allows targets to access features that we have added to more recent versions of Mbed OS. At the same time, it removes the RTOS and provides fewer features than Mbed OS 6, so it's smaller and therefore suitable for ultraconstrained devices. 

This document describes how to configure an Mbed OS 2 target for bare metal, and validate the port.
```

iriark01: Note to self: come back to the "this document" sentence after I've read the full doc

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+The Mbed OS bare metal profile removes the RTOS and provides fewer features compared to Mbed OS 6. This profile offers the same functionality as Mbed OS 2 and allows targets to access features that we have added to more recent versions of Mbed OS. This document describes how to configure an Mbed OS 2 target for bare metal and validate the port.
+
+## Configuring your target
+
+You will use the blinky bare metal Mbed OS example and make it run on your target. The configuration will be validated upon successful run of the example.
```
iriark01:
```suggestion
You will get the Blinky bare metal Mbed OS example running on your target. A successful run of the example will validate the configuration.
```

iriark01: Note to self: why future tense?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+The main steps are:
+1. Clone [mbed-os-example-blinky-baremetal](https://github.com/ARMmbed/mbed-os-example-blinky-baremetal).
+1. Edit `targets.json`. In this step, you will configure your target to support Mbed OS 6 and add bare metal configuration parameters.
+1. Build and run mbed-os-example-blinky-baremetal. In this step, you will validate the configuration and troubleshoot any issue.
```
iriark01:
```suggestion
1. Build and run the example. In this step, you will validate the configuration and troubleshoot any issues.
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Clone mbed-os-example-blinky-baremetal
+
+Run the following command:
```
iriark01:
```suggestion
1. Run the following command:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+- `git clone https://github.com/ARMmbed/mbed-os-example-blinky-baremetal`
```
iriark01:
```suggestion
    `git clone https://github.com/ARMmbed/mbed-os-example-blinky-baremetal`
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+And then change directory:
```
iriark01:
```suggestion
1. Change directory:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+- `cd mbed-os-example-blinky-baremetal`
```
iriark01:
```suggestion
    `cd mbed-os-example-blinky-baremetal`
```

bulislaw (approved):
Looks good. My only concern is the stack size configuration, but we merge this as is and decide on stack later.

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Edit targets.json
+
+Find your target in `mbed-os/targets/targets.json` and update its configuration as described below.
+
+Configure your target to support Mbed OS 6:
+- Remove the `release_versions` property as it is no longer required.
```
iriark01:
```suggestion
1. Remove the `release_versions` property; it is no longer required:
```

iriark01: Note to self: all these steps need to be numbered and the code block indents fixed

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+```
+"release_versions": ["2"]
+```
+
+- Add the `supported_application_profiles` property to indicate that the application profile supported by this target is bare metal.
```
iriark01:
```suggestion
1. To indicate that the application profile supported by this target is bare metal, add the `supported_application_profiles` property:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+- Override `supported_c_libs` property to link with the smaller C libraries. The default for all targets is defined as follows: 
+```
+"supported_c_libs": {
+    "arm":  ["std"],
+    "gcc_arm":  ["std", "small"],
+    "iar": ["std"]
+}
+```
+
+Both the ARM and GCC_ARM toolchains support optimized versions of the C standard libraries, microlib and newlib-nano respectively. We recommend using them with the bare metal profile to achieve lower memory footprints. Ultra constrained targets should override `supported_c_libs` as follows:
```
iriark01:
```suggestion
Both the ARM and GCC_ARM toolchains support optimized versions of the C standard libraries - microlib and newlib-nano, respectively. We recommend using them with the bare metal profile for lower memory footprints. Ultraconstrained targets should override `supported_c_libs`:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+```
+"supported_c_libs": {
+    "arm":  ["small"],
+    "gcc_arm":  ["small"]
+}
+```
+
+For each toolchain, if there is enough memory to link with the standard library, add the corresponding "std" library to the list. For example:
```
iriark01:
```suggestion
For each toolchain, if there is enough memory to link with the standard library, add the corresponding `std` library to the list. For example:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+```
+"supported_c_libs": {
+    "arm":  ["std", "small"],
+    "gcc_arm":  ["std", "small"],
+    "iar": ["std"]
+}
+```
+<span class="notes">**Note:** For ARM, your target scatter file needs to have a two-region memory model. If the build system throws an error (presence of undefined symbols `Image$$ARM_LIB_HEAP$$ZI$$Base`, `Image$$ARM_LIB_HEAP$$ZI$$Length`, `Image$$ARM_LIB_HEAP$$ZI$$Limit`) when compiling with microlib, you can find more information [here](https://os.mbed.com/docs/mbed-os/v6.0-preview/reference/using-small-c-libraries.html).</span>
```
iriark01:
```suggestion
<span class="notes">**Note:** For ARM, your target scatter file needs to have a two-region memory model. If the build system throws an error ("presence of undefined symbols `Image$$ARM_LIB_HEAP$$ZI$$Base`, `Image$$ARM_LIB_HEAP$$ZI$$Length`, `Image$$ARM_LIB_HEAP$$ZI$$Limit`") when compiling with microlib, you can find more information [in the small C libraries documentation](https://os.mbed.com/docs/mbed-os/v6.0-preview/reference/using-small-c-libraries.html).</span>
```

iriark01: Note to self: replace with relative link

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+<span class="notes">**Note:** The IAR toolchain does not have a small C library.</span>
```
iriark01: I want to move this closer to line 46, as it's a natural question to ask about seeing GCC and ARM mentioned. Any objections?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+- Set the `c_lib` property to indicate which C library should be used by the build tools. If your target has `"default_lib": "small"` defined, then replace it with `"c_lib" : "small"`. Otherwise add `"c_lib" : "small"`. We recommend that ultra constrained devices running bare metal applications link with the small C libraries by default.
```
iriark01:
```suggestion
- We recommend that ultraconstrained devices running bare metal applications link with the small C libraries by default. To indicate which C library should be used by the build tools, set the `c_lib` property:
    - If your target has `"default_lib": "small"` defined, then replace it with `"c_lib" : "small"`. 
    - Otherwise, add `"c_lib" : "small"`. 
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+<span class="notes">**Note:** If a toolchain does not support a small C library and `c_lib` is set to `small`, the build tools will revert to linking with the standard C library, provided that it is listed in `supported_c_libs` for the given toolchain.</span>
```
iriark01:
```suggestion
<span class="notes">**Note:** If a toolchain does not support a small C library and `c_lib` is set to `small`, the build tools will revert to linking with the standard C library (provided that it is listed in `supported_c_libs` for the given toolchain).</span>
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+- If `default_toolchain` is set to `uARM`, then replace it with `ARM` and remove `uARM` from `supported_toolchains`. Support for the uARM toolchain, which is the ARM toolchain with microlib, has been removed. Setting `c_lib` to `small` ensures that microlib is used when the ARM toolchain is selected.
```
iriark01:
```suggestion
- If `default_toolchain` is set to `uARM`, then
    1. Replace it with `ARM`.
    1. Remove `uARM` from `supported_toolchains`. 
        Support for the uARM toolchain, which is the ARM toolchain with microlib, has been removed. Setting `c_lib` to `small`, as explained above, ensures that microlib is used when the ARM toolchain is selected.
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+<span class="notes">**Note:** As mentioned above, to successfully build with microlib, the target must define a two-region memory model. You might need to replace the content of the TOOLCHAIN_ARM_STD linker file with that of TOOLCHAIN_ARM_MICRO which includes a two-region memory model linker file.</span>
```
iriark01:
```suggestion
<span class="notes">**Note:** As mentioned above, to successfully build with microlib, the target must define a two-region memory model. You might need to replace the content of the `TOOLCHAIN_ARM_STD` linker file with that of `TOOLCHAIN_ARM_MICRO`, which includes a two-region memory model linker file.</span>
```

iriark01: Question: under what circumstances will I need to do this?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+- If your board does not have a low power ticker, ensure that tickless is enabled using the microsecond ticker as follows:
```
iriark01:
```suggestion
- If your board does not have a low power ticker, ensure that tickless is enabled using the microsecond ticker:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+```
+"overrides": {
+    "tickless-from-us-ticker": true
+}
+```
+
+- It might be necessary to reduce the stack size allocated for your target if it does not have enough RAM. The stack size is configured by setting a value for the `boot-stack-size` attribute; this value must be a multiple of 8 for alignment purposes. By default all targets are configured to have a boot stack size of 0x1000 (4096 bytes) in bare metal. However, this must be overridden if inadequate for your target. We recommend to reduce the boot stack size to 0x400 (1024 bytes) if your target has 8KB of RAM and to 0x300 (768 bytes) if your target has 4KB of RAM.
```
iriark01:
```suggestion
- By default, all bare metal targets are configured to have a boot stack size of 0x1000 (4,096 bytes). If this is too much for your target, you will need to override the stack allocation to match your target's available RAM. 

    The stack size is configured by setting a value for the `boot-stack-size` attribute; this value must be a multiple of 8 for alignment purposes. We recommend that you reduce the boot stack size to 0x400 (1,024 bytes) if your target has 8KB of RAM and to 0x300 (768 bytes) if your target has 4KB of RAM.
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+```
+"overrides": {
+    "boot-stack-size": "0x400"
+}
+```
+
+### Build and run mbed-os-example-blinky-baremetal
```
iriark01:
```suggestion
### Build and run the Blinky bare metal example
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+Build the application and program your target: `mbed compile -m <YOUR_TARGET> -t <TOOLCHAIN> --flash --sterm`.
```
iriark01:
```suggestion
Build the application and program your target: 

```
mbed compile -m <YOUR_TARGET> -t <TOOLCHAIN> --flash --sterm
```
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+The application running on the target should print a text on the console. Repeat for all supported toolchains.
+
+Troubleshoot any issue.
```
iriark01, line 121: Can we point them to useful resources for that?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Validating the port
+
+To validate the bare metal target configuration, you will execute the Mbed OS greentea test suite with the bare metal configuration. A subset of the tests are automatically skipped by the framework either because the underlying functionality has not been ported to bare metal or because some tests require RTOS features, for examples more complex tests based on multi-threading.
```
iriark01:
```suggestion
To validate the bare metal target configuration, execute the Mbed OS Greentea test suite with the bare metal profile. This profile causes Greentea to skip a subset of the tests, either because the underlying functionality has not been ported to bare metal or because some tests require RTOS features (for examples, more complex tests based on multi-threading).
```

iriark01: I changed "configuration" to "port" because there were two configurations in the paragraph and it can be confusing on first reading. You used "port" at the beginning of the document so I assumed it's okay.

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+- First, change directory.
```
iriark01: Note to self: numbered steps + indent fixes

iriark01:
```suggestion
1. Change directory:
```

docs/program-setup/bare_metal/mbed2_porting.md
iriark01: Since I changed directory a fairly long time ago, should we add a step for making sure I'm in the right directory?

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+```
+cd mbed-os
+```
+- Then execute the greentea test suite with the bare metal configuration for the supported toolchains.
```
iriark01:
```suggestion
1. Execute the Greentea test suite with the bare metal configuration for the supported toolchains:
```

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+```
+mbed test -m <YOUR_TARGET> -t <TOOLCHAIN> --app-config TESTS/configs/baremetal.json
+```
+It can be useful to append `--compile` and fix build issues first before running tests with `--run`.
```
iriark01:
```suggestion
<span class="tips">**Tip:** You can append `--compile` and fix build issues before running tests with `--run`.</span>
```

iriark01: Neither compile nor run are in the command we show, which makes me wonder whether we meant to say "append one or the other". If so, I'd rather show both commands in full (people like copy/pasting).

docs/program-setup/bare_metal/mbed2_porting.md
```diff
@@ -0,0 +1,125 @@
...
+
+All tests should pass (or be automatically skipped).
```
iriark01: Can we point them somewhere with more information in case the tests fail?

docs/program-setup/bare_metal/mbed2_porting.md
iriark01: Do we need to make sure they have Greentea properly set up, or is it part of something we know they have?

iriark01 (changes requested):
I left some questions and suggestions. 
Where it says "note to self" it's for me to do as a tidy-up before a merge.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: That's fine.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: You will need to do this if there is a compilation error as mentioned above (line 62). This is a bit of a repeat and could be combined with the note above.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: I don't think we have any troubleshooting guide.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: This is the main directory for the example so I don't think it is required.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: I will put back configuration but then use bare metal "profile".

docs/program-setup/bare_metal/mbed2_porting.md
evedon: I don't think we have a troubleshooting guide.

docs/program-setup/bare_metal/mbed2_porting.md
evedon: We coud link to https://os.mbed.com/docs/mbed-os/v5.15/tools/greentea-testing-applications.html

iriark01:
Waiting on 6-beta release to merge and publish.

evedon:
Code related PRs are all merged to master so you can merge this PR.

PR #1294, opened 2020-04-16
Update CircleCI config for branch "v5.15"

micque01:
Pull latest config changes from https://github.com/ARMmbed/docs-ci-config.git

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -9,10 +9,10 @@ references:
 executors:
   doc_builder:
     docker:
-      - image: mbed/doc_builder:latest
-        auth:
-          username: $DOCKER_USER
-          password: $DOCKER_PASS
+      - image: $AWS_QA_ACCOUNT_ID.dkr.ecr.us-west-2.amazonaws.com/otg/doc-builder:latest
+        aws_auth:
+          aws_access_key_id: $AWS_QA_ACCESS_KEY_ID
+          aws_secret_access_key: $AWS_QA_SECRET_KEY
 
 commands:
   setup_aws_credentials:


micque01:
@AnotherButler @iriark01 The Travis tests are failing, I'm not sure what is causing this. Is this a blocker to getting these PRs merged?

iriark01:
I think you can merge, since you only worked on CircleCI so it can't be this PR that broke it.

@jamesbeyond can you help me figure out why Travis is failing?

jamesbeyond:
> I think you can merge, since you only worked on CircleCI so it can't be this PR that broke it.
> 
> @jamesbeyond can you help me figure out why Travis is failing?

Hi @iriark01 ,
The reason that travis is failing is because it checks out the latest mbed-os, instead of using 5.15 version of mbed-os:
https://github.com/ARMmbed/mbed-os-5-docs/blob/v5.15/.travis.yml#L26
some of the mbed-os API deprecations, the 5.15 version doc code snippets not compatible with  latest mbed-os any more. the solution would be adding `git checkout mbed-os-5.15 ` after that line

iriark01:
Thank you @jamesbeyond . Is that something you can do, or do I need to get another team invovled?

jamesbeyond:
> Thank you @jamesbeyond . Is that something you can do, or do I need to get another team invovled?

Maybe @micque01 could you make that fix in this PR ? it is failing at the moment, 

you just need to put
`git checkout mbed-os-5.15`
after
https://github.com/ARMmbed/mbed-os-5-docs/blob/v5.15/.travis.yml#L26

micque01:
Hi @jamesbeyond, thank you for the suggestion for the fix. However I am reluctant to fix it here as it is outside the scope of this PR and the maintenance of these tests is not in the web team's remit. A similar fix also needs to be applied to other branches v5.11, v5.12, v5.13, and v5.14, as it seems it is broken there too.

Is there a team responsible for the maintenance of these tests?

iriark01:
Merging, since Travis failure shouldn't impact CircleCI configs

PR #1301, opened 2020-04-20
fix travis for 5.15 branch

jamesbeyond:
This is the fix for https://github.com/ARMmbed/mbed-os-5-docs/pull/1294
@iriark01


iriark01:
Thank you so much!

PR #1306, opened 2020-05-01
Wait APIs

evedon:
Re-introduced the wait API.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -195,7 +195,7 @@
                                     "path": "docs/api/rtos/Kernel.md"
                                 },
                                 {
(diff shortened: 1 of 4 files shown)


iriark01:
@evedon this used to be in Platform, but seems very RTOS-related. I'm not sure where to put it.

evedon:
Leave it in Platform. Or we create a `Time` category.

iriark01:
Merging; the final structure of the API hasn't been determined but so long as it's in the JSON and the API index it won't be forgotten.

PR #1307, opened 2020-05-04
Update link to DAPLINK target porting guide

desmond-blue:
The current link to DAPLINK target porting guide is incorrect, fix it by link it to https://github.com/ARMmbed/DAPLink/blob/master/docs/PORT_TARGET_FAMILY.md


iriark01:
Thank you. I'll add it to the live versions.

PR #1318, opened 2020-05-18
Add intro to Execution

gpsimenos:


docs/reference/runtime/Execution.md
```diff
@@ -1,5 +1,9 @@
 # Execution
 
+Embedded systems are systems that are built for performing a sole dedicated duty. They are typically constrained in resources and contain only the necessary components to perform their duty. Most embedded systems are powered by a single microcontroller chip. That microcontroller is typically also limited in resources such as memory. Software for an embedded system must be carefully designed to make the best use of such limited resources. Mbed OS is often too large for such systems in its default configuration, but it supports additional modes that might be a better option depending on your target system.
```
hugueskamba: Regarding "...supports additional modes that...".
There are only two modes "Full profile" and "Bare metal profile". Since the first part of the sentence talks about the full profile, there is only one additional mode.
Should the sentence perhaps say:
"...supports a mode that..."

hugueskamba (commented):
LGTM aside from the comment below.

docs/reference/runtime/Execution.md
gpsimenos: Thanks for the review, now fixed.

hugueskamba (approved):

iriark01:
Thanks for the info about Threads - it's really helpful.

Will you be able to provide an intro to the whole Program Setup book, giving a high-level view of why we're covering the selected concepts, serial comms, build profiles and rules, the config system and target config? I would like the collection of topics to make sense to users.


https://os.mbed.com/docs/mbed-os/development/porting/porting-bootstrap.html

docs/reference/runtime/Execution.md
```diff
@@ -1,5 +1,9 @@
 # Execution
 
+Embedded systems are systems that are built for performing a sole dedicated duty. They are typically constrained in resources and contain only the necessary components to perform their duty. Most embedded systems are powered by a single microcontroller chip. That microcontroller is typically also limited in resources such as memory. Software for an embedded system must be carefully designed to make the best use of such limited resources. Mbed OS is often too large for such systems in its default configuration, but support a bare metal mode that might be a better option depending on your target system.
```
iriark01:
```suggestion
Embedded systems are dedicated to performing a single duty. They are typically constrained in resources and contain only the necessary components to perform that duty. Most embedded systems are powered by a single microcontroller chip, which is typically limited in resources such as memory. Software for an embedded system must be carefully designed to make the best use of these limited resources. Mbed OS in its default configuration is often too large for constrained systems, but supports a bare metal mode that might be a better option for your target system.
```

iriark01: - I'm not sure what the relationship between the two paragraphs is. The content about constrained systems isn't related to the Threading, and the idea that some systems can't handle Threading is mentioned in both paragraphs.
- Bare metal is explained in detail in its own book, and in the general intro to Mbed OS.

In short, I think you don't need the first paragraph.

docs/reference/runtime/Execution.md
```diff
@@ -1,5 +1,9 @@
...
+
+Threading is a fundamental technique in software development, which allows multiple tasks to run concurrently. As microcontrollers become more capable, it has become possible to use threading in embedded systems, using schedulers, either standalone or included in a real time operating system (RTOS). In single-microcontroller embedded systems, a thread is an independent segment of the program which executes within the single process runnning on the microcontroller. Threads provide a lot of flexibility to the developer but come at the cost of resources for the scheduler. Mbed, in its default configuration, provides an RTOS to enable threading in applications where resources are not critical. There is also a bare metal profile which is optimized for constained targets and does not include the RTOS or threading capabilities.
```
iriark01:
```suggestion
In single-microcontroller embedded systems, a thread is an independent segment of the program that executes within the single process running on the microcontroller. Threading is a fundamental technique in software development, which allows multiple tasks to run concurrently using schedulers, either standalone or included in a real time operating system (RTOS). Threads provide a lot of flexibility to the developer but come at the cost of resources for the scheduler. Mbed OS, in its default full profile, provides an RTOS for applications where resources are not critical. There is also a bare metal profile, which is optimized for constrained targets and does not include the RTOS or threading capabilities.
```

iriark01 (changes requested):

gpsimenos:
Thanks for the review @iriark01. I agree we should add an intro to Program Setup, can you please open a ticket in Jira for this?

iriark01:
https://jira.arm.com/browse/IOTCORE-1567

docs/reference/runtime/Execution.md
```diff
@@ -1,5 +1,9 @@
...
+
+In single-microcontroller embedded systems, a thread is an independent segment of the program that executes within the single process running on the microcontroller. Threading is a fundamental technique in software development, which allows multiple tasks to run concurrently using schedulers, either standalone or included in a real time operating system (RTOS). Threads provide a lot of flexibility to the developer but come at the cost of resources for the scheduler. Mbed OS, in its default full profile, provides an RTOS for applications where resources are not critical. There is also a bare metal profile, which is optimized for constrained targets and does not include the RTOS or threading capabilities.
```
evedon, line 5: It would be useful to be more specific about Mbed OS does.
Regarding thread scheduling, you could mention that there are two types:
- Cooperative
   - Active thread needs to yield execution for another thread to run
   - Not supported by Mbed OS
- Preemptive
   - Central piece of software (scheduler) responsible for picking and running the threads
   - Different algorithms can be used
     - Priority based round robin is used by Mbed OS
     - Threads are executed in equal time slots according to their priority

evedon, line 5: Include a link to bare metal.
You could also explain that in bare metal there is no scheduler (by default) and all activities are interrupt driven.

docs/reference/runtime/Execution.md
```diff
@@ -1,5 +1,9 @@
 # Execution
 
+Embedded systems are dedicated to performing a single duty. They are typically constrained in resources and contain only the necessary components to perform that duty. Most embedded systems are powered by a single microcontroller chip, which is typically limited in resources such as memory. Software for an embedded system must be carefully designed to make the best use of these limited resources. Mbed OS in its default configuration is often too large for constrained systems, but supports a bare metal mode that might be a better option for your target system.
```
evedon, line 3: I don't particularly like the first sentence. Embedded systems can be very complex.

I would rewrite the last sentence to mention the `full profile` versus the `bare metal profile` and emphasize that the latter is targeted at ultraconstrained devices.

evedon (changes requested):

iriark01:
Please rebase to `development` - 6.0.0-preview is a public, frozen branch for the alpha/beta releases

gpsimenos:
Replaced by https://github.com/ARMmbed/mbed-os-5-docs/pull/1331

PR #1321, opened 2020-05-19
(Preceding snippet PR) Bare metal applications with microlib should not return from main()

LDong-Arm:
REQUIRES https://github.com/ARMmbed/mbed-os-examples-docs_only/pull/116

Arm microlib cannot handle exit from `main()`: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0475f/BAJHEGCE.html

This is not an issue in the RTOS mode in which the main thread is one of the several threads created on boot. But a bare-metal application needs to explicitly avoid returning from `main()`. This PR documents this fact and suggests ways around it. 

Even though it affects microlib only, I added the advice to the generic section which is easier to notice. And an application should in principle be compatible with any configurations.

### DIFF
--- a/docs/bare_metal/c_small_libs.md
+++ b/docs/bare_metal/c_small_libs.md
@@ -24,18 +24,18 @@ This links your application with `microlib` for the `ARM` toolchain and `newlib-
 
 ### Non-returning main() required
 
-Exiting from `main()` is not supported by [Arm microlib](#arm-microlib) and causes a bare metal application to crash. Here we show two ways to prevent this.
+Arm microlib doesn't support exiting from `main()`; attempting to exit from `main()` causes a bare metal application to crash. Here we show two ways to prevent this.
 
 #### Sleep in a loop
 
-One recommended technique is sleep in a loop at the end of `main()`:
+One recommended technique is to sleep in a loop at the end of `main()`:
 ```
 while (true) {
     sleep();
 }
 ```
 
-This is energy-efficient compared to an empty `while (true) {}` loop which keeps the processor running. A loop is still needed, because `sleep()` returns after the system is woken up by an interrupt.
+This is energy-efficient compared to an empty `while (true) {}` loop, which keeps the processor running. A loop is still needed, because `sleep()` returns after the system is woken up by an interrupt.
 
 #### Dispatching an EventQueue
 
--- a/docs/bare_metal/c_small_libs.md
+++ b/docs/bare_metal/c_small_libs.md
@@ -66,7 +66,7 @@ After you have completed the steps above, add `small` to the `supported_c_libs` 
 
 ### Non-returning main()
 
-Arm microlib doesn't support exiting from `main()`; attempting to exit from `main()` causes a bare metal application to crash. Here we show two ways to prevent this.
+Arm microlib doesn't support returning from `main()`; attempting to return from `main()` causes a bare metal application to crash. Here we show two ways to prevent this.
 
 #### Sleeping in a loop
 
--- a/docs/bare_metal/c_small_libs.md
+++ b/docs/bare_metal/c_small_libs.md
@@ -89,4 +89,4 @@ The call to `queue.dispatch_forever()` never returns, as long as we don't break 
 
 ### Note on uARM toolchain
 
-The uARM toolchain is the ARMC6 toolchain with the Arm microlib, the C micro-library. This toolchain will be deprecated after 5.15.
+The uARM toolchain is the ARMC5 toolchain with the Arm microlib, the C micro-library. This toolchain will be deprecated after 5.15.
--- a/docs/bare_metal/c_small_libs.md
+++ b/docs/bare_metal/c_small_libs.md
@@ -22,7 +22,7 @@ You can build with the smaller C libraries by creating an `mbed_app.json` with t
 
 This links your application with `microlib` for the `ARM` toolchain and `newlib-nano` for the `GCC_ARM` toolchain.
 
-<span class="notes">**Note:** You bare-metal application should _not_ return from `main()` if it uses Arm `microlib`. Please see [Non-returning main()](#non-returning-main) for advice.</span>
+<span class="notes">**Note:** If your application uses the Arm microlib, it should not return from `main()`. Please see [non-returning main() below](#non-returning-main) for advice.</span>
 
 ## Newlib-nano
 
@@ -77,7 +77,7 @@ while (true) {
 }
 ```
 
-This is energy-efficient compared to an empty `while (true) {}` loop, which keeps the processor running. A loop is still needed, because `sleep()` returns after the system is woken up by an interrupt.
+This is energy efficient compared to an empty `while (true) {}` loop, which keeps the processor running. A loop is still needed, because `sleep()` returns after the system is woken up by an interrupt.
 
 #### Dispatching an EventQueue
 
@@ -85,7 +85,7 @@ If your application is based on an `EventQueue`, dispatching it at the end of `m
 
 [![View code](https://www.mbed.com/embed/?url=https://github.com/ARMmbed/mbed-os-examples-docs_only/blob/master/APIs_RTOS/EventQueue_ex_2/)](https://github.com/ARMmbed/mbed-os-examples-docs_only/blob/master/APIs_RTOS/EventQueue_ex_2/main.cpp)
 
-The call to `queue.dispatch_forever()` never returns, as long as we don't break the dispatch anywhere. The `EventQueue` class puts the system to sleep to save energy between events.
+The call to `queue.dispatch_forever()` never returns, as long as you don't break the dispatch anywhere. The `EventQueue` class puts the system to sleep to save energy between events.
 
 ### Note on uARM toolchain
 
--- a/docs/bare_metal/c_small_libs.md
+++ b/docs/bare_metal/c_small_libs.md
@@ -89,4 +89,4 @@ The call to `queue.dispatch_forever()` never returns, as long as you don't break
 
 ### Note on uARM toolchain
 
-The uARM toolchain is the ARMC5 toolchain with the Arm microlib, the C micro-library. This toolchain will be deprecated after 5.15.
+The uARM toolchain is the ARMC6 toolchain with the Arm microlib, the C micro-library. This toolchain will be deprecated after 5.15.
--- a/docs/bare_metal/c_small_libs.md
+++ b/docs/bare_metal/c_small_libs.md
@@ -29,7 +29,7 @@ Exiting from `main()` is not supported by [Arm microlib](#arm-microlib) and caus
 #### Sleep in a loop
 
 One recommended technique is sleep in a loop at the end of `main()`:
-```C
+```
 while (true) {
     sleep();
 }
--- a/docs/bare_metal/c_small_libs.md
+++ b/docs/bare_metal/c_small_libs.md
@@ -22,6 +22,29 @@ You can build with the smaller C libraries by creating an `mbed_app.json` with t
 
 This links your application with `microlib` for the `ARM` toolchain and `newlib-nano` for the `GCC_ARM` toolchain.
 
+### Non-returning main() required
+
+Exiting from `main()` is not supported by [Arm microlib](#arm-microlib) and causes a bare metal application to crash. Here we show two ways to prevent this.
+
+#### Sleep in a loop
+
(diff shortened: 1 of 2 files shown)


docs/bare_metal/c_small_libs.md
```diff
@@ -22,6 +22,29 @@ You can build with the smaller C libraries by creating an `mbed_app.json` with t
...
+
+This is energy-efficient compared to an empty `while (true) {}` loop which keeps the processor running. A loop is still needed, because `sleep()` returns after the system is woken up by an interrupt.
+
+#### Dispatching an EventQueue
+
+If your application is based on an `EventQueue`, dispatching it at the end of `main()` works as well:
+
+[![View code](https://www.mbed.com/embed/?url=https://github.com/ARMmbed/mbed-os-examples-docs_only/blob/master/APIs_RTOS/EventQueue_ex_2/)](https://github.com/ARMmbed/mbed-os-examples-docs_only/blob/master/APIs_RTOS/EventQueue_ex_2/main.cpp)
+
+The call to `queue.dispatch()` never returns, as long as we don't break the dispatch anywhere. The `EventQueue` class puts the system to sleep to save energy between events.
```
hugueskamba: Should we mention dispach_forever() instead?

docs/bare_metal/c_small_libs.md
```diff
@@ -22,6 +22,29 @@ You can build with the smaller C libraries by creating an `mbed_app.json` with t
 
 This links your application with `microlib` for the `ARM` toolchain and `newlib-nano` for the `GCC_ARM` toolchain.
 
+### Non-returning main() required
+
+Exiting from `main()` is not supported by [Arm microlib](#arm-microlib) and causes a bare metal application to crash. Here we show two ways to prevent this.
```
hugueskamba: I don't see why someone would want to return from main() in a microcontroller especially if the is no OS.

I think the reason we are not explaining the lack of argument passed to main() is also reason enough not mention returning from main().

docs/bare_metal/c_small_libs.md
LDong-Arm: `dispatch_forever()` is better, just the code snippet (in a separate repo) has `dispatch()`. Probably I should make a PR to update the snippet.

docs/bare_metal/c_small_libs.md
LDong-Arm: The only case I can think of is when an application does everything in interrupts instead of `main()`. @jenia81: Could you confirm this was your case?

docs/bare_metal/c_small_libs.md
LDong-Arm: Done

docs/bare_metal/c_small_libs.md
evedon: It is useful to add this information which was raised as an issue.

evedon (approved):

LDong-Arm:
@evedon The Travis failure was again caused by the JSON snippet. I rebased this PR, it should go away now.

Update: Now the JSON issue goes away, C snippet fails. I'll remove the "c" as well...

docs/bare_metal/c_small_libs.md
```diff
@@ -22,6 +22,29 @@ You can build with the smaller C libraries by creating an `mbed_app.json` with t
 
 This links your application with `microlib` for the `ARM` toolchain and `newlib-nano` for the `GCC_ARM` toolchain.
 
+### Non-returning main() required
```
iriark01: First question - if this is only for Arm microlib, why is it not in that section?

docs/bare_metal/c_small_libs.md
iriark01:
```suggestion
Arm microlib doesn't support exiting from `main()`; attempting to exit from `main()` causes a bare metal application to crash. Here we show two ways to prevent this.
```

docs/bare_metal/c_small_libs.md
```diff
@@ -22,6 +22,29 @@ You can build with the smaller C libraries by creating an `mbed_app.json` with t
...
+
+Exiting from `main()` is not supported by [Arm microlib](#arm-microlib) and causes a bare metal application to crash. Here we show two ways to prevent this.
+
+#### Sleep in a loop
+
+One recommended technique is sleep in a loop at the end of `main()`:
```
iriark01:
```suggestion
One recommended technique is to sleep in a loop at the end of `main()`:
```

docs/bare_metal/c_small_libs.md
```diff
@@ -22,6 +22,29 @@ You can build with the smaller C libraries by creating an `mbed_app.json` with t
...
+```
+while (true) {
+    sleep();
+}
+```
+
+This is energy-efficient compared to an empty `while (true) {}` loop which keeps the processor running. A loop is still needed, because `sleep()` returns after the system is woken up by an interrupt.
```
iriark01:
```suggestion
This is energy-efficient compared to an empty `while (true) {}` loop, which keeps the processor running. A loop is still needed, because `sleep()` returns after the system is woken up by an interrupt.
```

iriark01 (changes requested):

docs/bare_metal/c_small_libs.md
LDong-Arm: The section here (how to enable small c_lib) is more noticeable. I was afraid if this is added to the Arm microlib section below, app developers may not always notice it.

Maybe I can move it to the microlib section and add a note here to point the user to have a look.

docs/bare_metal/c_small_libs.md
```diff
@@ -62,6 +64,29 @@ After you have completed the steps above, add `small` to the `supported_c_libs`
...
+If your application is based on an `EventQueue`, dispatching it at the end of `main()` works as well:
+
+[![View code](https://www.mbed.com/embed/?url=https://github.com/ARMmbed/mbed-os-examples-docs_only/blob/master/APIs_RTOS/EventQueue_ex_2/)](https://github.com/ARMmbed/mbed-os-examples-docs_only/blob/master/APIs_RTOS/EventQueue_ex_2/main.cpp)
+
+The call to `queue.dispatch_forever()` never returns, as long as we don't break the dispatch anywhere. The `EventQueue` class puts the system to sleep to save energy between events.
+
 ### Note on uARM toolchain
 
-The uARM toolchain is the ARMC6 toolchain with the Arm microlib, the C micro-library. This toolchain will be deprecated after 5.15.
+The uARM toolchain is the ARMC5 toolchain with the Arm microlib, the C micro-library. This toolchain will be deprecated after 5.15.
```
LDong-Arm: Fixing a pretty major typo...

LDong-Arm:
@iriark01 I've moved the paragraphs to the Arm microlib section, please review

docs/bare_metal/c_small_libs.md
```diff
@@ -22,6 +22,8 @@ You can build with the smaller C libraries by creating an `mbed_app.json` with t
 
 This links your application with `microlib` for the `ARM` toolchain and `newlib-nano` for the `GCC_ARM` toolchain.
 
+<span class="notes">**Note:** You bare-metal application should _not_ return from `main()` if it uses Arm `microlib`. Please see [Non-returning main()](#non-returning-main) for advice.</span>
```
iriark01:
```suggestion
<span class="notes">**Note:** If your application uses the Arm microlib, it should not return from `main()`. Please see [non-returning main() below](#non-returning-main) for advice.</span>
```

docs/bare_metal/c_small_libs.md
```diff
@@ -62,6 +64,29 @@ After you have completed the steps above, add `small` to the `supported_c_libs`
 }
 ```
 
+### Non-returning main()
+
+Arm microlib doesn't support exiting from `main()`; attempting to exit from `main()` causes a bare metal application to crash. Here we show two ways to prevent this.
```
iriark01: So we say both "exit" and "return" on this page. We should stick to one.

docs/bare_metal/c_small_libs.md
```diff
@@ -62,6 +64,29 @@ After you have completed the steps above, add `small` to the `supported_c_libs`
...
+#### Sleeping in a loop
+
+One recommended technique is to sleep in a loop at the end of `main()`:
+```
+while (true) {
+    sleep();
+}
+```
+
+This is energy-efficient compared to an empty `while (true) {}` loop, which keeps the processor running. A loop is still needed, because `sleep()` returns after the system is woken up by an interrupt.
```
iriark01:
```suggestion
This is energy efficient compared to an empty `while (true) {}` loop, which keeps the processor running. A loop is still needed, because `sleep()` returns after the system is woken up by an interrupt.
```

docs/bare_metal/c_small_libs.md
```diff
@@ -62,6 +64,29 @@ After you have completed the steps above, add `small` to the `supported_c_libs`
...
+
+#### Dispatching an EventQueue
+
+If your application is based on an `EventQueue`, dispatching it at the end of `main()` works as well:
+
+[![View code](https://www.mbed.com/embed/?url=https://github.com/ARMmbed/mbed-os-examples-docs_only/blob/master/APIs_RTOS/EventQueue_ex_2/)](https://github.com/ARMmbed/mbed-os-examples-docs_only/blob/master/APIs_RTOS/EventQueue_ex_2/main.cpp)
+
+The call to `queue.dispatch_forever()` never returns, as long as we don't break the dispatch anywhere. The `EventQueue` class puts the system to sleep to save energy between events.
```
iriark01:
```suggestion
The call to `queue.dispatch_forever()` never returns, as long as you don't break the dispatch anywhere. The `EventQueue` class puts the system to sleep to save energy between events.
```

docs/bare_metal/using_bare_metal.md
```diff
@@ -9,6 +9,8 @@ Here is a code snippet that can work for both Mbed OS profiles; it prints text a
 
 [![View code](https://www.mbed.com/embed/?url=https://github.com/ARMmbed/mbed-os-examples-docs_only/blob/master/APIs_RTOS/EventQueue_ex_2/)](https://github.com/ARMmbed/mbed-os-examples-docs_only/blob/master/APIs_RTOS/EventQueue_ex_2/main.cpp)
 
+<span class="notes">**Note:** To be compatible with Arm microlib, a bare metal application should not return from `main()`. In this example the `queue.dispatch_forever()` call never returns. For more details, see [Non-returning main() required](../baremetal/using-small-c-libraries.html#non-returning-main-required).</span>
```
iriark01:
```suggestion
<span class="notes">**Note:** To be compatible with Arm microlib, a bare metal application should not return from `main()`. In this example, the `queue.dispatch_forever()` call never returns. For more details, see [Non-returning main()](../baremetal/using-small-c-libraries.html#non-returning-main).</span>
```

iriark01: For the record, I don't know if the link will work when the header has `()`. Please open all future PRs from a branch, not a fork, so I can have a preview.

iriark01 (changes requested):

docs/bare_metal/c_small_libs.md
LDong-Arm: I will use "return"

docs/bare_metal/using_bare_metal.md
LDong-Arm: My VS Code previewer works with `()`, hopefully it's standard. I'll use branches for future PRs.

iriark01 (approved):

iriark01:
Not sure how many people need to approve it; let me know when it's okay to merge.

docs/bare_metal/c_small_libs.md
evedon: This was correct. Originally uARM toolchain was using ARMC5 but since ARMC5 was deprecated it was using ARMC6 https://github.com/ARMmbed/mbed-os/blob/master/tools/build_api.py/#L199

LDong-Arm: Ah didn't know there was a quirk to point uARM to ARMC6. I'll change this sentence back.

evedon (approved):

evedon:
This is ready for merge

LDong-Arm:
@evedon This depends on the code snippet https://github.com/ARMmbed/mbed-os-examples-docs_only/pull/116. We need to get that in as well.

PR #1324, opened 2020-05-21
Bare metal tutorial: do not label/syntax-check partial JSON excerpt

LDong-Arm:
This should fix the syntax check failure.

### DIFF
--- a/docs/bare_metal/using_bare_metal.md
+++ b/docs/bare_metal/using_bare_metal.md
@@ -54,7 +54,7 @@ This example depends on the `EventQueue` class, so you need to add the library t
 
     For example: `mbed-os/events/mbed_lib.json`:
 
-    ```json
+    ```
     {
         "name": "events",
         "config": {


docs/bare_metal/using_bare_metal.md
```diff
@@ -54,15 +54,7 @@ This example depends on the `EventQueue` class, so you need to add the library t
...
-    ```json
-    {
-        "name": "events",
-        "config": {
-            "present": 1,
-            ...
-        }
-    }
-    ```
+    [![View code](https://www.mbed.com/embed/?url=https://github.com/ARMmbed/mbed-os/blob/master/events/)](https://github.com/ARMmbed/mbed-os/blob/master/events/mbed_lib.json)
```
LDong-Arm: I'm not sure if embedding a file from mbed-os repo is ideal, and whether we should use a fixed revision instead of master. Any better ideas?

iriark01: Fixed version is better, since we won't necessarily want the same file for 6.1. But I don't think we normally take files from mbed-os. @evedon might know.

evedon: I liked the version without the link better. 
You are only looking for the "name" field so if you can't use "..." then only keep the first line.

iriark01: Sorry, is this what's been breaking Travis?
This doesn't look like a code snippet that should work, just a simple example. In this case I think you should just remove `json` so Travis doesn't run on it.
I got confused and thought Travis was failing because of the bare metal example. Sorry about that.

LDong-Arm: Or maybe we can combine with the line above which becomes:

> For example, in `mbed-os/events/mbed_lib.json`: `"name": "events"`.

iriark01: That's not as readable as the first version, where you broke it into lines. 
If you don't use `JSON` Travis will just ignore it.

LDong-Arm: Done

evedon (approved):

PR #1334, opened 2020-06-02
A serial driver should be installed on Windows 7 only

### LINKED ISSUE #1332: Remove/update references to the Windows serial driver
# Docs issue template

The Serial driver is no longer needed on Windows 10.
In fact, it creates a problem when trying to use WebUSB.
See https://github.com/ARMmbed/DAPLink/issues/487
I also hit this problem myself and it took me a while to find out.

## Page link

https://os.mbed.com/docs/mbed-os/v5.15/tutorials/windows-serial-driver.html
And possibly many more that refer to windows serial driver.

## Expected to find...

Details that the driver is needed for old versions of Windows (7 and older) and it's not required for Windows 10.
Discourage the installation of the serial driver.

## Found

Recommendations to install the driver on Windows

---

FYI @iriark01 @thegecko @arekzaluski

LDong-Arm:
Fixes #1332 

This PR
* stresses on the serial driver doc page that it only applies to Windows 7
* update other docs that link to the serial driver doc page

### DIFF
--- a/docs/program-setup/serial/serial_communication.md
+++ b/docs/program-setup/serial/serial_communication.md
@@ -2,7 +2,7 @@
 
 The Arm Mbed microcontroller on your board can communicate with a host PC over the same USB cable that you use for programming.
 
-<span class="tips">If you're working on Windows earlier than Windows 10, you might need to [install a serial driver](../program-setup/windows-serial-driver.html).</span>
+<span class="tips">If you're working on Windows 7, you might need to [install a serial driver](../program-setup/windows-serial-driver.html).</span>
 
 This allows you to:
 
--- a/docs/program-setup/serial/serial_driver.md
+++ b/docs/program-setup/serial/serial_driver.md
@@ -1,8 +1,10 @@
 # Windows serial driver
 
+<span class="warnings">**Warning**: This page is for **Windows 7 only**. Do NOT install the serial driver on Windows 8 or above - serial ports work out of the box.</span>
+
 You can connect your board to your computer over USB. This works out of the box on Linux and macOS. If you are using Windows, you may need to install a serial port driver:
 
-1. Download the [Arm Mbed Windows serial port driver](http://os.mbed.com/media/downloads/drivers/mbedWinSerial_16466.exe).
+1. Download the [Arm Mbed Windows serial port driver](http://os.mbed.com/media/downloads/drivers/mbedWinSerial_16466.exe) (**Windows 7 only**).
 1. Plug in your Arm Mbed device over USB. It mounts as an Mbed drive.
 1. Close all Explorer windows showing the Mbed drive.
 1. Run the installer. This may take some time or display a few "unsigned driver" warnings.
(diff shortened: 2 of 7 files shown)


LDong-Arm:
The wording is "Windows 7" rather than "earlier than Windows 8" (which may cause confusion - is it inclusive of 8?) or "Windows 7 or earlier" (I don't think we support Vista, etc.), does it sound okay?

MarceloSalazar:
I assume we don't want to actively advertise the need of the driver but keep the content as it's searchable. So maybe references from lora, dap and possibly UnbufferedSerial/BufferedSerial could be removed.

Note most mbed-os-example-*** apps use the serial port but we're not advertising the need for the driver (I don't think we should).

LDong-Arm:
@MarceloSalazar Completely agree. I just removed links to the serial driver doc from pages other than intro and serial communication which are directly relevant.

MarceloSalazar (approved):
Looks good - thanks!

JoeTheGuitarist (approved):

docs/program-setup/serial/serial_driver.md
```diff
@@ -1,8 +1,10 @@
 # Windows serial driver
 
+<span class="warnings">**Warning**: This page is for **Windows 7 only**. Do NOT install the serial driver on Windows 8 or above - serial ports work out of the box.</span>
```
iriark01:
```suggestion
<span class="warnings">**Warning**: This page is for **Windows 7 only**. Do not install the serial driver on Windows 8 or newer; serial ports work out of the box with these versions.</span>
```

docs/program-setup/serial/serial_communication.md
```diff
@@ -2,7 +2,7 @@
 
 The Arm Mbed microcontroller on your board can communicate with a host PC over the same USB cable that you use for programming.
 
-<span class="tips">If you're working on Windows earlier than Windows 10, you might need to [install a serial driver](../program-setup/windows-serial-driver.html).</span>
+<span class="tips">If you're working on Windows 7, you might need to [install a serial driver](../program-setup/windows-serial-driver.html).</span>
```
iriark01, line 5: So what about Windows 8? Does that work?

iriark01 (changes requested):

docs/program-setup/serial/serial_communication.md
LDong-Arm: Yes, it works on Windows 8.

For example, https://github.com/ARMmbed/mbed-cli-windows-installer/blob/b76911430dffbb6e6e751e98a19ac4e160a7c3b4/source/install.nsi#L238-L243

thegecko: To clarify, it shouldn't be used on Windows 8 either

LDong-Arm: Sorry by "works" I meant no driver needed

PR #1338, opened 2020-06-04
Remove product ID requests

MarceloSalazar:
Note we explicitly want to avoid the need to assign a product/board ID for custom boards.
The intention is to help developers to build products that don't depend on an internal database.

Note we'll document the contribution process for Mbed Enabled board and that will include information on requests to board ID (more info IOTOSM-28).

@iriark01

### DIFF
--- a/docs/porting/custom-target-porting.md
+++ b/docs/porting/custom-target-porting.md
@@ -243,8 +243,6 @@ There are more directory levels than target configuration levels because many ta
    mbedls --mock 0764:IMAGINARYBOARD
    ```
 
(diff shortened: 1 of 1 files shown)


donatieng (approved):
LGTM, moving forward it would be good to eliminate this product ID/imaginary board story entirely. Custom hardware is based on a MCU, not a dev board and won't have onboard DAPLink, so this whole section feels like a documented hack to me :). But we're going in the right direction.

iriark01:
Marcelo, do you want this for 6.0 as well? If I merge only into `development` it'll be hidden until 6.1. I'm happy to cherry pick into 6.0

MarceloSalazar:
Thanks for reviewing this.
I'd like this fix to be delivered asap, so for 6.0 and future version as well.

iriark01:
6.0 commit: 1982b0b9a731557083cd1d73d2109e5bd0cad957

PR #1348, opened 2020-07-09
psa: Add missing FEATURE_EXPERIMENTAL_API in path

Patater:
We were missing a "FEATURE_EXPERIMENTAL_API" in a path in the PSA porting guide. Add it.


iriark01:
Will cherry pick into 6.1

PR #1350, opened 2020-07-10
Clarifying PSA/non-PSA language

iriark01:
Following a question from an engineer, clarifying the PSA/non-PSA Mbed TLS recompile portion.


Patater (approved):

iriark01:
I'll cherry pick into 6.1, just cause it's such a small change it's not a problem

PR #1354, opened 2020-08-04
Use python3 for travis job

jainvikas8:
Upgrade Travis job to use Python 3.8 instead of Python 2.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 # Language and cache settings
 language: python
-python: 2.7
+python: 3.8
 cache:
   pip: true
   directories:


Patater (approved):

iriark01:
Will this work with tests for old versions of Mbed OS? We're still supporting docs as far back as 5.12

jainvikas8:
> Will this work with tests for old versions of Mbed OS? We're still supporting docs as far back as 5.12

As long as `pip install -r mbed-os/requirements.txt` and `mbed cli` support python3 pip dependencies, I think it should be backward compatible.

iriark01:
Okay. I'm gonna merge it and see what happens.

PR #1358, opened 2020-08-10
New tools - installation

iriark01:
~Pending on Andy making a decision about the product name, but ready for engineering review.~
Mbed CLI 2


docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,53 @@
...
+<!--well this is a disaster-->
+
+Starting with version 6.x, Mbed OS is moving from Mbed CLI to a new build tool: it's cmake-based,
+
+You'll still need Mbed CLI for older versions of Mbed OS. You can install both tools side by side.<!--seems like Mbed CLI also requires Python 3, so there shouldn't be any issues-->
+
+
+# Install or upgrade
+
+Mbed Tools is Python based, so you can install it with pip.
```
Patater: "The Mbed Tools are Python based"?

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,53 @@
...
+
+<span class="tips">**Tip:** We recommend using a virtual environment such as [Pipenv](https://github.com/pypa/pipenv/blob/master/README.md) to avoid Python dependency conflicts.</span>
+
+## Prerequisite
+
+- Python 3. Install for [Windows](https://docs.python.org/3/using/windows.html), [Linux](https://docs.python.org/3/using/unix.html) or [macOS](https://docs.python.org/3/using/mac.html).
+- Pip (if not included in your Python installation). [Install for all operating systems](https://pip.pypa.io/en/stable/installing/).
+- cmake. [Install for all operating systems](https://cmake.org/install/).
+- Ninja [Install for all operating systems](https://github.com/ninja-build/ninja/wiki/Pre-built-Ninja-packages).
+<!--are there versions for cmake and ninja?-->
```
Patater: At least version 3.18.1 for CMake

v1.x for Ninja should suffice

docs/tools/tools_intro.md
```diff
@@ -19,7 +19,18 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
 
 For more information, please see the [Online Compiler page](../build-tools/mbed-online-compiler.html).
 
-## Arm Mbed CLI
+## Command line tools
+
+Mbed OS has two versions of its command line tool:
+
+- For 6.y and newer, use the new build tools.
+- For 6.x and older, use Mbed CLI.
```
Patater: It's not a hard switch. We would support Mbed CLI for 6.y and newer as well. It's only with perhaps 7.x (although not decided yet) that we'd drop Mbed CLI support. We will recommend using the new tools over the old ones in a future Mbed OS release, but not yet.

docs/tools/new_tools_temp.md
iriark01: But it's a single tool. 
Which is one of the problems I had with the name: It's a plural name but there's just one tool. 
Hence waiting on Andy.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,53 @@
...
+Mbed Tools is Python based, so you can install it with pip.
+<!--other than being a super vague name, it's a problem that it's a single tool with a plural name. "Mbed Tools is"  just parses as an error!-->
+
+<span class="tips">**Tip:** We recommend using a virtual environment such as [Pipenv](https://github.com/pypa/pipenv/blob/master/README.md) to avoid Python dependency conflicts.</span>
+
+## Prerequisite
+
+- Python 3. Install for [Windows](https://docs.python.org/3/using/windows.html), [Linux](https://docs.python.org/3/using/unix.html) or [macOS](https://docs.python.org/3/using/mac.html).
+- Pip (if not included in your Python installation). [Install for all operating systems](https://pip.pypa.io/en/stable/installing/).
+- cmake. [Install for all operating systems](https://cmake.org/install/).
```
iriark01:
```suggestion
- cmake. [Install version 3.18.1 or newer for all operating systems](https://cmake.org/install/).
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,53 @@
...
+- Ninja [Install for all operating systems](https://github.com/ninja-build/ninja/wiki/Pre-built-Ninja-packages).
```
iriark01:
```suggestion
- Ninja [Install version 1.0 or newer for all operating systems](https://github.com/ninja-build/ninja/wiki/Pre-built-Ninja-packages).
```

docs/tools/tools_intro.md
```diff
@@ -19,7 +19,18 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
 
 For more information, please see the [Online Compiler page](../build-tools/mbed-online-compiler.html).
 
-## Arm Mbed CLI
+## Command line tools
+
+Mbed OS has two versions of its command line tool:
+
+- For 6.y and newer, use the new build tools.
```
iriark01:
```suggestion
- For 6.y and newer, use the new build tools or Mbed CLI.
```

docs/tools/new_tools_temp.md
rwalton-arm: "The new command line tool is a Python package named `mbed-tools`, so you can install it with pip."?

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,53 @@
+# New tools - draft
+
+<!--well this is a disaster-->
+
+Starting with version 6.x, Mbed OS is moving from Mbed CLI to a new build tool: it's cmake-based,
```
rwalton-arm: At the moment I'm finding the way this is worded slightly confusing, we say our new "build tool" is cmake-based, then below we say it's Python based. We should make the distinction between CMake and the Python command line tools clearer.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,53 @@
...
+
+You'll still need Mbed CLI for older versions of Mbed OS. You can install both tools side by side.<!--seems like Mbed CLI also requires Python 3, so there shouldn't be any issues-->
+
+
+# Install or upgrade
+
+Mbed Tools is Python based, so you can install it with pip.
+<!--other than being a super vague name, it's a problem that it's a single tool with a plural name. "Mbed Tools is"  just parses as an error!-->
+
+<span class="tips">**Tip:** We recommend using a virtual environment such as [Pipenv](https://github.com/pypa/pipenv/blob/master/README.md) to avoid Python dependency conflicts.</span>
```
rwalton-arm: `pipenv` isn't a virtual environment, it's a virtual environment management tool.

docs/tools/new_tools_temp.md
iriark01: The whole intro is a bit of a mess (hence the "well this is a disaster"). When I get more info on the tool, it'll be easier to know what to say here. But if you can explain the distinction, that'll help.

docs/tools/new_tools_temp.md
iriark01: We're gonna wait for a proper decision about name

docs/tools/new_tools_temp.md
iriark01: Do we have a tool we want to recommend, then?

rwalton-arm: We do want to recommend `pipenv` to manage the development environment, at the moment, but we don't want to refer to `pipenv` as "a virtual environment". A virtual environment is a self-contained directory tree that contains a Python installation for a particular version of Python, plus a number of additional packages. `pipenv` is a "virtual environment management tool". We probably want to say something along the lines of: "We recommend using a virtual environment to avoid any Python dependency conflicts. We recommend the `pipenv` tool to manage your virtual development environment."

docs/tools/new_tools_temp.md
rwalton-arm: Sorry, I should've added an explanation. At a high level the distinction is:
* Ninja is our build system of choice.
* CMake is an open-source system that generates the build environment and manages the build process in a compiler independent manner.
* mbed-tools (or whatever we end up calling it) is our new command line tool that parses the build configuration from the Mbed configuration system and outputs it in a format CMake can read. mbed-tools will also provide a user friendly interface to CMake and Ninja, allowing the user to configure, generate and execute the build system with a single command.

Hopefully that helps!

iriark01: That will be lovely for the intro page (the next file in this review), and should help me figure out what people are installing here. Thank you!

iriark01:
@rwalton-arm please review again

rwalton-arm (commented):
LGTM

0Grit:
IMO the 'tools' being Mbed specific is part of the root cause of this strange issue.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+    mbed-tools init <PATH>
+    ```
+
+- To create a project without downloading a copy of Mbed OS (reuse an existing copy):
+
+    ```
+    mbed-tools init -c <PATH>
+    ```
+
+    <!--so how do I tell it where the existing copy is? In Mbed Studio you have to explicitly "tell it" where the local copy is-->
```
jainvikas8: Example: 
If we have a mbed-os-example-blinky application and we may execute it as `mbed-tools init .` inside the folder.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+
+## List connected devices
+
+To list devices connected over USB:
+
+```
+mbed-tools devices
+```
+
+<!--why? what do I do with this info? is this for the config?-->
```
jainvikas8: This is the for `mbed-tools config` or to list connected devices.  (info)

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Set environment variables
+
+<!--like what? when? why? should this be with installation instructions?-->
```
jainvikas8: The variables used by Mbed Tools can be set either directly via environment variables or
using a `.env` file containing the variable definitions as follows:

VARIABLE=value

Environment variables take precedence, meaning the values set in the file will be overridden
by any values previously set in your environment.

There's also a way to access "private" boards listed in the online DB. `mbed-tools env`, prints an explanation of how to do that using an authorisation token.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+
+To override default environment variables:
+
+```
+mbed-tools env
+```
+
+## Configure Mbed OS
+
+<!--I don't understand what this actually does. Does it generate a config file? why do I need to do this? what does "for use with" mean - do I use Mbed Tools with the target and toolchain, or am I configuring Mbed OS itself? Do I always have to do this before I build, or can I do it at the same time as the build, or can I just skip it sometimes?-->
```
jainvikas8: Does it generate a config file?  -- yes
what does "for use with" mean -- It means we use this with specific target and toolchain
am I configuring Mbed OS itself?  -- no, this is configuring build system for a particular target with a toolchain before building it.
Do I always have to do this before I build? -- No, once configured you don't need to, unless you need to change the parameters - target,, toolchain.... etc...

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+
+To prepare the Mbed configuration information for use with a specific target and toolchain:
+
+```
+mbed-tools configure -m <target> -t <toolchain>
+```
+
+Currently the supported targets are `K64F`, `DISCO_L475VG_IOT01A`, `NRF52840_DK`
+
+<!--just for configuring, or can you only build with those?-->
```
jainvikas8: Config and build.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Build using CMake
+
+After preparing the Mbed configuration information, you can use CMake to build your application:<!--can, but don't have to? what are the other options?-->
+
+```console
+cmake -S . -B cmake_build -GNinja -DCMAKE_BUILD_TYPE=<profile>
+cmake --build cmake_build
+```
+<!--what are all the other parameters?-->
```
jainvikas8: These are Cmake options
```   
Options
       -S <path-to-source>
              Path to root directory of the CMake project to build.
       -B <path-to-build>
              Path to directory which CMake will use as the root of build
              directory.
              If the directory doesn't already exist CMake will make it.
```

-GNinja -- we are using Ninja tool
-DCMAKE_BUILD_TYPE -- Build type

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+<!--what does the first command do, and what does the second one do?-->
```
jainvikas8: `cmake -S . -B cmake_build -GNinja -DCMAKE_BUILD_TYPE=<profile>`
This one sets the build parameters

the second does the actual build.

docs/tools/new_tools_temp.md
rwalton-arm: The tool doesn't support "telling it where an existing copy is" yet, it (meaning an existing copy of mbed-os) would have to be symlinked into the program folder. We are going to support this at some point when we implement "workspace" support, but it's not done yet.

docs/tools/new_tools_temp.md
rwalton-arm: It parses the mbed config system (mbed_lib.json, mbed_app.json and targets.json) for a particular target and toolchain, and outputs a cmake script that is used to configure the build system. After the cmake script has been generated you can reuse it if you're building for the same target and toolchain. If you decide to change target and/or toolchain, you have to run the configure command again. Note: we will soon be implementing a build command to simplify the config and build steps so the user doesn't have to run multiple commands. The build command isn't implemented yet.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+mbed-tools configure -m <target> -t <toolchain>
+```
+
+Currently the supported targets are `K64F`, `DISCO_L475VG_IOT01A`, `NRF52840_DK`
+
+<!--just for configuring, or can you only build with those?-->
+
+## Build using CMake
+
+After preparing the Mbed configuration information, you can use CMake to build your application:<!--can, but don't have to? what are the other options?-->
```
rwalton-arm: There are no other options, you *must* use CMake to generate the build environment and invoke the build system (which is Ninja in our case); the configuration script the tool generates with the `configure` command is CMake specific.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+
+```console
+cmake -S . -B cmake_build -GNinja -DCMAKE_BUILD_TYPE=<profile>
+cmake --build cmake_build
+```
+<!--what are all the other parameters?-->
+<!--what does the first command do, and what does the second one do?-->
+- We recommend using a dedicated **cmake_build** folder to keep things organised.<!--so do I replace cmake_build with a new value?-->
+- The `profile` value can be `release`, `debug` or `develop`.<!--same as normal, or something specific to CMake? https://os.mbed.com/docs/mbed-os/v6.2/program-setup/build-profiles-and-rules.html-->
```
rwalton-arm: Same as the old profiles.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+
+<!--what does this generate and where? is this where we add instructions for flashing the board?-->
```
rwalton-arm: It generates the binary and hex files in the `cmake_build` folder. "Flashing" (i.e invoking `cp` to copy the binary to the USB disk presented by DAPlink) isn't implemented in the new tools yet, but it's on our backlog.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Example applications
+
+<!--can I create a new one or do I have to use an example? I'm asking because these were listed as part of the build instructions file, I think-->
```
rwalton-arm: You can create a new project and build it, that's what the `init` command does. This is just a list of the examples that currently support CMake.

docs/tools/tools_intro.md
```diff
@@ -19,7 +19,25 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
...
+- For 6.y and newer, use the new build tools or Mbed CLI.
+- For 6.x and older, use Mbed CLI.
+
+You can have both tools installed side by side, so you can work with multiple versions of Mbed OS.
+
+### Mbed Tools - Mbed OS 6.y and newer
+
+Starting with version 6.x, Mbed OS is moving from Mbed CLI to a new build tool: <!--that has no proper name yet-->. It uses Ninja as a build system, and CMake to generate the build environment and manage the build process in a compiler-independent manner. <!--need better phrasing-->
+
+Mbed Tools parses the Mbed OS build configuration and outputs it to a format CMake can read. It also provides a use friendly interface to CMake and Ninja so you can configure, generte and execute builds with a single command.
```
rwalton-arm:
```suggestion
Mbed Tools parses the Mbed OS build configuration and outputs it to a format CMake can read. It also provides a user friendly interface to CMake and Ninja so you can configure, generate and execute builds with a single command.
```

docs/tools/new_tools_temp.md
iriark01: How actually do I get an example going? How do I tell it to go download a copy of mbed-os-example-blinky, and do I need to use `init` in an example?

iriark01: How do we use SymLink, then? If it's not been done before in Mbed OS, we need to explain it.

docs/tools/new_tools_temp.md
iriark01: I think this needs to be in the setup stage; you wouldn't expect people to constantly set environment variables, would you? It certainly doesn't strike me as a step in every single project.

docs/tools/new_tools_temp.md
iriark01:
```suggestion
The Mbed OS configuration system parses the configuration files in your project (mbed_lib.json, mbed_app.json and targets.json) for a particular target and toolchain, and outputs a CMake script. The build system uses this script to build for your target, using your toolchain. 

If you're rebuilding for the same target and toolchain, you can keep using the same CMake script, so you won't have to use the `configure` command again for each build. If you change your target or toolchain, run the `configure` command again to generate a new CMake script.```

File: docs/tools/new_tools_temp.md [hunk 21]
```diff
@@ -0,0 +1,144 @@
...
+
+```
+mbed-tools env
+```
+
+## Configure Mbed OS
+
+<!--I don't understand what this actually does. Does it generate a config file? why do I need to do this? what does "for use with" mean - do I use Mbed Tools with the target and toolchain, or am I configuring Mbed OS itself? Do I always have to do this before I build, or can I do it at the same time as the build, or can I just skip it sometimes?-->
+
+To prepare the Mbed configuration information for use with a specific target and toolchain:
```
iriark01:
```suggestion
To generate a script for a new target and toolchain combination:
```

docs/tools/new_tools_temp.md
iriark01: Okay, we need to mention this much earlier. I'll add a line.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Upgrade
+
+Use pip to upgrade:
+
+```
+python -m pip install mbed-tools --upgrade
+```
+
+# Use
```
iriark01:
```suggestion
# Use

Currently, the new Mbed Tools supports three boards: `K64F`, `DISCO_L475VG_IOT01A` and `NRF52840_DK`.

```

docs/tools/new_tools_temp.md
iriark01: What does the output look like? Does it give me the target name so I can use it for my config command?

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+## Configure Mbed OS
+
+<!--I don't understand what this actually does. Does it generate a config file? why do I need to do this? what does "for use with" mean - do I use Mbed Tools with the target and toolchain, or am I configuring Mbed OS itself? Do I always have to do this before I build, or can I do it at the same time as the build, or can I just skip it sometimes?-->
+
+To prepare the Mbed configuration information for use with a specific target and toolchain:
+
+```
+mbed-tools configure -m <target> -t <toolchain>
```
iriark01: What are the possible toolchain values? Same as for Mbed CLI?

docs/tools/new_tools_temp.md
iriark01:
```suggestion
After preparing the Mbed configuration information, use CMake to build your application:
```

docs/tools/new_tools_temp.md
iriark01: Okay. I will show these as two steps, then, with an explanation.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,144 @@
...
+
+After preparing the Mbed configuration information, you can use CMake to build your application:<!--can, but don't have to? what are the other options?-->
+
+```console
+cmake -S . -B cmake_build -GNinja -DCMAKE_BUILD_TYPE=<profile>
+cmake --build cmake_build
+```
+<!--what are all the other parameters?-->
+<!--what does the first command do, and what does the second one do?-->
+- We recommend using a dedicated **cmake_build** folder to keep things organised.<!--so do I replace cmake_build with a new value?-->
```
iriark01: This question was missed: do I replace `cmake_build` with a new value?

docs/tools/new_tools_temp.md
iriark01: So flashing is drag&drop?

docs/tools/new_tools_temp.md
iriark01: How do I actually get them? What's the new `mbed import`?

docs/tools/new_tools_temp.md
jainvikas8: Yes, target name is given.

docs/tools/new_tools_temp.md
jainvikas8: yes, it doesn't need to be set always...

docs/tools/new_tools_temp.md
rwalton-arm: At the moment flashing is not supported by the tool and is left to the user. They can use drag&drop.

docs/tools/new_tools_temp.md
rwalton-arm: We have a `clone` command which is the new `mbed import`. I don't think we have documented that command yet.

docs/tools/new_tools_temp.md
jainvikas8: script = Application configuration ?

docs/tools/new_tools_temp.md
jainvikas8: Yes, GCC_ARM and ARM

docs/tools/new_tools_temp.md
rwalton-arm: Sorry, not sure I understand the question. We create the `cmake_build` folder with the `cmake -B cmake_build` command. Why would we need to replace it with a new value? I assume this recommendation is here so people don't do an "in source" build and pollute the source tree with build system artifacts.

jainvikas8: `cmake_build` is the folder. It is recommended to use a folder for the build, as this will keep things organised. Normally a default name of that folder is `cmake_build`. 
We mentioned in the command `cmake --build cmake_build` because the user can copy and paste to execute it.

docs/tools/new_tools_temp.md
rwalton-arm: `init` is just for creating a new project, it's similar to `mbed new`. We also have a `clone` command for downloading examples, which is similar to `mbed import` but we haven't documented that command, yet. I think we should just say "copy" rather than symbolic linking as I'm not sure we've tested symlinked CMake builds work yet. The `-c` workflow with the `init` command won't be supported until we have workspace support in the tools, which isn't implemented yet.

docs/tools/new_tools_temp.md
jainvikas8: 'these' boards would be good, rather than being specific?

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+    mbed-tools init <PATH>
+    ```
+
+- To create a project without downloading a copy of Mbed OS (reuse an existing copy):
+
+    ```
+    mbed-tools init -c <PATH>
+    ```
+
+<!--this seems to be an absolute path - had to start with `~/`. Can they work with relative paths? can they work with "here" to tell mbed-tools to work in the current directory?-->
```
jainvikas8: `mbed-tools init -c .` can be used and also with relative paths like `mbed-tools init -c .\mbed-os-example-blinky`

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+
+1. Build:
+
+    ```
+    cmake --build cmake_build
+    ```
+
+    This generates two files: BIN and HEX in the build output directory (`cmake_build` in this example).
+
+1. Drag and drop the generated file to your board. <!--which file? is it up to them?-->
```
jainvikas8: Mostly it's in the format of **application.hex** for Example:  `mbed-os-example-blinky.hex`

docs/tools/new_tools_temp.md
rwalton-arm: It works with relative paths... You have to at least specify the directory name to create as it uses that as the project name. `mbedtools init -c test` works.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+- To create a project without downloading a copy of Mbed OS (reuse an existing copy):
+
+    ```
+    mbed-tools init -c <PATH>
+    ```
+
+<!--this seems to be an absolute path - had to start with `~/`. Can they work with relative paths? can they work with "here" to tell mbed-tools to work in the current directory?-->
+
+
+<!--If I can't symlink - if I have to copy my local version of Mbed OS - then this is less about savings space and more about allowing me to take a non-default version of Mbed OS, then?-->
```
rwalton-arm: It will be about saving space, but we haven't implemented workspace support yet..

docs/tools/new_tools_temp.md
```diff
@@ -46,23 +46,49 @@ Use pip to install:
...
     ```
 
+## Set environment variables
+
+<!--like what? when? why? should this be with installation instructions?-->
+
+To override default environment variables:
+
+```
+mbed-tools env <VARIABLE>=<value>
```
rwalton-arm: What's this? `mbedtools env` just prints information about the environment variables you can set if you want to access private boards in the online database, it doesn't allow you to set variables like that...

docs/tools/new_tools_temp.md
```diff
@@ -46,23 +46,49 @@ Use pip to install:
     python -m pip install mbed-tools --pre
     ```
 
+## Set environment variables
+
+<!--like what? when? why? should this be with installation instructions?-->
```
rwalton-arm: If you run `mbedtools env` it will print a detailed description to the console...

```
$ mbedtools env
These variables used by Mbed Tools can be set either directly via environment variables or
using a `.env` file containing the variable definitions as follows:

VARIABLE=value

Environment variables take precedence, meaning the values set in the file will be overriden
by any values previously set in your environment.

+---------------------------------------------------------------------------------+
| Do not upload `.env` files containing private tokens to version control! If you |
| use this package as a dependency of your project, please ensure to include the  |
| `.env` in your `.gitignore`.                                                    |
+---------------------------------------------------------------------------------+


MBED_API_AUTH_TOKEN

        Token to use when accessing online API.

        Mbed Targets uses the online mbed board database at os.mbed.com as its data source.
        A snapshot of the board database is shipped with the package, for faster lookup of known
        boards. Only public boards are stored in the database snapshot. If you are fetching data
        for a private board, mbed-targets will need to contact the online database.

        To fetch data about private boards from the online database, the user must have an account
        on os.mbed.com and be member of a vendor team that has permissions to see the private board.
        An authentication token for the team member must be provided in an environment variable named
        `MBED_API_AUTH_TOKEN`.

MBED_DATABASE_MODE

        Database mode to use when retrieving board data.

        Mbed Targets supports an online and offline mode, which controls where to look up the board database.

        The board lookup can be from either the online or offline database, depending
        on the value of an environment variable called `MBED_DATABASE_MODE`.

        The mode can be set to one of the following:

        - `AUTO`: the offline database is searched first, if the board isn't found the online database is searched.
        - `ONLINE`: the online database is always used.
        - `OFFLINE`: the offline database is always used.

        If `MBED_DATABASE_MODE` is not set, it defaults to `AUTO`.
```

docs/tools/new_tools_temp.md
```diff
@@ -78,67 +104,83 @@ To create a new Mbed OS project in a specified path:
...
+<!--this seems to be an absolute path - had to start with `~/`. Can they work with relative paths? can they work with "here" to tell mbed-tools to work in the current directory?-->
 
-## List connected devices
 
-To list devices connected over USB:
+<!--If I can't symlink - if I have to copy my local version of Mbed OS - then this is less about savings space and more about allowing me to take a non-default version of Mbed OS, then?-->
+
+### Use an example application
+
+To create a local copy of an example application, use the `clone` command with the full GitHub URL listed below:
```
rwalton-arm: You don't need to give the full github URL, you can use `mbedtools clone <example-name>` for ARMmbed examples.

docs/tools/new_tools_temp.md
```diff
@@ -46,23 +46,49 @@ Use pip to install:
     python -m pip install mbed-tools --pre
```
rwalton-arm: Do we mention anywhere that `mbed-tools` is only compatible with python 3.6 and above?

docs/tools/new_tools_temp.md
iriark01: So. How and why do you actually use this idea? Because the impression I got from the original text is that env is how you modify things, not list them.

docs/tools/new_tools_temp.md
iriark01: What does "mostly" mean? Does it depend on the board type?

jainvikas8: Sorry for the confusion, but it is in **application.hex** format.

docs/tools/new_tools_temp.md
rwalton-arm: You can either specify the variables in a `.env` file and the tool will discover them, or you can set them per terminal session the same way you would set any environment variable https://www.schrodinger.com/kb/1842. The "why" is what `mbedtools env` tells you, as pasted in the comment below. `mbedtools env` doesn't list any set variables, it only describes the variables the tool recognises and why you might need to set them (most users will not need to set them at all).

docs/tools/new_tools_temp.md
iriark01: We say Python 3, not 3.6. Line 21

rwalton-arm: OK. We should specify Python 3.6 or later, as it won't work correctly with Python 3.5 or below. Annoyingly Python 3 has many changes to standard library APIs between 3.0 and 3.6 that aren't entirely backwards compatible.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+### Use an example application
+
+To create a local copy of an example application, use the `clone` command with the full GitHub URL listed below:
+
+```
+mbed-tools clone <link to GitHub example> <PATH>
+````
+
+- Blinky: [https://github.com/ARMmbed/mbed-os-example-blinky](https://github.com/ARMmbed/mbed-os-example-blinky)
+- BLE button: [https://github.com/ARMmbed/mbed-os-example-ble/tree/master/BLE_Button](https://github.com/ARMmbed/mbed-os-example-ble/tree/master/BLE_Button)
```
jainvikas8: This needs to be changed to 
```
- BLE button: [mbed-os-example-ble](https://github.com/ARMmbed/mbed-os-example-ble)
```
Cloning with `https://github.com/ARMmbed/mbed-os-example-ble/tree/master/BLE_Button` will not work.

Also only BLE button example is only supported.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+# Install or upgrade
+
+Mbed Tools is a Python package called `mbed-tools`, so you can install it with pip.
+<!--other than being a super vague name, it's a problem that it's a single tool with a plural name. "Mbed Tools is"  just parses as an error!-->
+
+<span class="tips">**Tip:** We recommend using a virtual environment to avoid any Python dependency conflicts. We recommend the [Pipenv](https://github.com/pypa/pipenv/blob/master/README.md) tool to manage your virtual development environment.
+
+## Prerequisite
+
+- Python 3. Install for [Windows](https://docs.python.org/3/using/windows.html), [Linux](https://docs.python.org/3/using/unix.html) or [macOS](https://docs.python.org/3/using/mac.html).
```
iriark01:
```suggestion
- Python 3.6 or newer. Install for [Windows](https://docs.python.org/3/using/windows.html), [Linux](https://docs.python.org/3/using/unix.html) or [macOS](https://docs.python.org/3/using/mac.html).
```

docs/tools/new_tools_temp.md
iriark01: So the command is `setenv`, rather than `env`? (that's according to the KB you linked to)

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+
+- To install a pre-release or development version:
+
+    ```
+    python -m pip install mbed-tools --pre
+    ```
+
+## Set environment variables
+
+<!--like what? when? why? should this be with installation instructions?-->
```
iriark01: Mbed Tools has two environment variables: 

- `MBED_API_AUTH_TOKEN`: Token to access private board information stored for a vendor team. 
-  `MBED_DATABASE_MODE`: Use online or offline mode. Possible values:
        - `AUTO`: Search the offline database first; search the online database only if the board wasn't found offline. This is the default value.
        - `ONLINE`: Alway use the online database.
        - `OFFLINE`: Always use the offline database.

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+
+To override default environment variables:
+
+```
+mbed-tools env <VARIABLE>=<value>
```
iriark01:
```suggestion
To list the current values:

`mbed-tools env `

To set values, create an `.env` file in the root directory of the project. The file should contain definitions in the `<VARIABLE>=<value>` format. 
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+You can also use a `.env` file containing definitions in the same `<VARIABLE>=<value>` format. Note that environment variables set using the command line will override those listed in the file.
```
iriark01:
```suggestion
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+    ```
+
+<!--this seems to be an absolute path - had to start with `~/`. Can they work with relative paths? can they work with "here" to tell mbed-tools to work in the current directory?-->
+
+
+<!--If I can't symlink - if I have to copy my local version of Mbed OS - then this is less about savings space and more about allowing me to take a non-default version of Mbed OS, then?-->
+
+### Use an example application
+
+To create a local copy of an example application, use the `clone` command with the full GitHub URL listed below:
```
iriark01:
```suggestion
To create a local copy of an example application, use the `clone` command with the example name listed below:
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+
+```
+mbed-tools clone <link to GitHub example> <PATH>
```
iriark01:
```suggestion
mbed-tools clone <example> <PATH>
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+````
+
+- Blinky: [https://github.com/ARMmbed/mbed-os-example-blinky](https://github.com/ARMmbed/mbed-os-example-blinky)
```
iriark01:
```suggestion
- [mbed-os-example-blinky](https://github.com/ARMmbed/mbed-os-example-blinky)
```

docs/tools/new_tools_temp.md
iriark01:
```suggestion
- [mbed-os-example-ble](https://github.com/ARMmbed/mbed-os-example-ble) - use the BLE button example.
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+- BLE button: [https://github.com/ARMmbed/mbed-os-example-ble/tree/master/BLE_Button](https://github.com/ARMmbed/mbed-os-example-ble/tree/master/BLE_Button)
+- Cellular: [https://github.com/ARMmbed/mbed-os-example-cellular](https://github.com/ARMmbed/mbed-os-example-cellular)
```
iriark01:
```suggestion
- [mbed-os-example-cellular](https://github.com/ARMmbed/mbed-os-example-cellular)
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+- DeviceKey: [https://github.com/ARMmbed/mbed-os-example-devicekey](https://github.com/ARMmbed/mbed-os-example-devicekey)
```
iriark01:
```suggestion
- [mbed-os-example-devicekey](https://github.com/ARMmbed/mbed-os-example-devicekey)
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+- KVStore: [https://github.com/ARMmbed/mbed-os-example-kvstore](https://github.com/ARMmbed/mbed-os-example-kvstore)
```
iriark01:
```suggestion
- [mbed-os-example-kvstore](https://github.com/ARMmbed/mbed-os-example-kvstore)
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+- LoRaWAN: [https://github.com/ARMmbed/mbed-os-example-lorawan](https://github.com/ARMmbed/mbed-os-example-lorawan)
```
iriark01:
```suggestion
- [mbed-os-example-lorawan](https://github.com/ARMmbed/mbed-os-example-lorawan)
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+- Mbed Crypto: [https://github.com/ARMmbed/mbed-os-example-mbed-crypto](https://github.com/ARMmbed/mbed-os-example-mbed-crypto)
```
iriark01:
```suggestion
- [mbed-os-example-mbed-crypto](https://github.com/ARMmbed/mbed-os-example-mbed-crypto)
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+- NFC: [https://github.com/ARMmbed/mbed-os-example-nfc](https://github.com/ARMmbed/mbed-os-example-nfc)
```
iriark01:
```suggestion
- [mbed-os-example-nfc](https://github.com/ARMmbed/mbed-os-example-nfc)
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+- Sockets: [https://github.com/ARMmbed/mbed-os-example-sockets](https://github.com/ARMmbed/mbed-os-example-sockets)
```
iriark01:
```suggestion
- [mbed-os-example-sockets](https://github.com/ARMmbed/mbed-os-example-sockets)
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+    - -GNinja: To use the Ninja tool.
+    - -DCMAKE_BUILD_TYPE: Build type. The value (`profile`) can be `release`, `debug` or `develop`, as [explained in program setup](../program-setup/build-profiles-and-rules.html).
+
+1. Build:
+
+    ```
+    cmake --build cmake_build
+    ```
+
+    This generates two files: BIN and HEX in the build output directory (`cmake_build` in this example).
```
iriark01:
```suggestion
    This generates two files in the build output directory (`cmake_build` in this example): HEX and BIN. <!--find a place where we explain where HEX is useful and link to it-->.
```

docs/tools/new_tools_temp.md
iriark01:
```suggestion
1. Drag and drop the generated file to your board. 
```

docs/tools/new_tools_temp.md
iriark01: Might be best to remove these instructions, then.

docs/tools/new_tools_temp.md
jainvikas8: It generates both `HEX` and `bin` files, which can be used.

docs/tools/new_tools_temp.md
jainvikas8: The description is printed when `mbed-tools env` is used... I don't see why this information is required to be repeated.

docs/tools/new_tools_temp.md
iriark01: But when I asked how you know which one to use, the answer I got was "it's HEX".

jainvikas8: The build will generate both hex and bin. It’s up to the developer to use either of them. But the recommendation is the `HEX` file.

iriark01: What's the advantage of the HEX file?

jainvikas8: The advantage with hex is specifying the address for the image. During flashing only the address range mentioned in the `hex` are modified.

iriark01: That actually sounds like a pretty big deal that we should maybe consider focusing on. Maybe not right in the middle of the build process, though. Maybe the build tools intro? Do they all do this?

docs/tools/new_tools_temp.md
iriark01: I'm not sure you and Vikas gave exactly the same instructions. So the options are:
relative:
`.` for "here"
`.\folder` for "new folder here"
or absolute

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,186 @@
...
+
+    ```
+    python -m pip install mbed-tools --pre
+    ```
+
+## Set environment variables
+
+<!--like what? when? why? should this be with installation instructions?-->
+
+To override default environment variables:
```
iriark01:
```suggestion
Mbed Tools has two environment variables:

- `MBED_API_AUTH_TOKEN`: Token to access private board information stored for a vendor team.
- `MBED_DATABASE_MODE`: Use online or offline mode. Possible values:
    - AUTO: Search the offline database first; search the online database only if the board wasn't found offline. This is the default value.
    - ONLINE: Alway use the online database.
    - OFFLINE: Always use the offline database.```

File: docs/tools/new_tools_temp.md [hunk 49]
```diff
@@ -0,0 +1,189 @@
...
+    ```
+
+    The path can be:
+    - Absolute. `init` will create the folder if it doesn't exist.
+    - Relative.
+
+    If you have already created a project folder, you can use `.`
+
+    If you want the `init` command to create a project folder, use `.\<folder-name>`.
+
```
urutva: @iriark01 We should also document `-c` option which allows sym-linking existing mbd-os. Please check https://github.com/ARMmbed/mbed-tools/issues/65

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,189 @@
...
+To set values, create an `.env` file in the root directory of the project. The file should contain definitions in the `<VARIABLE>=<value>` format.
+
+For more information, use `mbed-tools env --help`
+
+### Mbed OS configuration
+
+The Mbed OS configuration system parses the configuration files in your project (mbed_lib.json, mbed_app.json and targets.json) for a particular target and toolchain, and outputs a CMake script. The build system uses this script to build for your target, using your toolchain.
+
+<span class="tips">**Tip:** If you're rebuilding for the same target and toolchain, you can keep using the same CMake script, so you won't have to use the `configure` command again for each build. If you change your target or toolchain, run the `configure` command again to generate a new CMake script.</span>
```
urutva:
```suggestion
<span class="tips">**Tip:** If you're rebuilding for the same target and toolchain, you can keep using the same CMake script, so you won't have to use the `configure` command again for each build. If you change any of mbed_lib.json, mbed_app.json, targets.json, target or toolchain, run the `configure` command again to generate a new CMake script.</span>
```

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,189 @@
...
+
+- `MBED_API_AUTH_TOKEN`: Token to access private board information stored for a vendor team.
+- `MBED_DATABASE_MODE`: Use online or offline mode. Possible values:
+    - `AUTO`: Search the offline database first; search the online database only if the board wasn't found offline. This is the default value.
+    - `ONLINE`: Alway use the online database.
+    - `OFFLINE`: Always use the offline database.
+
+To set values, create an `.env` file in the root directory of the project. The file should contain definitions in the `<VARIABLE>=<value>` format.
+
+For more information, use `mbed-tools env --help`
```
jainvikas8: `mbed-tools env`

docs/tools/new_tools_temp.md
```diff
@@ -0,0 +1,189 @@
...
+## Upgrade
+
+Use pip to upgrade your version:
+
+```
+python -m pip install mbed-tools --upgrade
+```
+
+# Use
+
```
urutva: @iriark01 We should document the verbose logging somewhere. I'm not sure about the location. I think `Use` chapter might be a good fit.
Example:
`mbed-tools -v configure -m <target> -t <toolchain>`
TIP: repeat character `v` to increase logging level. For example `-vv` or `-vvv`.

iriark01: How many Vs can you use?

docs/tools/new_tools_temp.md
iriark01: I removed that because it wasn't fully functional

docs/tools/new_tools_temp.md
iriark01: Have we figured out what does what? I found the Slack conversation really confusing

docs/tools/new_tools_temp.md
urutva: It can be any number. Following might help:
```
None -> Show only errros
-v -> Show warnings + errors
-vv -> Show information + warnings + errors
-vvv -> Show debug + information + warnings + errors
```
Repeating more than thrice will have the same effect as repeating thrice. But `mbed-tools` won't throw any error for repeating more then thrice.

docs/tools/new_tools_temp.md
```diff
@@ -187,3 +190,18 @@ Use CMake to build your application:
     This generates two files in the build output directory (`cmake_build` in this example): HEX and BIN. <!--find a place where we explain where HEX is useful and link to it-->.
 
 1. Drag and drop the generated file to your board.
+
+## Logging
+
+To specify the log level, use the verbose logging option (`-v`) before the first argument:
```
jainvikas8, line 196:
```suggestion
To specify the log level, use the verbose logging option (`-v`) before the first argument.

If you don't use `-v`, by default the log will show only errors. These are the following log levels:
```

jainvikas8:
```suggestion
To specify the log level, use the verbose logging option (`-v`) before the first argument.

If you don't use `-v`, by default the log will show only errors. These are the following log levels:
```

docs/tools/new_tools_temp.md
urutva: @iriark01 Did you face any errors when you ran`mbed-tools init -c`?

iriark01: I was simply told it doesn't work yet

jainvikas8: This has been added recently, please review and resolve this comment @urutva

urutva: @iriark01 @jainvikas8 I've tried `mbed-tools init -c` after commenting and it works as expected. It should be added to the document.

iriark01: It's in line 42 in the file about using it

docs/tools/mbed_cli_2/install.md
```diff
@@ -0,0 +1,53 @@
...
+
+- Python 3.6 or newer. Install for [Windows](https://docs.python.org/3/using/windows.html), [Linux](https://docs.python.org/3/using/unix.html) or [macOS](https://docs.python.org/3/using/mac.html).
+- Pip (if not included in your Python installation). [Install for all operating systems](https://pip.pypa.io/en/stable/installing/).
+- CMake. [Install version 3.18.1 or newer for all operating systems](https://cmake.org/install/).
+- Ninja. [Install version 1.0 or newer for all operating systems](https://github.com/ninja-build/ninja/wiki/Pre-built-Ninja-packages).
+- One of the support toolchains listed [in the build tools overview](../build-tools/index.html).
+
+## Install
+
+Some header files and static libraries are required by Pip and is installed by python development package.
```
iriark01:
```suggestion
Pip requires some header files and static libraries. Install the Python development package to get them:
```

docs/tools/mbed_cli_2/install.md
```diff
@@ -0,0 +1,53 @@
...
+    python -m pip install mbed-tools==<version number in major.minor.patch format>
+    ```
+
+- To install a pre-release or development version:
+
+    ```
+    python -m pip install mbed-tools --pre
+    ```
+
+<span class="tips">**Tip:** On a Unix-like system, you may need to ensure that `~/.local/bin` is accessible and in your path. </span>
```
iriark01:
```suggestion
<span class="tips">**Tip:** On a Unix-like system, ensure that `~/.local/bin` is accessible and in your path. </span>
```

Patater:
`feature-cmake` has been merged to mbed-os. We should merge this PR so that the new feature has docs.

jainvikas8:
Forced pushed to resolve conflicts and rebase

docs/tools/mbed_cli_2/intro.md
```diff
@@ -0,0 +1,7 @@
+# Mbed CLI 2
+
+Starting with version 6.x, Mbed OS is moving to Mbed CLI 2. It uses Ninja as a build system, and CMake to generate the build environment and manage the build process in a compiler-independent manner.
```
Patater: I believe the x should be 5. The first Mbed OS release with CMake support is Mbed OS 6.5.

```diff
- 6.x
+ 6.5
```

docs/tools/mbed_cli_2/intro.md
```diff
@@ -0,0 +1,7 @@
...
+
+Mbed CLI 2 parses the Mbed OS build configuration and outputs it to a format CMake can read. It also provides a user friendly interface to CMake and Ninja so you can configure, generate and execute builds with a single command.
+
+<span class="notes">**Note:** You'll still need Mbed CLI 1 for older versions of Mbed OS (6.x and older). You can install both tools side by side.</span>
```
Patater:
```diff
- 6.x
+ 6.4
```

docs/tools/tools_intro.md
```diff
@@ -19,9 +19,28 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
...
 For more information, please see the [Online Compiler page](../build-tools/mbed-online-compiler.html).
 
-## Arm Mbed CLI
+## Command line tools
 
-We created the Mbed command-line tool (Mbed CLI), a Python-based tool, specifically for Mbed OS. For more information, see the [Mbed CLI page](../tools/developing-mbed-cli.html).
+Mbed OS has two versions of its command line tool:
+
+- For 6.y and newer, use Mbed CLI 1 or Mbed CLI 2.
+- For 6.x and older, use Mbed CLI 1.
```
Patater:
```diff
- 6.x
+ 6.4
```

docs/tools/tools_intro.md
```diff
@@ -19,9 +19,28 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
 
 For more information, please see the [Online Compiler page](../build-tools/mbed-online-compiler.html).
 
-## Arm Mbed CLI
+## Command line tools
 
-We created the Mbed command-line tool (Mbed CLI), a Python-based tool, specifically for Mbed OS. For more information, see the [Mbed CLI page](../tools/developing-mbed-cli.html).
+Mbed OS has two versions of its command line tool:
+
+- For 6.y and newer, use Mbed CLI 1 or Mbed CLI 2.
```
Patater:
```diff
- 6.y
+ 6.5
```

docs/tools/tools_intro.md
```diff
@@ -19,9 +19,28 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
...
+
+### Mbed CLI 2 - Mbed OS 6.y and newer
+
+Starting with version 6.x, Mbed OS is moving to Mbed CLI 2. It uses Ninja as a build system, and CMake to generate the build environment and manage the build process in a compiler-independent manner.
+
+Mbed CLI 2 parses the Mbed OS build configuration and outputs it to a format CMake can read. It also provides a user friendly interface to CMake and Ninja so you can configure, generate and execute builds with a single command.
+
+For more information, see the [Mbed CLI 2 page](../build-tools/mbed-cli-2.html).
+
+<span class="notes">**Note:** You'll still need Mbed CLI 1 for older versions of Mbed OS (6.x and older). You can install both tools side by side.</span>
```
Patater:
```diff
- 6.x
+ 6.4
```

docs/tools/tools_intro.md
```diff
@@ -19,9 +19,28 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
...
+
+### Mbed CLI 1 - Mbed OS 6.x and older
```
Patater:
```diff
- 6.x
+ 6.4
```

docs/tools/tools_intro.md
```diff
@@ -19,9 +19,28 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
...
 
-We created the Mbed command-line tool (Mbed CLI), a Python-based tool, specifically for Mbed OS. For more information, see the [Mbed CLI page](../tools/developing-mbed-cli.html).
+Mbed OS has two versions of its command line tool:
+
+- For 6.y and newer, use Mbed CLI 1 or Mbed CLI 2.
+- For 6.x and older, use Mbed CLI 1.
+
+You can have both tools installed side by side, so you can work with multiple versions of Mbed OS.
+
+### Mbed CLI 2 - Mbed OS 6.y and newer
```
Patater:
```diff
- 6.y
+ 6.5
```

docs/tools/tools_intro.md
```diff
@@ -19,9 +19,28 @@ The Mbed Online Compiler is our in-house IDE and should be familiar to anyone wh
...
+
+Starting with version 6.x, Mbed OS is moving to Mbed CLI 2. It uses Ninja as a build system, and CMake to generate the build environment and manage the build process in a compiler-independent manner.
```
Patater:
```diff
- 6.x
+ 6.5
```

Patater (changes requested):

docs/tools/tools_intro.md
```diff
@@ -23,22 +23,22 @@ For more information, please see the [Online Compiler page](../build-tools/mbed-
...
+- For 6.5 and newer, use Mbed CLI 1 or Mbed CLI 2.
+- For 6.4 and older, use Mbed CLI 1.
 
 You can have both tools installed side by side, so you can work with multiple versions of Mbed OS.
 
-### Mbed CLI 2 - Mbed OS 6.y and newer
+### Mbed CLI 2 - Mbed OS 6.5 and newer
 
-Starting with version 6.x, Mbed OS is moving to Mbed CLI 2. It uses Ninja as a build system, and CMake to generate the build environment and manage the build process in a compiler-independent manner.
+Starting with version 6.4, Mbed OS is moving to Mbed CLI 2. It uses Ninja as a build system, and CMake to generate the build environment and manage the build process in a compiler-independent manner.
```
Patater:
```diff
- 6.4
+ 6.5
```

Patater (changes requested):

Patater (approved):

PR #1362, opened 2020-08-18
porting: Rename MBED_BOOT_STACK_SIZE to MBED_CONF_TARGET_BOOT_STACK_SIZE

rwalton-arm:
MBED_BOOT_STACK_SIZE is a duplicated definition of
MBED_CONF_TARGET_BOOT_STACK_SIZE, which we are removing as it is
causing issues in mbed-os.

This commit updates the porting guide to only mention
MBED_CONF_TARGET_BOOT_STACK_SIZE when referring to the define for the
boot stack size.

### DIFF
--- a/docs/porting/target/bootstrap.md
+++ b/docs/porting/target/bootstrap.md
@@ -28,7 +28,7 @@ If you are updating your own linker script, you must:
     - Arm - The boot stack is the `ARM_LIB_STACK` region.
     - GCC_ARM - The boot stack starts at the symbol `__StackLimit` and ends at the symbol `__StackTop`.
 - Add defines for a relocatable application - `MBED_APP_START` and `MBED_APP_SIZE`.
-- Add the define for boot stack size - `MBED_BOOT_STACK_SIZE`.
+- Add the define for boot stack size - `MBED_CONF_TARGET_BOOT_STACK_SIZE`.
 - Add preprocessing directive `#! armcc -E` (ARM compiler only).
 
 If you are using the below linker script, then you need to update all the defines in the `/* Device specific values */` section for your target.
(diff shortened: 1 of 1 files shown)


Patater (approved):
LGTM

iriark01:
Which versions of Mbed OS does this apply to?

rwalton-arm:
> Which versions of Mbed OS does this apply to?

We haven't merged the PR to change this yet, so it won't be until at least 6.2.2.

docs/porting/target/bootstrap.md
```diff
@@ -54,9 +54,9 @@ If you are using the below linker script, then you need to update all the define
   #define MBED_APP_SIZE  MBED_ROM_SIZE
 #endif
 
-#if !defined(MBED_BOOT_STACK_SIZE)
+#if !defined(MBED_CONF_TARGET_BOOT_STACK_SIZE)
```
jeromecoutant, line 57: Is this really useful?
In Mbed environment, this macro has to be defined... or we have some other issue ?
Which use case do you protect ?
Thx

docs/porting/target/bootstrap.md
```diff
@@ -28,7 +28,7 @@ If you are updating your own linker script, you must:
     - Arm - The boot stack is the `ARM_LIB_STACK` region.
     - GCC_ARM - The boot stack starts at the symbol `__StackLimit` and ends at the symbol `__StackTop`.
 - Add defines for a relocatable application - `MBED_APP_START` and `MBED_APP_SIZE`.
-- Add the define for boot stack size - `MBED_BOOT_STACK_SIZE`.
+- Add the define for boot stack size - `MBED_CONF_TARGET_BOOT_STACK_SIZE`.
```
jeromecoutant, line 31: Maybe, indicate that define has to be define in rtos json lib config ?

Patater: I am not sure this is required to specified in the Mbed config. The linkerscript can specify a fallback value. I may be wrong about this being optional, though.

docs/porting/target/bootstrap.md
Patater: The intended (and likely desired) behavior here is to prefer MBED_CONF_TARGET_BOOT_STACK_SIZE if present, and to fall back to MBED_BOOT_STACK_SIZE if not. If MBED_BOOT_STACK_SIZE is not defined, then we fall back to a hard-coded value.

jeromecoutant: Yes, I can see this :-)
But which uses case do you see to add this fall back system ?

iriark01:
@Patater the code PR was merged. Please can you work through the discussions here so we can merge this?

Patater:
Nothing left to discuss on docs. The doc matches what landed on Mbed OS. This PR should be merged.

PR #1367, opened 2020-09-18
Fix inconsistency in Arm Compiler version

LDong-Arm:
The table documents the current Arm Compiler supported by Mbed OS to be 6.14, but the link text still says 6.13.
This PR fixes the inconsistency and updates the MDK version to 5.31 which contains Arm Compiler 6.14.


evedon (commented):
That version of Keil MDK does not support 6.14.
But http://www2.keil.com/mdk5/531 does

LDong-Arm:
> That version of Keil MDK does not support 6.14.
> But http://www2.keil.com/mdk5/531 does

Good finding. I wonder if we can just update the MDK link.

LDong-Arm:
Updated the MDK link

PR #1369, opened 2020-09-30
Editing TensorFlow tutorial

iriark01:
Ready for engineering review. 

Please look at the source for commented-out questions. 

You can also see a preview (on the VPN): https://os.mbed.com/docs/mbed-os/TensforFlow_edit/mbed-os-pelion/machine-learning-with-tensorflow-and-mbed-os.html


docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
 Here is what you will need to complete the guide:
 
-* Computer that supports Mbed CLI (version 1.10.4)
-
-* NXP FRDM K66F
-
-* Mini-USB cable
-
-* Python 2.7. Using pyenv is recommended to manage python versions.
+* Ubuntu 20.04 LTS. If you are using a Windows machine, install [VirtualBox](https://www.virtualbox.org/wiki/Downloads)<!--originally you said 6.0 - that version's support ended in July. Can they install the latest version, or do they need 6.0 specifically?--> and then [Install Ubunto on it](https://www.virtualbox.org/wiki/Linux_Downloads).
```
spartacosan: Regarding VirtualBox, it should not make a difference as long as they are able to run Ubuntu 20.04 on it. Although their website states "Please also use version 6.0 if you need to run VMs with software virtualization, as this has been discontinued in 6.1. Version 6.0 will remain supported until July 2020". So, I would suggest that for the sake of consistency users stick with version 6.0 to ensure the demo works accordingly. 

 Also, there is a tiny spelling mistake, it should be "[Install Ubuntu on it]", not "[Install Ubunto on it]".

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
-![IMAGE ALT TEXT HERE](http://img.youtube.com/vi/x5MhydijWmc/0.jpg)
+    <span class="notes">**Note**: On a Mac, you might have to use gmake instead of make.</span>
 
-[Watch here](http://www.youtube.com/watch?v=x5MhydijWmc)
+* Python 2.7. We recommend [using pyenv to manage Python versions](https://pypi.org/project/pyenv/).<!--The install page for Mbed CLI now requires 3.7. Why are we asking for 2.7?-->
 
+* Git.
 
-**Getting started**
+* xxd.<!--unless is exists by default on all Linux computers?-->
```
spartacosan: Well, you are right, it does come with Ubuntu although it could be the case some minimalist distros don't include it for some reason. But since the tutorial focuses on Ubuntu we could leave this out.

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
+    ```
 
-**How to install Ubuntu 20.04 on VirtualBox in Windows 10**
+* Make 3.82 or newer (if you are using an older version, see the troubleshooting section for instructions).
 
-![IMAGE ALT TEXT HERE](http://img.youtube.com/vi/x5MhydijWmc/0.jpg)
+    <span class="notes">**Note**: On a Mac, you might have to use gmake instead of make.</span>
 
-[Watch here](http://www.youtube.com/watch?v=x5MhydijWmc)
+* Python 2.7. We recommend [using pyenv to manage Python versions](https://pypi.org/project/pyenv/).<!--The install page for Mbed CLI now requires 3.7. Why are we asking for 2.7?-->
```
spartacosan: We tried with Python 3.7 but had issues from the TensorFlow side. See point 5 [here](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/lite/micro/examples/micro_speech#deploy-to-nxp-frdm-k66f).

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
-**Getting started**
+* xxd.<!--unless is exists by default on all Linux computers?-->
 
-TensorFlow Lite for Microcontrollers supports several devices out of the box, and is relatively easy to extend to new devices. For this guide, we will focus on the **NXP FRDM K66F**.
+* A development board that can run [TensorFlow Lite for Microcontrollers](https://www.tensorflow.org/lite/microcontrollers#supported_platforms). We tested on the [NXP FRDM-K66F](https://os.mbed.com/platforms/FRDM-K66F/) development board.
 
-![IMAGE ALT TEXT HERE](https://raw.githubusercontent.com/COTASPAR/K66F/master/images/nxp_k66f_board.jpeg)
+* A mini-USB cable.
 
+<!--MarkDown doesn't support embedded YouTube players, which is one reason I removed the videos. Another is that it's better practice to link to official sites, not to videos uploaded by "unofficial" users who may remove them without warning and may, for all we know, be in legal disputes with the companies who created those products (I admit that isn't very likely with "how to install" videos). Also, not sure these videos will work in all countries.-->
```
spartacosan: Understood, I'll keep that in mind for future guides.

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
-1.  Download and build the sample application
+1. Navigate into the project directory and build it:
 
+    ```
+    cd tensorflow
 
-2.  Deploy the sample to your K66F
+    make -f tensorflow/lite/micro/tools/make/Makefile TARGET=mbed TAGS="nxp_k66f" generate_micro_speech_mbed_project
+    ```
+    <!--I got a "no such file or directory error". I think it's because there was already `cd tensorflow` and then the path in the command also has `tensorflow`. Removing `tensorflow` from the path removed the error-->
```
spartacosan: Okay, just to confirm, the full directory is tensorflow/tensorflow/lite/micro/tools/make/Makefile meaning, you

> git clone https://github.com/tensorflow/tensorflow.git
> cd tensorflow
> make -f tensorflow/lite/micro/tools/make/Makefile TARGET=mbed TAGS="nxp_k66f" generate_micro_speech_mbed_project 

and this should not generate an error. Correct?

iriark01: When I CDed into tensorflow and then also used it in the path, I think it went tensorflow/tensorflow/lite
It worked only when I removed the tensorflow from the path

iriark01: I can't tell you if it actually *works* because Brew won't update my Make. But it stopped giving the path error.

spartacosan: Ok

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
+    mbed deploy
 
-While you wait for the project to download, let’s explore the project files on [GitHub](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/lite/micro/examples/micro_speech) and learn how this TensorFlow Lite for Microcontrollers example works.
+    mbed compile -m K66F -t GCC_ARM
+    ```
 
-The code samples audio from the microphone on the K66F. The audio is run through a Fast Fourier transform to create a spectrogram. The spectrogram is then fed into a pre-trained machine learning model. The model uses a [convolutional neural network](https://developers.google.com/machine-learning/practica/image-classification/convolutional-neural-networks) to identify whether the sample represents either the command “yes” or “no”, silence, or an unknown input. We will explore how this works in more detail later in the guide.
+    <!--The install page for Mbed CLI now requires 3.7. Why are we asking for 2.7?-->
+    <!--need to explain what `config root .` and `deploy` do and why we need them - neither one is part of a standard workflow where you use Mbed CLI to import an application, so this is a special case-->
```
spartacosan: From what I understand, `mbed config root .` sets the current mbed-os folder as the root (meaning `mbed` command actually knows what Mbed OS project we want to run the following commands on) which then allows us to run `mbed deploy` which can pull the libraries needed (in our case in tensorflow/tensorflow/lite/micro/tools/make/gen/mbed_cortex-m4/prj/micro_speech/mbed/mbed-os/requirements.txt) and installs them. I was not aware that this is not a standard workflow, this is the workflow presented in the [original guide hosted on developer.arm.com](https://developer.arm.com/solutions/machine-learning-on-arm/developer-material/how-to-guides/build-arm-cortex-m-voice-assistant-with-google-tensorflow-lite/download-and-build-the-sample-application). If you know of a better way of doing this I'd be glad to explore that workflow and update the guide.

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
+    <!--and why are we compiling here? We compile again two steps down, with the flash parameter-->
```
spartacosan: You are right. This is the format presented in the original guide referenced above. However, we could take that out. One reason could be that we compile to ensure everything is working and only then we compile + flash the device.

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
 
-Here are descriptions of some interesting source files:
+    For some compilers<!--we only support two, and you specifically asked for GNU, so in what case will this happen?-->, you may get a compilation error in `mbed_rtc_time.cpp`. Go to  `mbed-os/platform/mbed_rtc_time.h`  and comment out line 32 and line 37:
```
spartacosan: That was coming from the TensorFlow instructions at some point while we were developing the demo. However, I can't seem to find this warning anymore on their repo other than in the K66F instructions which I pushed myself but I just tried this workflow again and it works without commenting anything out so they must have fixed this issue. We can omit this step.

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
+1. Connect your board to your computer over USB. On your board, when the Ethernet port is facing you, the micro USB port is to its left.
 
--   [micro_features/model.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/examples/micro_speech/micro_features/model.cc). This file is the machine learning model itself, represented by a large array of unsigned char values.
+    Your board appears as storage on your computer. If your system does not recognize the board with the `mbed detect` command, follow the instructions for setting up [DAPLink](https://armmbed.github.io/DAPLink/?board=FRDM-K66F).
 
--   [command_responder.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/examples/micro_speech/command_responder.cc) is called every time a potential command has been identified.
+1. Flash the application to the board:
+    <!--I compiled two steps ago - that probably needs to be removed-->
```
spartacosan: Yes, we could either remove the previous compile step or this one to be less redundant. I'd say it might be better to keep this one as it flashes automatically to the board.

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
+    ```
+    mbed compile -m K66F -t GCC_ARM –flash
+    ```
+1. Deploy the example to your K66F<!--but we flshed already. What are we doing here?-->
```
spartacosan: Again, following the format in the original [guide on developer.arm.com](https://developer.arm.com/solutions/machine-learning-on-arm/developer-material/how-to-guides/build-arm-cortex-m-voice-assistant-with-google-tensorflow-lite/single-page). They had two ways of doing things. One where we compile and flash on the same command `mbed compile -m K66F -t GCC_ARM –flash` and one where we just compile then copy the binary over to the board to flash it, as shown below with `cp ./BUILD/K66F/GCC_ARM/mbed.bin /Volumes/K66F/` .

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
 
--   [main.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/examples/micro_speech/main.cc). This file is the entry point for the Mbed program, which runs the machine learning model using TensorFlow Lite for Microcontrollers.
+    Copy the binary file that we built earlier to the USB storage.
 
+    Note: if you have skipped the previous steps<!--which ones?-->, download the [binary file]() to proceed.
```
spartacosan: all the steps up to compiling steps

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
 
-After the project has downloaded, you can run the following commands to navigate into the project directory and build it:
+    The file is at:
 
-```cd tensorflow```
+    ```
+    cp ./BUILD/K66F/GCC_ARM/mbed.bin /Volumes/K66F/
+    ```
 
+    Depending on your operating system <!--aren't we making everyone use Linux?-->, the exact copy command and paths may vary.
```
spartacosan: Good catch. I should have removed this from the original guide.

iriark01:
On hold until it's updated to Mbed OS 6

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
 
-# Deploy the example to your K66F
+To solve this problem, our code runs inference as often as it can<!--here you lost me, probably because I don't deal with ML - is this in the training again? where is it getting new samples?-->, depending on the speed of the device, and averages all of the results within a rolling 1000ms window. The code in the file [recognize_commands.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/examples/micro_speech/recognize_commands.cc) performs this action. When the average for a given category in a set of predictions goes above the threshold<!--lost me here too - are we talking about training or recognising actual users?-->, as defined in [recognize_commands.h](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/examples/micro_speech/recognize_commands.h), we can assume a valid result.
```
spartacosan: This part doesn't refer to training data but rather samples when the model is running on the device. Because we don't know at what point in the "window" a person will say the keyword this means we need to run inference many times so that we effectively take many snapshots of many windows then we analyse those "windows" and then take the average result as that's what is most likely being said. For a deeper explanation you can search "Windowing" on this [book on TinyML](https://tinymlbook.files.wordpress.com/2020/01/tflite_micro_preview.pdf).

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
--   Look at how the TensorFlow model gets converted to the TensorFlow Lite format.
+        ```
+        const int g_tiny_conv_micro_features_model_data_len = 18216;
+        ```
 
--   Convert the TensorFlow Lite model into a C source file.
+    1. Save the file.
 
--   Modify the code and deploy to the device.
+1. Copy the `tiny_conv_micro_features_model_data.cc` <!--where is this file? it's not the one I was just editing-->file into the `tensorflow/tensorflow/lite/micro/tools/make/gen/mbed_cortex-m4/prj/micro_speech/mbed/tensorflow/lite/micro/examples/micro_speech/micro_features` directory.
```
spartacosan: Apologies. This file was originally called `tiny_conv_micro_features_model_data.cc` when the project was[ first made](https://github.com/uTensor/tf_microspeech/blob/master/micro_features/tiny_conv_micro_features_model_data.cc) but this is now just[ `micro_features/model.cc`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/examples/micro_speech/micro_features/model.cc) .

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
+    const char* kCategoryLabels[kCategoryCount] = {  
+    "silence",  
+    "unknown",  
+    "yes",  
+    "no",  
+    };
+    ```
 
-To save time, we will skip this step and instead download the [tiny_conv.tflite](https://developer.arm.com/-/media/Files/downloads/Machine%20learning%20how-to%20guides/tiny_conv.tflite?revision=495eb362-4325-49b8-b3ba-3141df0c9b95&la=en&hash=0F37BA2C5DE95A1561979CDD18973171767A47C3).
+    The code uses this array to map the output of the model to the correct value. Because we specified our wanted_words<!--is that underscore correct?--> as “up, down” in the training script, we should update this array to reflect these words in the same order.
```
spartacosan: Yes, the underscore is correct but it should actually be all caps "WANTED_WORDS" as on the [python training script](https://github.com/tensorflow/tensorflow/blob/5d4a29eaf590b4a3068ef4d0b7bea9d4f7bd9369/tensorflow/lite/micro/examples/micro_speech/train/train_micro_speech_model.ipynb).

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,394 @@
...
-First, go to the following directory:
+1. Rebuilt the project:
 
-```tensorflow/lite/micro/examples/micro_speech/```
+    ```
+    mbed compile -m K66F -t GCC_ARM
+    ```
 
-and open the file:
+1. Copy the binary to the USB storage of the device<!--why aren't we using flash?-->, using the same method that you used earlier.
```
spartacosan: I guess we can leave this out and show it as before. I'm not sure what the logic behind showing two ways of doing it was when the original guide was made.

spartacoos (commented):
Let me know if I missed any of your questions or if you need me to look at some of the questions in more detail such as why we are flashing in two different ways. I can discuss with Alessandro Grande from Arm who made the original guide with Pete Warden from Google.

spartacoos (changes requested):
I think this covers your comments plus the porting to Mbed OS 6.3. Feel free to ping me on Slack if anything else needs changing.

docs/reference/TensorFlow.md
iriark01: The first compile step now has to come out, since it compiled with an old version of Mbed OS.

docs/reference/TensorFlow.md
iriark01:
```suggestion

```

docs/reference/TensorFlow.md
iriark01:
```suggestion
* Ubuntu 20.04 LTS. If you are using a Windows machine, install [VirtualBox](https://www.virtualbox.org/wiki/Downloads) and then [Install Ubuntu on it](https://www.virtualbox.org/wiki/Linux_Downloads).
```

docs/reference/TensorFlow.md
iriark01:
```suggestion
To solve this problem, our code runs inference as often as it can, depending on the speed of the device, and averages all of the results within a rolling 1000ms window. The code in the file [recognize_commands.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/examples/micro_speech/recognize_commands.cc) performs this action. When the average for a given category in a set of predictions goes above the threshold, as defined in [recognize_commands.h](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/examples/micro_speech/recognize_commands.h), we can assume a valid result.
```

docs/reference/TensorFlow.md
iriark01:
```suggestion
1. Copy the `model.cc` file into the `tensorflow/tensorflow/lite/micro/tools/make/gen/mbed_cortex-m4/prj/micro_speech/mbed/tensorflow/lite/micro/examples/micro_speech/micro_features` directory.
```

docs/reference/TensorFlow.md
iriark01:
```suggestion
    The code uses this array to map the output of the model to the correct value. Because we specified our WANTED_WORDS as “up, down” in the training script, we should update this array to reflect these words in the same order.
```

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
+    ```
+    cd <path_to_tensorflow>/tensorflow/lite/micro/tools/make/gen/mbed_cortex-m4/prj/micro_speech/mbed
+    ```
 
-First, go to the following directory:
+1. Rebuilt the project:
 
-```tensorflow/lite/micro/examples/micro_speech/```
+    ```
+    mbed compile -m K66F -t GCC_ARM
```
iriark01:
```suggestion
    mbed compile -m K66F -t GCC_ARM --flash
```

docs/reference/TensorFlow.md
iriark01:
```suggestion

```

docs/reference/TensorFlow.md
iriark01: And that doesn't cause problems with Mbed CLI?

docs/reference/TensorFlow.md
iriark01: @donatieng can you confirm/deny?

iriark01 (commented):
Need to approve some changes I've made, and I would like @donatieng to weigh in on some things.

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
+<!--https://github.com/COTASPAR/K66F/blob/master/README.md-->
+As an IoT developer, you might think of machine learning as a server-side technology: sensors on your device capture data and send it to the cloud, where Machine Learning (ML) models on hefty machines make sense of it. A network connection is obligatory, and you are going to expect some latency, not to mention hosting costs.
 
-As an IoT developer, you might think of machine learning as a server-side technology. In the traditional view, sensors on your device capture data and send it to the cloud, where Machine Learning (ML) models on hefty machines make sense of it. A network connection is obligatory, and you are going to expect some latency, not to mention hosting costs.
+But with the launch of [TensorFlow Lite for Microcontrollers](https://www.tensorflow.org/lite/microcontrollers), you can run machine learning inference on extremely low-powered devices, like the Cortex-M microcontroller series. You can deploy your ML models to the IoT devices themselves, bring ML closer to your sensors, remove your reliance on a network connection, and skip the round trip to the server for lower latency.
```
donatieng:
```suggestion
But with the launch of [TensorFlow Lite for Microcontrollers](https://www.tensorflow.org/lite/microcontrollers), you can run machine learning inference on extremely low-power devices, like Arm Cortex-M series-based microcontrollers. You can deploy your ML models to the IoT devices themselves, bringing ML closer to your sensors, remove your reliance on a network connection, and skip the round trip to the server for lower latency. A primary benefit of doing local inference is that you're reducing your power consumption compared to transferring data over a wireless connection for instance.
```

donatieng: Also a key benefit of doing local inference is that you're reducing your power consumption compared to transferring data over a wireless connection for instance.

docs/reference/TensorFlow.md
donatieng: This all feels very specific. Can we reword and say that this tutorial has been tested on Ubuntu 20.04 - because presumably it will work on other platforms (or even other versions of Ubuntu)? Also VirtualBox is one suggested option but other virtualization SW would work too.

docs/reference/TensorFlow.md
donatieng: Python 3.x please! What are the issues on the TF-side (apart from outdated tutorial)?

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
 
-We will deploy a sample application that uses the microphone on the K66F and a TensorFlow machine learning model to detect the words “yes” and “no”.
+1. Navigate into the project directory:
 
-To do this, we will show you how to complete the following steps:
+    ```
+    cd tensorflow/tensorflow/lite/micro/mbed/
+    ```
+    
+1. Change the DebugLog function to use UnbufferedSerial instead since the Serial API was deprecated in Mbed OS 6:
```
donatieng, line 59: Can we upstream this? Or link an open PR in the doc?

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
+    cd tensorflow/lite/micro/tools/make/gen/mbed_cortex-m4/prj/micro_speech/mbed
+    ```
+1. Connect your board to your computer over USB. On your board, when the Ethernet port is facing you, the micro USB port is to its left.
+    Your board appears as storage on your computer. If your system does not recognize the board with the `mbed detect` command, follow the instructions for setting up [DAPLink](https://armmbed.github.io/DAPLink/?board=FRDM-K66F).
+    
+1. Execute the following to flash the application onto the board:
 
-# Download and build the sample application
+    ```
+    mbed config root .
```
donatieng, line 90: Probably a shortcoming of the TFLiteU Makefile above, ideally this wouldn't be needed so candidate for upstreaming too.

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
 
-**Install Arm toolchain and Mbed CLI**
--   Download [Arm cross compilation](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads) toolchain. Select the correct toolchain for the OS that your computer is running. For Windows users, if you have already set up the Linux virtual environment, install the toolchain there.
+    mbed deploy
+    
+    mbed update mbed-os-6.3.0
```
donatieng: 6.4 was just released ;)

donatieng: Should also be upstreamed

docs/reference/TensorFlow.md
donatieng: See above, I think this is a shortcoming of the Makefile here

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
--   After Mbed CLI is installed, tell Mbed where to find the Arm embedded toolchain.  
-  ```    
- mbed config -G GCC_ARM_PATH <path_to_your_arm_toolchain>/bin
- ```
-
-**Important:** We recommend running the following commands from inside the Mbed CLI terminal that gets launched with the Mbed CLI Application. This is because it will be much quicker to set up, because it resolves all your environment dependencies automatically.
+1. Speak into the board's microphone: Saying "Yes" will print "Yes" and "No" will print "No" on the serial port:
 
+    ```
+    Heard yes (208) @116448ms
```
donatieng, line 103: What's 208?

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
+## Background: Converting the trained model
 
-The board’s USB UART shows up as  ```/dev/tty.usbmodemXXXXXXX```.We can use ‘screen’ to access the serial console. Although, ‘screen’ is not installed on Linux by default, you can use apt-get install screen to install the package.
+To get a converted model that can run on the controller itself from the trained model, we need to run a conversion script: the [TensorFlow Lite converter](https://www.tensorflow.org/lite/convert). This tool makes our model as small and efficient as possible, and converts it to a TensorFlow Lite FlatBuffer. To reduce the size of the model, we used a technique called [quantization](https://www.tensorflow.org/lite/performance/post_training_quantization). All weights and activations in the model get converted from 32-bit floating point format to an 8-bit and fixed-point format, as you can see in the following command:
 
-Run the following command in a separate terminal:
+<span class="images">![Convert the model to the TensorFlow Lite format code](https://raw.githubusercontent.com/COTASPAR/K66F/master/images/convert_model.png)</span>
```
donatieng: Can this image be made part of this PR's scope?

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
-Heard unknown (241) @117984ms  
-Heard no (201) @124992ms
+1. Download the [tiny_conv.tflite file](https://developer.arm.com/-/media/Files/downloads/Machine%20learning%20how-to%20guides/tiny_conv.tflite?revision=495eb362-4325-49b8-b3ba-3141df0c9b95&la=en&hash=0F37BA2C5DE95A1561979CDD18973171767A47C3).
+1. Convert this model into a C file that works with Mbed OS projects.
 
-```
+    Use a tool called xxd to generate a file called `model.cc`:
 
-Congratulations! You are now running a machine learning model that can recognize keywords on an Arm Cortex-M4 microcontroller, directly on your K66F.
+    ```
```
donatieng, line 200: What is it? What does it do?

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
-We are going to use another pre-trained model to recognize “up” and “down”, instead. If you are interested in the full flow including the training of the model refer to the [Supplementary information: model training](https://developer.arm.com/solutions/machine-learning-on-arm/developer-material/how-to-guides/build-arm-cortex-m-voice-assistant-with-google-tensorflow-lite/supplementary-information-model-training) section of this guide.
+    1. Go to the bottom of the file and find the unsigned int variable:
 
-To build our new ML application we will now follow these steps:
+        ```
+        unsigned int tiny_conv_tflite_len = 18216;
+        ```
 
--   Download a pretrained model that has been trained and frozen using TensorFlow.
+    1. Change the declaration to the following code, but do not change the number assigned to it, even if your number is different from the one in this guide:
```
donatieng: Why? What is it?

donatieng: "Change the variable's name and constness to"?

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
-Finally, save the file, then copy the ```tiny_conv_micro_features_model_data.cc```file into the ```tensorflow/tensorflow/lite/micro/tools/make/gen/mbed_cortex-m4/prj/micro_speech/mbed/tensorflow/lite/micro/examples/micro_speech/micro_features``` directory.
+    ```
+    void RespondToCommand(tflite::ErrorReporter* error_reporter,  
+    int32_t current_time, const char* found_command,  
+    uint8_t score, bool is_new_command) {  
+        if (is_new_command) {  
+            error_reporter->Report("Heard %s (%d) @%dms", found_command, score,  
+            current_time);  
+        if(strcmp(found_command, "up") == 0) {  
+            lcd. Clear(0xFF0F9D58);
```
donatieng: Where did that LCD screen come from 😄?

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
 
-In the previous section of this guide, we explained the build process for a keyword spotting example application.
+### Interpreting the results
 
-Now that the build has completed, we will look in this section of the guide at how to deploy the binary to the K66F and test to see if it works.
+The RespondToCommand method in [command_responder.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/micro/examples/micro_speech/command_responder.cc) is called when a command a user speals has been recognised. Currently, this results in a line being printed to the serial port. Later in this guide, we will modify the code to display the result on the screen.
```
donatieng: The K66F board doesn't have a screen - do we need a shield?

docs/reference/TensorFlow.md
```diff
@@ -1,405 +1,370 @@
...
-```
+    ```
+    export PS1='> '
+    ```
 
-**Note**: On a Mac, you might have to use gmake instead of make, to run your commands.
+- Error: `Requires make version 3.82 or later (current is 3.81)`
 
-Error: -bash: mbed: command not found
+    If you encounter this error, install Brew and Make:
```
donatieng: Assuming you're on a Mac

donatieng (changes requested):
Thanks @COTASPAR - some feedback below.

docs/reference/TensorFlow.md
spartacosan: Many of the other TFLu examples still use older versions of Mbed OS and thus if we upstream it that could break those examples. Ideally, all of the examples would be ported to the latest version of mbed. However, I think for now this should be okay while perhaps we try to engage with the TFLu team to port the other examples.

docs/reference/TensorFlow.md
spartacosan: This is a mistake. This board does not have an LCD screen and the command_responder.cc should not be changed if the model is re-trained.

docs/reference/TensorFlow.md
spartacosan: No, it doesn't have a screen nor did we use a shield. This is a mistake, the change to command_responder.cc is for boards with LCD displays. It does not apply to the K66F.

docs/reference/TensorFlow.md
spartacosan: That number is auto-generated by the converter script and the constant is the length of the model parameters. I suggest the following change "Change the declaration to the following code, but do not change the number assigned to it, as it is automatically generated to indicate the size of the model. Even if your number is different from the one in this guide:"

docs/reference/TensorFlow.md
spartacosan: Already described above "Convert this model into a C file that works with Mbed OS projects.

Use a tool called xxd to generate a file called model.cc:"

docs/reference/TensorFlow.md
spartacosan: This should be hosted on Arm's GitHub and link from there.

docs/reference/TensorFlow.md
spartacosan: Below the output (on line 116) we could add:
> the output shows found_command, score and current_time as described in the command_responder.cc file.

docs/reference/TensorFlow.md
alessandroDevs-zz: Agreed can we say: 

> This tutorial has been tested on Ubuntu 20.04 LTS. If you are using a Windows machine, we recommend using virtualization software to create a virtual Linux machine to run this tutorial.

docs/reference/TensorFlow.md
alessandroDevs-zz: are we good here?

docs/reference/TensorFlow.md
alessandroDevs-zz: As soon as this is published we will start to work with the TF team to discuss next steps.

docs/reference/TensorFlow.md
alessandroDevs-zz: @donatieng in order to feature proof this is there a way to update to the latest Mbed version? Perhaps the latest known version without breaking changes.

docs/reference/TensorFlow.md
alessandroDevs-zz: In the interest of getting this published, we could remove the below:

> Error: Requires make version 3.82 or later (current is 3.81)
> 
> If you encounter this error, install Brew and Make:
> 
> ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
> 
> brew install make

docs/reference/TensorFlow.md
spartacosan: Yes, just tested again with a Python 3.7 environment and it worked fine.

docs/reference/TensorFlow.md
donatieng: I meant, what is `xxd`? As a user, I don't want to run some arbitrary shell command on my machine without understanding what it does 😉

docs/reference/TensorFlow.md
donatieng: The "value"? (as opposed to "number")

docs/reference/TensorFlow.md
donatieng: Unfortunately we don't have a "update to the latest Mbed OS 6 release" command; good feedback though, it would be good to look at that in the new tools

docs/reference/TensorFlow.md
donatieng: (or you could say "if you're using macOS")

docs/reference/TensorFlow.md
donatieng: Should that be removed then?

docs/reference/TensorFlow.md
spartacosan: Ah, I see. We could add "[xxd](https://www.tutorialspoint.com/unix_commands/xxd.htm) is a UNIX command-line utility that creates a hex dump of a given file or standard input. It can also convert a hex dump back to its original binary form."

docs/reference/TensorFlow.md
alessandroDevs-zz: sounds good let's go with value 👍

docs/reference/TensorFlow.md
alessandroDevs-zz: we could, but not sure this is extremely relevant as this was an error that used to occur a while back, it might be obsolete by now

donatieng: OK, worth removing altogether then!

docs/reference/TensorFlow.md
spartacosan: Yes

docs/reference/TensorFlow.md
spartacosan: The question below on the [VMs](https://github.com/ARMmbed/mbed-os-5-docs/pull/1369#discussion_r512669546) doesn't let me respond so I'll say here: Yes, I  agree. Should we just say tested on Ubuntu 20.04? rather than walk users through how to get Ubuntu on their machine.

docs/reference/TensorFlow.md
donatieng: Needs to be updated to a relative link and image part of the PR

spartacoos (approved):
Implemented the changes from the comments.

donatieng (approved):
Did a bit of clean-up but it looks good, thanks a lot for the work @COTASPAR !

PR #1378, opened 2020-11-23
Update Bluetooth examples in the docs + Move Mbed OS references to v6.5 instead of v6.0

donatieng:
* Update Bluetooth examples in the docs
* Move Mbed OS references to v6.5 instead of v6.0
These two things need to happen at once as v6.5 Bluetooth examples have changed, and therefore the development doc cannot reference these new examples using v6.0 tags. For consistency we update all references to v6.5.


paul-szczepanek-arm (approved):

pan- (approved):

donatieng:
Meta-docs update: https://github.com/ARMmbed/documentation_docs/pull/31

ghost (approved):

PR #1379, opened 2020-11-24
update Arm compiler version support to 6.15 

saheerb:
mbed-os versions after 6.5 will start supporting Arm Compiler 6.15.


donatieng (approved):
LGTM @saheerb !

PR #1380, opened 2020-11-24
docs: Add description about flash option

urutva:
If `mbedtools compile` command is invoked with `-f/--flash` option then
the built binary will be flashed onto the connected device.

### DIFF
--- a/docs/tools/mbed_cli_2/use.md
+++ b/docs/tools/mbed_cli_2/use.md
@@ -178,6 +178,14 @@ To perform an iterative build on a previously configured project:
 mbed-tools build
 ```
 
+## Flashing the built program
+
+You can flash the built program to the connected target by adding the -f/--flash argument to the compile command:
+
+```
+mbed-tools compile -f
+```
+
 ## List an application's library dependencies
 
 To list all library dependencies with their URL and Git reference:


docs/tools/mbed_cli_2/use.md
```diff
@@ -178,6 +178,12 @@ To perform an iterative build on a previously configured project:
 mbed-tools build
 ```
 
+## Flashing the built program
+You can flash the built program to the connected target by adding the -f/--flash argument to the compile command:
+```
+mbed-tools build -f
```
jainvikas8: should we use `compile` here, I'm assuming we are going to merge this PR after new Mbed tools 3.6.0 release

jainvikas8 (changes requested):

docs/tools/mbed_cli_2/use.md
urutva: Ah good point. Copy/paste error.

jainvikas8 (approved):

docs/tools/mbed_cli_2/use.md
```diff
@@ -178,6 +178,12 @@ To perform an iterative build on a previously configured project:
 mbed-tools build
 ```
 
+## Flashing the built program
+You can flash the built program to the connected target by adding the -f/--flash argument to the compile command:
```
jainvikas8, line 183: Please add new lines when needed, please refer above section

urutva: fixed.

Patater (approved):

PR #1386, opened 2020-12-07
TDBStore no longer requires flash behaviour

LDong-Arm:
Since https://github.com/ARMmbed/mbed-os/pull/11987 was merged to Mbed OS, a flash device or a simulated flash device is no longer a requirement.


evedon (approved):

evedon:
@donatieng I don't have permission to merge

PR #1388, opened 2020-12-09
Fix memory leak in TF guide (development)

donatieng:


donatieng:
Approved offline by @alessandroDevs

PR #1395, opened 2021-01-15
Add deprecation section

0xc0170:
Document what it means to deprecate a repository. We haven't yet documented this but already have read only repos.

I added a simple commit to update the maintainers list as I am touching the same file already).


0xc0170:
@ARMmbed/mbed-os-maintainers @donatieng  Please review

donatieng (approved):

PR #1407, opened 2021-02-02
porting: Update the custom targets guide for CMake and Mbed CLI 2

Patater:
The custom targets documentation didn't cover how to make your custom
target buildable with CMake. Update it to cover all aspects of adding a
custom target such that it can be build using Mbed OS's CMake build
system.

### DIFF
--- a/docs/porting/custom-target-porting.md
+++ b/docs/porting/custom-target-porting.md
@@ -76,12 +76,15 @@ Follow these steps to create a custom port for Mbed OS:
    1. The `detect_code` changes from `0764` to `1234`. The `detect_code` is a
       unique four-digit hexadecimal value, also called a `Platform ID`, that
       identifies the board to the Mbed OS test tools. For Mbed Enabled boards,
-      this number is exposed through the debug interface with Mbed CLI by
-      typing `mbedls`.
-   1. The `macros_add` section changes to remove `USBHOST_OTHER` because the
-      new board does not use USB.
-   1. The `device_has_add` section changes to remove the `ANALOGOUT`, `CAN`,
-      and `USBDEVICE` drivers because the new board doesn't use those features.
+      this number is exposed through the debug interface with Mbed CLI 2 by
+      typing `mbed-tools detect`.
+   1. The `device_has_add` section changes to remove `USBDEVICE` because the
+      new board doesn't use that feature.
+   1. Add a `device_has_remove` section to remove `ANALOGOUT`, `ANALOGIN`,
+      `CAN`, `I2CSLAVE`, and `I2C_ASYNC` because the new board doesn't use
+      those features. The reason why `device_has_remove` is used in this case
+      is because the new board is inheriting from the MCU Family configuration
+      `MCU_STM32L475xG`, which has those device features by default.
 
    After making changes, the full contents look like this:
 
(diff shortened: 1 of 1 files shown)


docs/porting/custom-target-howto.md
```diff
@@ -0,0 +1,117 @@
...
+Next, we need to tell CMake about your custom target.
+
+The most common thing to add for custom targets is a custom pin configuration. As such, we'll assume you'll be wanting to customize your custom board's pin configuration too. To do this, you'll need to make a few files: at least a `PinNames.h` and possibly also a `PeripheralPins.h` or a `PeripheralNames.h`. Please refer to [the Mbed OS porting guide](https://os.mbed.com/docs/mbed-os/latest/porting/porting-custom-boards.html) on how to configure the pins for your board and which files exactly you'd need to create.
+
+```
+mbed-os-example-blinky
+├── CMakeLists.txt
+├── CONTRIBUTING.md
+├── custom_targets.json
+├── IMAGINARY              <--- The new folder with new files in it!
```
urutva: Currently, we follow `TARGET_<BOARD_NAME>` naming convention for board/target related files. Can we follow the same convention in the example as well?

docs/porting/custom-target-howto.md
```diff
@@ -0,0 +1,117 @@
...
+    ├── CMakeLists.txt
+    ├── PeripheralNames.h
+    ├── PeripheralPins.h
+    └── PinNames.h
```
urutva: From [the Mbed OS porting guide](https://os.mbed.com/docs/mbed-os/latest/porting/porting-custom-boards.html), `PeripheralPins.c` should also be added here.

docs/porting/custom-target-howto.md
```diff
@@ -0,0 +1,117 @@
...
+├── LICENSE
+├── main.cpp
+├── mbed-os.lib
+├── README.md
+└── resources
+    └── official_armmbed_example_badge.png
+```
+
+The CMakeLists.txt file should have at a minimum a new library target, as well as a list of which files or include directories your target needs. The [`target_link_libraries()`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) directive should specify which MCU CMake target your custom board depends on, so you'll enter here the MCU Cmake target name you noted previously.
```
urutva:
```suggestion
The ./IMAGINARY/CMakeLists.txt file should have at a minimum a new library target, as well as a list of which files or include directories your target needs. The [`target_link_libraries()`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) directive should specify which MCU CMake target your custom board depends on, so you'll enter here the MCU Cmake target name you noted previously.
```

urutva: Would adding the folder name helpful?

urutva (changes requested):

docs/porting/custom-target-howto.md
Patater: The targets aren't named TARGET_BLAH. The target name is actually just BLAH. For folder names, given the limitation of the old tools, we did name things TARGET_BLAH. I'll do that here for old time's sake.

docs/porting/custom-target-howto.md
Patater: Many of these are optional.

docs/porting/custom-target-howto.md
Patater: Good idea

Patater: I've removed this file, so this probably isn't relevant anymore.

urutva (approved):

0xc0170 (approved):

PR #1412, opened 2021-02-10
Mbed CLI 2 - Update with support for multiple targets

wernerlewis:
As Mbed CLI 2 now supports building for multiple targets in subdirectories, its useful to document this behavior.

In addition, as target and toolchain arguments are now required with `compile`, these have been added where needed. And a minor change has also been made to correct the message shown for the `configure` example.

### DIFF
--- a/docs/tools/mbed_cli_2/use.md
+++ b/docs/tools/mbed_cli_2/use.md
@@ -139,7 +139,7 @@ The Mbed OS configuration system parses the configuration files in your project 
 
     ```
     mbed-tools configure -m K64F -t GCC_ARM
-    mbed_config.cmake has been generated and written to '/Users/UserName/Development/Blinky/.mbedbuild'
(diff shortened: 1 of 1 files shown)


docs/tools/mbed_cli_2/use.md
```diff
@@ -139,7 +139,7 @@ The Mbed OS configuration system parses the configuration files in your project
 
     ```
     mbed-tools configure -m K64F -t GCC_ARM
-    mbed_config.cmake has been generated and written to '/Users/UserName/Development/Blinky/.mbedbuild'
+    mbed_config.cmake has been generated and written to '/Users/UserName/Development/Blinky/.mbedbuild/mbed_config.cmake'
```
Patater: We've moved this to the cmake_build location

Patater (changes requested):

Patater (approved):
LGTM

Issue #1414, opened 2021-02-12
"mbed-tools new ." doesn't compile


adepier:
start a new project with mbed-tools
in a new folder type:
 `mbed-tools new .`
then 
`mbed-tools compile -m NUCLEO_L432KC -t GCC_ARM `

you get the following error

> CMake Error at CMakeLists.txt:18 (mbed_configure_app_target):
>   mbed_configure_app_target Function invoked with incorrect arguments for
>   function named: mbed_configure_app_target
![image](https://user-images.githubusercontent.com/1138145/107746550-49374880-6d16-11eb-8e8c-c330943db76c.png)

mbed-tools --version
6.1.0

ciarmcom:
@adepier This issue has an incomplete or old issue template.For future reference please use an up to date clone of the repository before raising issues. Many thanks.

ciarmcom:
Thank you for raising this detailed GitHub issue. I am now notifying our internal issue triagers. 
*Internal Jira reference: https://jira.arm.com/browse/IOTOSM-3439*

PR #1415, opened 2021-02-15
Update use.md

adepier:
after "mbed-tools new .", the solution doesn't compile if you don't update CMakeLists.txt

### DIFF
--- a/docs/tools/mbed_cli_2/use.md
+++ b/docs/tools/mbed_cli_2/use.md
@@ -35,7 +35,7 @@ To create a new Mbed OS project in a specified path:
     - Absolute. `new` will create the folder if it doesn't exist.
     - Relative:
 
-        - If you have already created a project folder, you can use `.`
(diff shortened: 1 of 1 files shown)


docs/tools/mbed_cli_2/use.md
```diff
@@ -35,7 +35,7 @@ To create a new Mbed OS project in a specified path:
     - Absolute. `new` will create the folder if it doesn't exist.
     - Relative:
 
-        - If you have already created a project folder, you can use `.`
+        - If you have already created a project folder, you can use `.`, then update file CMakeLists.txt with: `set(APP_TARGET <created a project folder>` )
```
Patater, line 38: I don't believe we want to require users to manually have to modify CMakeLists.txt. Could you instead please raise a bug on [mbed-tools](https://github.com/ARMmbed/mbed-tools/issues)? It may be that we are missing this line in our template.

adepier: done "mbed-tools new ." doesn't compile #1414

Patater: Thanks

PR #1416, opened 2021-02-15
Update PSA porting guide

LDong-Arm:
This PR
* removes outdated secure/non-secure target naming convention - Mbed OS only provides the non-secure side, thus there's no point having different names for secure/non-secure sides of a target
* removes details steps to use mbed-os-tf-m-regression-tests, as more up-to-date instructions are in the mbed-os-tf-m-regression-tests repo itself
* removes the section "Enabling PSA at application level" which does not always work
* documents more configurations:
    * TF-M version supported
    * support for TF-M bootloader and two image slots
    * post binary hooks
* combines documentation for dual-core targets for Armv8-m targets
* replaces the Musca A1 example with Musca S1

### DIFF
--- a/docs/porting/security/psa_targets_in_mbed-os.md
+++ b/docs/porting/security/psa_targets_in_mbed-os.md
@@ -357,22 +357,6 @@ If an Armv8-M PSA target cannot inherit from `PSA_V8_M` because it has to inheri
         ]
 ```
 
-## Enabling PSA at application level
-
-Having an entropy source is crucial for Mbed TLS and PSA. The [entropy source porting document](../porting/entropy-sources.html) discusses why and how to add an entropy source.
-
-If your target doesn't have a True Random Number Generator (TRNG), configure it as a non-PSA target in `targets/targets.json`. In this scenario, if an application wants to use the target as a PSA target, then it is the application's responsibility to provide an entropy source and mark that target as a PSA target at the application level. To enable PSA for a target from an application, use the config option [target_overrides](../reference/configuration.html).
-
-An example of `mbed_app.json`:
-
-```json
-"target_overrides": {
-    "K64F": {
-        "inherits": ["PSA_V7_M"]
-    }
-}
-```
-
 ## Build and validation
 
 The [mbed-os-tf-m-regression-tests repository](https://github.com/ARMmbed/mbed-os-tf-m-regression-tests) contains build scripts, TF-M regression tests and PSA API compliance tests. Please follow instructions there to build and validate your TF-M target.
--- a/docs/porting/security/psa_targets_in_mbed-os.md
+++ b/docs/porting/security/psa_targets_in_mbed-os.md
@@ -375,73 +375,7 @@ An example of `mbed_app.json`:
 
 ## Build and validation
 
-The Python script `build_tfm.py` automates building TF-M and copying the TF-M binary to the location defined by the target attribute `tfm_delivery_dir`. Another Python script, `build_psa_compliance.py`, automates
-building PSA API compliance tests.
-
-The [mbed-os-tf-m-regression-tests repository](https://github.com/ARMmbed/mbed-os-tf-m-regression-tests) contains build scripts, TF-M regression tests and PSA API compliance tests.
-
-### Building TF-M and running regression tests
-
-To build TF-M and regression tests:
-
-1. Clone the [mbed-os-tf-m-regression-tests](https://github.com/ARMmbed/mbed-os-tf-m-regression-tests) repository.
-1. Update `mbed-os.lib` to the copy of Mbed OS that contains new target support, including the `targets.json` changes described in this document.
-1. Run:
-
-    ```
-    mbed deploy
-    ```
-
-1. Run:
-
-    ```
-    python3 build_tfm.py -m <new target> -t <toolchain> -c ConfigRegressionIPC.cmake
-    ```
-
-    The command builds a TF-M binary the service necessary to run regression tests.
-
-    <span class="notes">**Note:** Use this `ConfigRegressionIPC.cmake` only to run regression tests. The secure binary copied to Mbed OS (when you are ready to merge your new target) must be generated using the config option `ConfigCoreIPC.cmake`.</span>
-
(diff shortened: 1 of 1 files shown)


docs/porting/security/psa_targets_in_mbed-os.md
```diff
@@ -466,7 +396,7 @@ To build TF-M and create an Mbed OS pull request:
     python3 build_tfm.py -m <new target> -t <toolchain> --commit
     ```
 
-    The command builds TF-M with the config `ConfigCoreIPC.cmake`, copies the TF-M binary to the location defined by the target attribute `tfm_delivery_dir`, and commits the changes to Mbed OS.
+    The command builds TF-M with the config `ConfigCoreIPC`, copies the TF-M binary to the location defined by the target attribute `tfm_delivery_dir`, and commits the changes to Mbed OS.
```
jainvikas8: I think its `CoreIPC`

LDong-Arm: Good catch

docs/porting/security/psa_targets_in_mbed-os.md
```diff
@@ -39,16 +39,6 @@ Example of a single-core Armv7-M PSA target's inheritance:
...
-### Naming convention for dual-core and Armv8-M targets
-
-As explained above, dual-core and Armv8-M PSA targets require only an NSPE target name. For Armv8-M non-PSA targets, you can define both SPE and NSPE target names.
-
-The naming convention is:
-
-- `TargetName`: PSA non-secure target (NSPE).
-- `TargetName_NPSA_S`: Non-PSA secure target.
-- `TargetName_NPSA`: Non-PSA non-secure target.
-
```
jeromecoutant, line 51: So what is the new convention ?

For ex, current NUCLEO_L552ZE_Q is used as a legacy coretex M target without trust zone.
When we will support TFM, how this new target should be called ?

docs/porting/security/psa_targets_in_mbed-os.md
```diff
@@ -466,7 +396,7 @@ To build TF-M and create an Mbed OS pull request:
     python3 build_tfm.py -m <new target> -t <toolchain> --commit
```
jeromecoutant, line 289: This is a linux only command ?

Maybe mbed cli2 could be used ? mbedtools tfm -m xxx ?

docs/porting/security/psa_targets_in_mbed-os.md
LDong-Arm: > For ex, current NUCLEO_L552ZE_Q is used as a legacy coretex M target without trust zone.
> When we will support TFM, how this new target should be called ?

I think it should still be called NUCLEO_L552ZE_Q if we support only one flavour. Unless the intention is to provide users both TF-M and non-TF-M options?

docs/porting/security/psa_targets_in_mbed-os.md
LDong-Arm: It runs on Windows and macOS too, as long as python and dependencies are installed. Mbed CLI 2 support for PSA/TF-M targets and repositories will be added, at which point we'll update the documentation.

docs/porting/security/psa_targets_in_mbed-os.md
jeromecoutant: First intention was to provide both options.
But maybe we need to change our mind.

@LMESTM

LDong-Arm: > First intention was to provide both options.

This should be okay, we can still have two target names (for TF-M and non-TF-M flavours) in this case I think. I removed the naming guideline section from the doc, only because no targets currently follow them.

docs/porting/security/psa_targets_in_mbed-os.md
```diff
@@ -163,16 +163,20 @@ In this case, you must add the following additional attributes:
...
-A target can be categorized as a dual-core target if it has at least two cores that are either Armv7-M or Armv6-M. On dual-core PSA targets, TF-M runs on the SPE and provides PSA services.
+An Mbed OS (NSPE) target that supports TF-M must contain the following attributes (in addition to other target attributes defined in [porting a custom board](../porting/porting-a-custom-board.html) and [porting targets](../porting/porting-the-hal-apis.html)):
 
-An Mbed OS (NSPE) target must contain the following attributes (in addition to other target attributes defined in [porting a custom board](../porting/porting-a-custom-board.html) and [porting targets](../porting/porting-the-hal-apis.html)):
-
-* `inherits`: PSA generic target `PSA_DUAL_CORE`, unless the target has to inherit
+* `inherits`: PSA generic target `PSA_DUAL_CORE` or `PSA_V8_M`, unless the target has to inherit
   from one of its family targets.
+* `extra_labels`: `TARGET_`[label] paths to enable.
+  * `TFM_LATEST`: If the latest release of TF-M is supported by the target.
```
evedon: We need to specify that this is the latest release of TF-M as indicated in https://github.com/ARMmbed/mbed-os-tf-m-regression-tests as this is not necessarily the latest TFM release.

docs/porting/security/psa_targets_in_mbed-os.md
evedon: I'd recommend removing the naming guideline for now as it is not currently used. As @LDong-Arm says, both options could still be used and so this could be re-visited when a use case arises.

docs/porting/security/psa_targets_in_mbed-os.md
LDong-Arm: I've updated the wording:
* `TFM_LATEST`: If the current version of TF-M (that Mbed OS is integrated with) is supported by the target. The current version is indicated in [mbed-os-tf-m-regression-tests](https://github.com/ARMmbed/mbed-os-tf-m-regression-tests).

docs/porting/security/psa_targets_in_mbed-os.md
LDong-Arm: Done

evedon (approved):

evedon:
@donatieng Could you merge this PR please?

PR #1440, opened 2021-03-31
Remove TDB_EXTERNAL_NO_RBP and FILESYSTEM_NO_RBP from KVStore docs

harmut01:
Preceeding PR: https://github.com/ARMmbed/mbed-os/pull/14490

NO_RBP (no rollback protection) is intended to not require an internal
TDB, however, DeviceKey, which we use to derive SecureStore's
encryption key, still does. Currently, no internal TDB is created with
these two configurations, meaning there's no way to store the DeviceKey
and SecureStore doesn't work.

Update the documentation to reflect these changes.

### DIFF
--- a/docs/reference/configuration/storage-config.md
+++ b/docs/reference/configuration/storage-config.md
@@ -27,28 +27,20 @@ kvstore
     ├───tdb_external
     │       mbed_lib.json
     │
-    ├───tdb_external_no_rbp
-    │       mbed_lib.json
-    │
     ├───filesystem
-    │       mbed_lib.json
-    │
-    ├───filesystem_no_rbp
     │       mbed_lib.json
     │
     └───tdb_internal
             mbed_lib.json
 ```
 
-The KVStore configuration file structure includes six configuration files. You can use the topmost configuration file to set up the full configuration of the storage by defining a single parameter (`storage_type`) to one of the predefined configurations. You can use the configuration files in the subfoldersto implement the above top level configurations.
+The KVStore configuration file structure includes four configuration files. You can use the topmost configuration file to set up the full configuration of the storage by defining a single parameter (`storage_type`) to one of the predefined configurations. You can use the configuration files in the subfoldersto implement the above top level configurations.
 
 You can find the configuration files under `conf/<configuration name>`:
 
 - `conf/tdb_internal` - The storage type `TDB_INTERNAL` configuration is intended to be used when all data will be stored in internal memory only. There is no need for additional security features. A single TDBStore object will be allocated in internal flash.
 - `conf/tdb_external` - The storage type `TDB_EXTERNAL` configuration provides full security and is intended to be used when data is stored in external flash. It allocates: SecureStore, TDBStore in external flash and TDBStore in internal flash for rollback protection (RBP).
-- `conf/tdb_external_no_rbp` - The storage type `TDB_EXTERNAL_NO_RBP` configuration allows security but without rollback protection. This is similar to `tdb_external` but without the TDBStore in internal memory.
 - `conf/filesystem` - This configuration allocates SecureStore, FileSystemStore, filesystem, TDBStore in internal memory and the required block devices. The allocated file system is selected according to the COMPONENT set in `targets.json` (FATFS for SD card and LITTLEFS for SPIF). However, you can set this differently by overriding the respective parameter. Use this configuration if you need the file system with the POSIX API in addition to the set/get API.
-- `conf/filesystem_no_rbp` - The storage type `FILESYSTEM_NO_RBP` configuration allows security like the FILESYSTEM configuration but without rollback protection.
 
 A standalone block device is allocated for each component in internal and external memory and SD cards as required for the configurations. The full size of the memory allocated for each block device will be used by the respective component.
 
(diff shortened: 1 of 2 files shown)


docs/reference/configuration/storage-config.md
```diff
@@ -104,7 +94,7 @@ In this configuration, all KVStore C APIs will be mapped to the TDBStore in the
 
 Below is the `TDB_INTERNAL` configuration in `mbed_lib.json`:
 
-```
+``` json
```
Patater: This appears to work on GitHub OK. Do we know if our docs engine needs a space after the triple backticks or if it can't cope with a space there?

LDong-Arm: Noticed the same when reviewing the other PRs on this topic. What about just not have any space at all?

harmut01: I can't find any documentation that says there shouldn't be a space, although all the snippets I've seen don't add one. 

- https://www.doxygen.nl/manual/markdown.html#md_fenced
- https://www.markdownguide.org/extended-syntax/#syntax-highlighting

harmut01: I don't feel strongly either way, so I'll remove the space just in case.

LDong-Arm (approved):

docs/reference/configuration/storage-config.md
```diff
@@ -250,57 +209,12 @@ Below is the `FILESYSTEM` configuration in `mbed_lib.json`:
...
-```
-
-If the file system is not set, the default file system and block device will be applied, and `blockdevice`, `external_size` and `external_base_address` will be ignored.
 
 ### Configuration functions API
 
 Applications must call the function **storage_configuration()** to instantiate the required configuration. This function is defined as weak to allow the replacement of this function with a completely different implementation of the instantiation of components:
 
-```
+```c++
```
LDong-Arm: This failed Travis, because it's not a complete snippet - `_STORAGE_CONFIG` is not declared in the current scope. The documentation in the Mbed OS repo does provides a complete one: https://github.com/ARMmbed/mbed-os/blob/master/storage/docs/Configuration/CONFIGURATION.md#configuration-functions-api (or in your other PR).

For this PR, maybe we can fix this snippet by copying things over. But in my opinion it's better to avoid duplication - should we have this kind of docs in Mbed OS or here? @Patater

LDong-Arm (changes requested):

docs/reference/configuration/storage-config.md
harmut01: @LDong-Arm I'm not sure if this will work but can we include [kv_config.h](https://github.com/ARMmbed/mbed-os/blob/73896715e6202e187ba624fc3853299e73390f19/storage/kvstore/kv_config/include/kv_config/kv_config.h#L29-L30) in the code snippet?

LDong-Arm: Good idea! Each snippet is compiled as an Mbed application, so it might work. Try to push the change and see if Travis passes.

LDong-Arm: I don’t see `storage_configuration()` anywhere in Mbed OS code. The doc says

> Applications must call the function **storage_configuration()** to instantiate the required configuration.

but in fact no such kind of call is required, because [kvstore_global_api.cpp](https://github.com/ARMmbed/mbed-os/blob/mbed-os-6.9.0/storage/kvstore/kvstore_global_api/source/kvstore_global_api.cpp) calls [kv_init_storage_config()](https://github.com/ARMmbed/mbed-os/blob/mbed-os-6.9.0/storage/kvstore/kv_config/source/kv_config.cpp#L991-L1018) to initialize storage on every API call.

Maybe `kv_init_storage_config()` is what this doc actually means, but even then applications are not required to call it. I wonder if it makes sense to remove this section.

harmut01: You're right it, I can't see `storage_configuration()` either, its probably an error in the documentation. I might be missing something here but I think they may need to call `kv_init_storage_config()` if they aren't using the API defined in [kvstore_global_api.cpp](https://github.com/ARMmbed/mbed-os/blob/mbed-os-6.9.0/storage/kvstore/kvstore_global_api/source/kvstore_global_api.cpp). As I understand it this API works in parallel with KVStore, which the user should still be able to use without it.

LDong-Arm: [`kv_init_storage_config()`](https://github.com/ARMmbed/mbed-os/blob/mbed-os-6.9.0/storage/kvstore/kv_config/source/kv_config.cpp#L991-L1018) sets up _global_ instances of `KVStore`s (via `_storage_config_*()`) and adds those instances to `KVMap` whose purpose is to globally manage multiple `KVStore`s. And the main user of `KVMap` is `kvstore_global_api` which automatically calls `kv_init_storage_config()`. I'm not sure the map is supposed to be used directly by a user, without `kvstore_global_api` - the former is more like a utility just for the latter.

> As I understand it this API works in parallel with KVStore, which the user should still be able to use without it.

When a KVStore (TDBStore, FileSystemStore, SecureStore) is used directly without kvstore_global_api, it's a standalone instances locally managed by the application/user, and the global configurations we are looking at are irrelevant.

But anyway, the change looks okay to me for now - this is not originally the main point of this PR.

LDong-Arm (approved):
LGTM

LDong-Arm (approved):
As we switched from removal to deprecation, the documentation clean-up is still valid - we just retained the config and code with deprecation notices.

So this PR still looks good to me.

LDong-Arm:
@donatieng Could you please merge this? It's part of the ongoing release, thanks

PR #1441, opened 2021-04-06
mbed_cli_2: Add migration guide

### LINKED ISSUE #227: Document how to migrate to the new Mbed tools
### Description

As an old-time Mbed user, I want to understand how to use the new Mbed tools

- Create a migration guide to help teach the new tools
- How to do what you used to do with the old tools, but with the new tools now

### Issue request type

- [X] Enhancement
- [ ] Bug

### Tasks

- [ ] Come up with lower-level commands where we've removed top level command
- [ ] Add rationale for command removal
- [ ] mbed-os-5-docs review

Patater:
Add a migration guide to help show how to do things that Mbed CLI 1
could do using Mbed CLI 2.

Fixes ARMMbed/mbed-tools#227

### DIFF
--- a/docs/tools/mbed_cli_2/migration_guide.md
+++ b/docs/tools/mbed_cli_2/migration_guide.md
@@ -1,0 +1,212 @@
+# Migration guide
+
+If you haven't already read the [introduction](./intro.md), then start from there.
+
+The background color indicates:
+
+* <span style="background-color:#ccffcc;">Green: New Functionality
+  introduced in Mbed CLI 2 (mbed-tools)</span>
+* <span style="background-color:#ffe6cc;">Orange: Functionality
+  modified in Mbed CLI 2 (mbed-tools) compared to Mbed CLI 1 (mbed-cli)</span>
+* <span style="background-color:#ffcccc;">Red: Functionality deprecated
+  in Mbed CLI 2 (mbed-tools)</span>
+
+## Installers
+<span style="background-color:#ffe6cc;">Windows, macOS and Linux
+installers may be made available at later point of time.</span><br>
+
+## Commands
+
+### Device management
+<span style="background-color:#ffcccc;">The subcommand `mbed-cli
+device-management` (device management) is deprecated. Refer to your cloud
+provider's documentation on how to cloud-manage your devices.</span><br>
+
+### Repository management
+<span style="background-color:#ffcccc;">Support for Mercurial and the
+subcommand `mbed-cli publish` (publish program or library) is deprecated.
+Hosted repositories on "mbed.org" are not supported from Mbed CLI 2. Version
+control with git can be used as an alternative. Hosting for git repositories is
+available on [GitHub](https://github.com/).</span><br>
+
+<span style="background-color:#ffcccc;">The subcommand `mbed-cli cache`
+(repository cache management) is deprecated. No replacement is
+supported.</span><br>
+
+<span style="background-color:#ffcccc;">The subcommand `mbed-cli releases`
+(show release tags) is deprecated. We recommend using standard `git` commands
+instead, for example `git tag -l` to list all tagged releases.</span><br>
+
+<span style="background-color:#ffe6cc;">The subcommand `mbed-cli
+update` (update to branch, tag, revision or latest) is not implemented. Use
+standard `git` commands instead. From your application directory: to check out
+a branch of Mbed OS, `git -C mbed-os checkout branchname`; to checkout the Mbed
+OS 6.8.0 release, `git -C mbed-os checkout mbed-os-6.8.0`; to check out Mbed OS
+revision `3e24a7ea9602`, `git -C mbed-os checkout 3e24a7ea9602`; to checkout
+the latest released version of Mbed OS `git -C mbed-os checkout
+latest`.</span><br>
+
+<span style="background-color:#ffcccc;">The subcommand `mbed-cli export`
+(generate an IDE project) is deprecated. Use [CMake
+generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html)
+instead. While CMake supports many generators, including many not supported by
+Mbed CLI 1, not all of Mbed CLI 1's exporters have replacements available yet.
+The available project generators are listed below.</span><br>
+
+- [CodeBlocks](https://cmake.org/cmake/help/latest/generator/CodeBlocks.html)
+- [Eclipse](https://cmake.org/cmake/help/latest/generator/Eclipse%20CDT4.html)
+- [Make](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#id10)
+- [Qt Creator](https://doc.qt.io/qtcreator/creator-project-cmake.html)
+- [Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cmake-tools)
+
+<span style="background-color:#ffcccc;">If you'd like to help us prioritize
+which CMake project generators to work on first, please give your thumbs up to
+the following issues you care most about, or [raise a merge request to
+CMake](https://gitlab.kitware.com/cmake/cmake/-/merge_requests):</span><br>
+
+- [CrossCore® Embedded Studio](https://github.com/ARMmbed/mbed-tools/issues/249)
+- [e<sup>2</sup> studio](https://github.com/ARMmbed/mbed-tools/issues/250)
+- [Embitz](https://github.com/ARMmbed/mbed-tools/issues/251)
+- [IAR Embedded Workbench for Arm](https://github.com/ARMmbed/mbed-tools/issues/252)
+- [MCUXpresso](https://github.com/ARMmbed/mbed-tools/issues/253)
+- [NetBeans](https://github.com/ARMmbed/mbed-tools/issues/254)
+- [STM32CubeIDE](https://github.com/ARMmbed/mbed-tools/issues/257)
+- [Keil uVision](https://github.com/ARMmbed/mbed-tools/issues/256)
+
+### Library management
+
+<span style="background-color:#ffcccc;">The subcommands `mbed-cli add` (add
+library from URL) and `mbed-cli remove` (remove library) are deprecated.
+Instead, use `git clone` or `mbed-tools import` to clone a library. Then,
+manually create a `reponame.lib` file that contains a single line in the form
+`https://github.com/ARMmbed/reponame#branch-or-tag` to fetch a specific branch
+or tag. If you want to fetch the default branch, you don't need to add
+`#branch-or-tag-name`.</span><br>
+
+<span style="background-color:#ffcccc;">The subcommand `mbed-cli new --library`
+(force creation of an Mbed library) is deprecated. Create a new Mbed library by
+creating a new folder and adding it to git version control. Then, create a
+`reponame.lib` in the application you wish to use the new library from, as
+described above (in the `mbed-cli add` paragraph).</span><br>
+
+<span style="background-color:#ffcccc;">The subcommand `mbed-cli ls` (view
+dependency tree) is deprecated. The subcommand `mbed-tools deploy` will now
+display all dependencies.</span><br>
+
+### Tool configuration
+<span style="background-color:#ffcccc;">The subcommand `mbed-cli config` (tool
+configuration) is deprecated. There is no need to configure an Mbed-specific
+compiler path in Mbed CLI 2; the compiler path is determined in a standard way
+by CMake instead. Other configuration options, like `target`, `toolchain`,
+`protocol`, `depth` and `cache`, are not supported in Mbed CLI 2.</span><br>
+
+<span style="background-color:#ffcccc;">The subcommand `mbed-cli toolchain`
+(set or get default toolchain) is deprecated. Pass the desired toolchain on the
+command line when compiling with `mbed-tools compile --toolchain GCC_ARM`, to,
+for example, select `GCC_ARM` as the toolchain.</span><br>
+
+<span style="background-color:#ffcccc;">The subcommand `mbed-cli target` (set
+or get default target) is deprecated. Pass the desired target on the command
+line when compiling with `mbed-tools compile --mbed-target
+DISCO_L475VG_IOT01A`, to, for example, select `DISCO_L475VG_IOT01A` as the
+Mbed target.</span><br>
+
+
+### Creating an Mbed program
+<span style="background-color:#ffe6cc;">The subcommand `mbed-tools new` creates
+a new Mbed program by default and supports only one command-line-option
+`--create-only` (create a program without fetching mbed-os). The following
+command-line options, previously supported in Mbed CLI 1, are deprecated:
+
+* `--scm`: We now always use git.
+* `--program`: There is no need to specify this. `mbed-tools new` always
+  creates a program by creating a top-level application CMakeLists.txt that
+  invokes `add_executable()`.
+* `--library`: There is no need to specify this. `mbed-tools new` always
+  creates a program. If you wish to create a library, you can edit the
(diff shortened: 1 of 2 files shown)


docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+- [STM32CubeIDE](https://github.com/ARMmbed/mbed-tools/issues/257)
+- [Keil uVision](https://github.com/ARMmbed/mbed-tools/issues/256)
+
+### Library management
+
+<span style="background-color:#ffcccc;">The subcommands `mbed-cli add` (add
+library from URL) and `mbed-cli remove` (remove library) are deprecated.
+Instead, use `git clone` or `mbed-tools import` to clone a library. Then,
+manually create a `reponame.lib` file that contains a single line like
+`https://github.com/ARMmbed/reponame#branch-or-tag`, pointing to your repo at
```
rwalton-arm: The way this section is written is a little bit confusing when you first read it. Maybe it would be clearer if we said something like:

>... create a `reponame.lib` file that contains a single line in the form `https://github.com/user/reponame#branch-or-tag-name` to fetch a specific branch or tag. If you want to fetch the default branch you don't need to add `#branch-or-tag-name`.

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+the default branch (no `#` and nothing after the `#`), desired tag (append with
+`#tagname`), or desired branch (append with `#branchname`).</span><br>
+
+<span style="background-color:#ffcccc;">The subcommand `mbed-cli new --library`
+(force creation of an Mbed library) is deprecated. Create a new Mbed library by
+hand instead, using standard git commands.</span><br>
```
rwalton-arm: Is adding the `.lib` file a standard git command? The `.lib` file is what actually let's us know if this is a library, right? Maybe we should link to the bit where we describe how to create the `.lib` file here, or at least mention it.

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+* `--offline` (not needed, as there is no longer any Mbed-specific git cache)
+* `--no-requirements` (Mbed CLI 2 never automatically installs requirements)
+</span><br>
+
+### Importing an Mbed program
+<span style="background-color:#ffe6cc;">The subcommand `mbed-tools
+import` clones an Mbed project and library dependencies. The following
+command-line-options supported in Mbed CLI 1 are deprecated:
+
+* `--depth`: The new CLI optimistically imports the most shallow repository
```
rwalton-arm: Should it be "opportunistically" instead of "optimistically"?

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+  possible. Running `git clone --depth <depth> <repo_url> | cd <repo_name> &&
+  mbed-tools deploy` will create an mbed project with the required depth.
+* `--protocol`: To select the protocol, import using the URL for the
+  desired protocol. For instance, the tool allows you to import projects or
+  libraries over SSH as follows `mbed-tools import
+  git://github.com/ARMmbed/<some_driver>`. Alternatively, if you would prefer
+  to use SSH you can add the following into your .gitconfig to force the tool
```
rwalton-arm: "to always use SSH"?

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+or use `cmake --build cmake_build --target <library_name> --target
+<other_library_name>` to add any number of targets to your build.</span><br>
+
+<span style="background-color:#ffcccc;">The command-line-option "mbed
+compile --library" is deprecated. Use standard CMake `add_library()`
+instead.</span><br>
+
+<span style="background-color:#ffcccc;">The command-line-option `mbed
+compile -D MACRO_NAME` is deprecated. Use `mbed-tools configure` followed by
+`cmake --build cmake_build -D MACRO_NAME` or `target_compile_definitions()` in
```
rwalton-arm: Don't you pass the macros in at CMake configuration time? Wouldn't the invocation be `cmake -S . -B cmake_build -DMACRO_NAME`?

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+application "CMakeLists.txt".</span><br>
+
+<span style="background-color:#ffe6cc;">The command-line-option `mbed
+compile --build` will be supported at later point of time. Although we don't
+currently have the option to explicitly state the build path, we have an ticket
```
rwalton-arm: "a ticket"

rwalton-arm (changes requested):

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+### Creating an Mbed program
+<span style="background-color:#ffe6cc;">The subcommand `mbed-tools new` creates
+a new Mbed program by default and supports only one command-line-option
+`--create-only` (create a program without fetching mbed-os). The other
+command-line-options supported in Mbed CLI 1, as follow, are deprecated:
+
+* `--scm` (now always git)
+* `--program` (now always a program)
+* `--library` (now always a program)
+* `--mbedlib` (instead, add with a separate subsequent command)
```
wernerlewis: From the CLI 1 documentation and in:
https://github.com/ARMmbed/mbed-cli/blob/master/mbed/mbed.py#L2151
https://github.com/ARMmbed/mbed-cli/blob/master/mbed/mbed.py#L140 
the `--mbedlib` flag in mbed-cli imports the Mbed 2 library instead of Mbed OS, so this line should explain this is deprecated.

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Configuring an Mbed program
+
+<span style="background-color:#ccffcc;">The subcommand `mbed-tools configure`
+generates an Mbed OS config CMake file named `mbed_config.cmake` in the path
+`cmake_build/<TARGET_NAME>/\<PROFILE>/\<TOOLCHAIN>/`.</span><br>
+
+### Compiling an Mbed program
+
+<span style="background-color:#ffcccc;">The subcommand "mbed-tools
```
wernerlewis: Should be ``` `mbed-tools compile` ``` for consistency, there are a couple of examples of this in this section which use"" instead.

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+line when compiling with `mbed-tools compile --mbed-target
+DISCO_L475VG_IOT01A`, to, for example, select `DISCO_L475VG_IOT01A` as the
+Mbed target.</span><br>
+
+
+### Creating an Mbed program
+<span style="background-color:#ffe6cc;">The subcommand `mbed-tools new` creates
+a new Mbed program by default and supports only one command-line-option
+`--create-only` (create a program without fetching mbed-os). The other
+command-line-options supported in Mbed CLI 1, as follow, are deprecated:
```
wernerlewis:
```suggestion
command-line-options supported in Mbed CLI 1, as follows, are deprecated:
```

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+
+* `--scm` (now always git)
+* `--program` (now always a program)
+* `--library` (now always a program)
+* `--mbedlib` (instead, add with a separate subsequent command)
+* `--depth` (no longer needed, as the Mbed CLI 2 automatically uses the shallowest clone or
```
wernerlewis:
```suggestion
* `--depth` (no longer needed, as Mbed CLI 2 automatically uses the shallowest clone or
```

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+</span><br>
+
+### Importing an Mbed program
+<span style="background-color:#ffe6cc;">The subcommand `mbed-tools
+import` clones an Mbed project and library dependencies. The following
+command-line-options supported in Mbed CLI 1 are deprecated:
+
+* `--depth`: The new CLI optimistically imports the most shallow repository
+  possible. Running `git clone --depth <depth> <repo_url> | cd <repo_name> &&
+  mbed-tools deploy` will create an mbed project with the required depth.
```
wernerlewis:
```suggestion
  mbed-tools deploy` will create an Mbed project with the required depth.
```

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,188 @@
...
+
+<span style="background-color:#ffe6cc;">The command-line-option `mbed
+compile --build` will be supported at later point of time. Although we don't
+currently have the option to explicitly state the build path, we have an ticket
+open to track the issue
+[#184](https://github.com/ARMmbed/mbed-tools/issues/184).</span><br>
+
+### Testing an Mbed program
+
+<span style="background-color:#ffe6cc;">The subcommand "mbed test"
```
wernerlewis:
```suggestion
<span style="background-color:#ffe6cc;">The subcommand `mbed test`
```

wernerlewis (changes requested):

docs/tools/mbed_cli_2/migration_guide.md
Patater: I thought this was about creating a library from a sort of template. I might have misunderstood the old docs and am not understanding what this command does. If it creates a lib file, this does indeed need updating.

LDong-Arm:
Could add "Fixes #227" in the PR description to link to the issue

Patater:
> Could add "Fixes #227" in the PR description to link to the issue

Added `Fixes ARMmbed/mbed-tools#227` to the commit message to make the link. Thanks.

LDong-Arm:
> > Could add "Fixes #227" in the PR description to link to the issue
> 
> Added `Fixes ARMmbed/mbed-tools#227` to the commit message to make the link. Thanks.

Ah I didn't realise it was a different project. It seems GitHub may not pick up links sometimes.

docs/tools/mbed_cli_2/migration_guide.md
Patater: Yes, this command does make a lib file. I'll update.

docs/tools/mbed_cli_2/migration_guide.md
Patater: This brings up a good grammar question. Which is correct?

1. `Your options are as follows:`
2. `Your options are as follow:`

Patater: TIL https://www.grammarphobia.com/blog/2013/02/as-follows.html

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,195 @@
...
+want to build either modify your application's `CMakeLists.txt` to add/exclude
+or use `cmake --build cmake_build --target <library_name> --target
+<other_library_name>` to add any number of targets to your build.</span><br>
+
+<span style="background-color:#ffcccc;">The command-line-option `mbed
+compile --library` is deprecated. Use standard CMake `add_library()`
+instead.</span><br>
+
+<span style="background-color:#ffcccc;">The command-line-option `mbed compile
+-DMACRO_NAME` is deprecated. Use `mbed-tools configure -DMACRO_NAME` followed
```
wernerlewis: `mbed-tools configure -DMACRO_NAME` is not supported. I think this section could probably just be reduced down to the option of adding the definition in `CMakeLists.txt`.

wernerlewis (changes requested):

docs/tools/mbed_cli_2/migration_guide.md
Patater: I'll describe how to do this with CMake commands, then.

rwalton-arm: `mbed-tools configure` doesn't let you pass in macros, does it? You have to run a cmake configure directly to pass them in on the command line. `mbed-tools configure` generates `mbed_config.cmake` and the build directory, but it doesn't run the cmake configure step to generate the build system.

rwalton-arm (changes requested):

docs/tools/mbed_cli_2/migration_guide.md
Patater: I'll describe how to do this with CMake commands, then.

wernerlewis (approved):
LGTM

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,198 @@
...
+</span><br>
+
+### Configuring an Mbed program
+
+<span style="background-color:#ccffcc;">The subcommand `mbed-tools configure`
+generates an Mbed OS config CMake file named `mbed_config.cmake` in the path
+`cmake_build/<TARGET_NAME>/<PROFILE>/<TOOLCHAIN>/`.</span><br>
+
+### Compiling an Mbed program
+
```
LDong-Arm, line 175: Another previously-undocumented Mbed CLI 1 feature is also not supported by Mbed CLI 2: use `detect` in place of a target name in `mbed-cli compile -m detect`, if only one board is connected. Is it something we want to support or drop?

rwalton-arm (approved):

docs/tools/mbed_cli_2/migration_guide.md
```diff
@@ -0,0 +1,198 @@
...
+### Creating an Mbed program
+<span style="background-color:#ffe6cc;">The subcommand `mbed-tools new` creates
+a new Mbed program by default and supports only one command-line-option
+`--create-only` (create a program without fetching mbed-os). The following
+command-line options, previously supported in Mbed CLI 1, are deprecated:
+
+* `--scm`: We now always use git.
+* `--program`: There is no need to specify this. We always assume a program.
+* `--library`: There is no need to specify this. We always assume a program.
```
LDong-Arm: Well, this is actually controlled by CMake, rather than what we "assume":
* An application's `CMakeLists.txt` should call `add_executable(<app_name>)` to build an executable and `mbed_set_post_build()` to generate a flashable binary (.bin or .hex or both)
* A library's `CMakeLists.txt` calls `add_library(<lib_name)` _without_ calling `mbed_set_post_build()` to build a static library (.a). This is sort of like `--library` of the old tools.

I've experimented that it's totally fine to build only libraries with our tools without an application, as long as the CMake definitions are correct.
(Users could mix and match, of course)

LDong-Arm:
^ A couple of last-minute comments

docs/tools/mbed_cli_2/migration_guide.md
rwalton-arm: This is correct and I missed it on my last round of reviews. We don't even "assume a program" in the tool, we assume CMake can generate Ninja build system instructions and the "all" target can be built. We should just say something like "set this in CMake" and perhaps link to a tutorial for configuring a library target.

Patater: `mbed-tools new` doesn't handle this difference and always assumes you want to make a program. We have one template, and it always makes an application. I'll add instructions on how to modify CMake to create a library.

LDong-Arm (approved):
LGTM

rwalton-arm (approved):

wernerlewis (approved):

donatieng (approved):
LGTM, thanks!

PR #1444, opened 2021-04-15
Workaround for Python requests library not handling redirections properly

donatieng:
One of our snippets URL is redirected to the same URL with a trailing slash `/`. When fetched by the requests Python library in the docs engine, the trailing slash is not taken into account leading to unlimited redirects.

Adding the slash to the URL in the relevant Markdown file fixes the issue.


donatieng:
Scratch that, Python's not at fault!

PR #1447, opened 2021-04-20
Std pin names

MarceloSalazar:
Minor fixes on titles to align with other docs

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -266,7 +266,7 @@
                             ]
                         },
                         {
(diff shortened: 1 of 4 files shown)


MarceloSalazar:
Note this doc can't be found: https://os.mbed.com/docs/mbed-os/development/apis/standard-pin-names-developer-guide.html

@donatieng thoughts on what might be wrong?

donatieng (approved):
Good stuff, thanks!

PR #1456, opened 2021-05-26
Chris dev

ChrisGooch-Arm:
Fix for two broken links in content (that have been there a while) found when running linkchecker on 6.11 docs.

Also temporary fix for docs.json to point to mbed-os-6.11.0-docs-utf-8-fix for doxygen output, as a workaround for the problem with non UTF-8 character preventing builds.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -8,7 +8,7 @@
             "hidden": true,
             "type": "doxygen",
             "source": "https://github.com/ARMmbed/mbed-os",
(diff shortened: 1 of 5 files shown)


ChrisGooch-Arm:
fixed some broken links and added item on introduction / docs updates section for new material added since last release

ChrisGooch-Arm (commented):
remove, no longer needed

donatieng (approved):
Just checked the preview sit - looks good, thanks!

PR #1461, opened 2021-06-24
Fix broken link on Fast Models page

donatieng:
The Fast Models documentation has been updated - this PR updates the link to the correct location


ChrisGooch-Arm (approved):
new link looks correct, same page as before (it moved in fast models project for 11.14). Is it worth mentioning the "User mode networking" title explicitly as well?

PR #1476, opened 2021-08-19
update intro section with link to new section on driver interfaces/polymorphism, plus minor fixes

ChrisGooch-Arm:


ChrisGooch-Arm:
@donatieng - this is to fix a few minor spelling etc. things in the new driver polymorphism section and to add blurb and link on the intro page "recently updated" section.

Are you happy with the wording in the intro about being experimental etc.? Also, I wonder if we should make the new section a page on its own rather than just appending to the drivers overview?

donatieng (approved):
Thanks Chris, thanks for adding the blurb - looks good. And I agree, as driver polymorphism targets advanced users only, it might make sense for this section to be in its own page.

PR #1479, opened 2021-08-26
Migrate TravisCI to GHA

jamesbeyond:
This PR Migrate TravisCI to GHA, due to our TravisCI license will expire soon.
The test is passed locally on my fork, should be good for review


jamesbeyond:
Hi @ChrisGooch-Arm could you review this PR

docs/api/connectivity/bluetooth/BLE.md
```diff
@@ -47,7 +47,7 @@ the events (methods) you are interested in, the others will inherit the do-nothi
 
 All BLE operations are executed on an instance of BLE accessible through a function in the header `ble/BLE.h`.
 
-```c
+```cpp
```
0xc0170, line 50: The fixes in .md files should be separate commit to understand what they are fixing

docs/api/drivers/driver_interfaces.md
```diff
@@ -72,7 +72,7 @@ class LEDAnimator {
     }
 
     LEDAnimator(mbed::interface::DigitalOut *red, mbed::interface::DigitalOut *green, mbed::interface::DigitalOut *blue) : _red(red), _green(green), _blue(blue), _internally_created(false) {
-    ...
```
ChrisGooch-Arm, line 75: was the ... intended to show that you can add your own code at this point? perhaps rather than simply leave blank, we should add some appropriate code comment on this line?

docs/api/connectivity/bluetooth/BLE.md
ChrisGooch-Arm: I agree a separate commit for the docs changes (even if they are to code samples) would be clearer. I will add comments here anyway.

docs/api/drivers/driver_interfaces.md
```diff
@@ -103,7 +103,7 @@ The inheritance of a driver from the new driver interface is conditional upon `F
 
 When `FEATURE_EXPERIMENTAL_API` is disabled, the following type alias is made instead:
 
-```c++
```
ChrisGooch-Arm, line 106: not sure why this is being removed here, as the other code snippets in this file are marked as being C++ in the same way.

docs/api/connectivity/bluetooth/BLE.md
jamesbeyond: Fixed

docs/api/drivers/driver_interfaces.md
jamesbeyond: yes, I guess so the `…` mean your own code. 
However this is not standard c/cpp syntax. the snippets won't compile. one task the GHA is doing is trying to complier those snippets make sure they are valid. so we have to change this? 
that about change that to a comment line?
```
// your own code
```

docs/api/drivers/driver_interfaces.md
jamesbeyond: yeah, this failed at  completion job,  as this code already part of mbed-os code, so compile this snippet will through out error.  unmark this as `c++` will suppress the  compile test for this snippet

jamesbeyond:
Sorry for the delay reply, back from holidays now 😀

jamesbeyond:
@ChrisGooch-Arm could you review this again?

docs/api/drivers/driver_interfaces.md
ChrisGooch-Arm: ok (the website doesn't seem to do any useful syntax highlighting based on this anyway)

docs/api/drivers/driver_interfaces.md
ChrisGooch-Arm: I think this is fine, though I am not 100% sure of the intention of this part of the example code. Worth checking with @pan- who reviewed this new chapter originally?

ChrisGooch-Arm: Based on the example code in the next chapter, perhaps this is sufficient / suitable for the comment?

// do something

PR #1480, opened 2021-08-30
mbed-cli 1: deprecate dm command

0xc0170:
Fixes https://github.com/ARMmbed/mbed-os/issues/15009

Users should use manifest tool from Pelion instead.

I used 6.15 as the next release. It should be the feature release (assuming there are some outstanding feature PRs).

We have `deprecated` folder in the root, shall I just move the old page there instead? And keep this page as it is here in this pull request?


donatieng (approved):
LGTM

donatieng:
I think that removing is probably the best option as DM is unsupported now.

